PR #2987, opened 2023-11-17
Update vite.config.ts

Osub:
fix  codemirror error

[vite]: Rollup failed to resolve import "codemirror" from "node_modules/vue-codemirror/dist/vue-codemirror.esm.js".
This is most likely unintended because it can break your application at runtime.

#### Summary of your change
 this module explicitly add it to
`build.rollupOptions.external`


wanghe-fit2cloud:
感谢。

PR #7196, opened 2024-11-27
feat(alert): Add alert settings

lan-yonghui:
Refs #6357
Refs #7005


frontend/src/assets/iconfont/iconfont.css
```diff
@@ -14,6 +14,22 @@
...
+  content: "\e61e";
+}
+
+.p-17:before {
+  content: "\e618";
+}
+
 .p-alert-1:before {
   content: "\e611";
 }
```
fit2cloudrd, line 35: None

I have carefully analyzed all the provided code snippets, and did not find any inconsistencies or issues that should be addressed. Also, there do not seem to be any notable optimizations needed for this project's requirements.

Therefore, I can confidently conclude:

There is no need for adjustments in terms of style or functionality based on current knowledge cutoff 2021-09-01, nor from a view at the time of checking today (November 27th, 2024). If you've identified any discrepancies or need further guidance, feel free to ask!

frontend/src/global/form-rules.ts
```diff
@@ -806,4 +820,9 @@ export const Rules: CommonRule = {
         validator: checkIpv4,
         trigger: 'blur',
     },
+    phone: {
+        validator: checkPhone,
+        required: true,
+        trigger: 'blur',
+    },
 };
```
fit2cloudrd, line 828: The provided code checks for common rules like Https checking and phone validation before returning the `commonRules` object with specific validators.

Potential Issues:

1. Regular expression mismatch: The current regular expression is designed to match all digits including those that might also be followed by an extension but we're trying to check just numbers here which causes some mismatches.
2. Input Validation missing logic: This part seems to validate only input text type field without performing actual input processing.

Solution Suggestions:

- Validate entire string not just regex matches: In the first condition of `if`, change it so this validates the whole value instead of simply matching against a regex pattern.

Optimization Suggestions:

1. Consolidate repetitive validations into single line functions to improve readability
      ```js
       const isNumberCheck = (input, msg) => {
           if (!/^[+-]?(\d+(\.\d*)?$)(%)？]{0,}$/.test(input)) {
               return new ValidationError(msg)
          };
```

2. Use functional programming style instead of imperative loops when implementing rule implementations:
   ```javascript

        const getValidatorValueByField = (fieldName, fieldValue, formItem) =>
            fieldValue ? { [fieldName]: fieldValue.toString() } : {};

        const getValueFromValidatedFormItems = async (validatorResults, formItemRulesMap) => {
             let resFields = {};
                Object.keys(validatorResults).forEach((key) => {
                    if(Object.prototype.hasOwnProperty.call(formItemRulesMap , key)){
                        const {fieldConfig} = formItemRulesMap[key];
                       // use 'getValuFromValidatedFormItems' function from above to create mapped values for each field
                        resFields[fieldConfig.propertyKey] =(await getValuesForPropRes(resFields, getFieldByName(fieldConfig.propName))) ;
                            await addValidationMessagesToFieldIfNecessary(fieldConfig);
                    
                    }
                 });
                   return {...resFields};
             
```

frontend/src/views/host/terminal/terminal/index.vue
```diff
@@ -436,6 +436,12 @@ onMounted(() => {
...
     }
+    :deep(.el-tabs__item:hover) {
+        color: var(--panel-terminal-tag-hover-text-color);
+    }
+    :deep(.el-tabs__item.is-active:hover) {
+        color: var(--panel-terminal-tag-active-text-color);
+    }
 }
 
 .tagButton {
```
fit2cloudrd, line 447: This code appears to be related to styling components used within Electron apps developed with Electron.js and React. The key differences from 2021-09-01 might include:

- There is no reference to the date `2024-11-27`
- No specific issue, potential issue or optimization suggestion has been provided

If you have specific questions about how various parts of this code work together, please let me know!

wanghe-fit2cloud (commented):
/lgtm

PR #7307, opened 2024-12-10
fix(container): Fix the issue of container command parsing failure

ssongliu:


frontend/src/views/container/container/operate/index.vue
```diff
@@ -630,15 +630,25 @@ const isFromApp = (rowData: Container.ContainerHelper) => {
...
+        } else if (match[2]) {
+            result.push(match[2].replaceAll('<quota>', '\\"'));
+        }
     }
-    return returnList;
+
+    return result;
 };
 defineExpose({
     acceptParams,
```
wanghe-fit2cloud, line 654: There doesn't appear to be any apparent coding errors or major issues within the provided code snippet. It contains functions that seem designed to parse and manage information from dialog data objects, including handling quotes appropriately during string representations.

However, for optimization, you could implement error checking around some conditional logic to handle scenarios more precisely rather than relying on default behavior which can lead to unexpected behaviors under certain conditions, such as `indexof` method usage not being robust against strings containing white spaces.

For future reference, it might also help to include comments explaining your intention behind each function, especially when dealing with complex structures like those mentioned here.

wanghe-fit2cloud (commented):
/lgtm

Issue #7284, opened 2024-12-06
[Bug] Improve password action buttons styling in database connection info


ruibaby:
### 1Panel Version

dev branch(b73dcc9388805c1e1f61ad757da226e9060ab59a)

### Problem Description

Found two styling issues with password field buttons:

1. Both Copy and Random buttons highlight simultaneously on hover
2. Missing separator between adjacent buttons

<img width="426" alt="image" src="https://github.com/user-attachments/assets/c1987484-23be-4302-ae44-ad41d52a95a1">
<img width="417" alt="image" src="https://github.com/user-attachments/assets/25359d73-6eb2-4233-8c2b-5e5412bd628c">

### Steps to Reproduce

Database -> Connection info

wanghe-fit2cloud:
Thanks for your feedback. We will optimize this in an upcoming version.

wanghe-fit2cloud:
The v1.10.22 version has been released.

PR #7309, opened 2024-12-10
style: Improve password action buttons styling in database connection info

lan-yonghui:
Refs #7284

### DIFF
--- a/frontend/src/views/database/mysql/conn/index.vue
+++ b/frontend/src/views/database/mysql/conn/index.vue
@@ -81,9 +81,11 @@
                             prop="password"
                         >
                             <el-input type="password" show-password clearable v-model="form.password">
+                                <template #suffix>
+                                    <CopyButton type="icon" :content="form.password" class="w-30" />
+                                </template>
                                 <template #append>
-                                    <CopyButton :content="form.password" />
-                                    <el-button @click="random" class="p-ml-5">
+                                    <el-button @click="random">
                                         {{ $t('commons.button.random') }}
                                     </el-button>
                                 </template>
--- a/frontend/src/views/database/postgresql/conn/index.vue
+++ b/frontend/src/views/database/postgresql/conn/index.vue
@@ -82,9 +82,11 @@
                             prop="password"
                         >
                             <el-input type="password" show-password clearable v-model="form.password">
+                                <template #suffix>
+                                    <CopyButton type="icon" :content="form.password" class="w-30" />
+                                </template>
                                 <template #append>
-                                    <CopyButton :content="form.password" />
-                                    <el-button @click="random" class="p-ml-5">
+                                    <el-button @click="random">
                                         {{ $t('commons.button.random') }}
                                     </el-button>
                                 </template>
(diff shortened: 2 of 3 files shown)


frontend/src/views/database/mysql/conn/index.vue
```diff
@@ -81,9 +81,11 @@
...
+                                <template #suffix>
+                                    <CopyButton type="icon" :content="form.password" class="w-30" />
+                                </template>
                                 <template #append>
-                                    <CopyButton :content="form.password" />
-                                    <el-button @click="random" class="p-ml-5">
+                                    <el-button @click="random">
                                         {{ $t('commons.button.random') }}
                                     </el-button>
                                 </template>
```
wanghe-fit2cloud, line 91: Since the code provided seems incomplete and not specific to a particular platform (e.g., Web, mobile app), I'm unable to determine any potential problems, optimization suggestions, or knowledge cutoffs related to it. Could you please share more details about where this piece of code is located?

frontend/src/views/database/postgresql/conn/index.vue
```diff
@@ -82,9 +82,11 @@
...
+                                <template #suffix>
+                                    <CopyButton type="icon" :content="form.password" class="w-30" />
+                                </template>
                                 <template #append>
-                                    <CopyButton :content="form.password" />
-                                    <el-button @click="random" class="p-ml-5">
+                                    <el-button @click="random">
                                         {{ $t('commons.button.random') }}
                                     </el-button>
                                 </template>
```
wanghe-fit2cloud, line 92: The provided code seems to be related to an email input field that contains password information using Vue.js. However, there is no specific issue mentioned in your request; it's just a generic comparison regarding text content.

For more context on the differences:

1. The code snippet starts with a template tag `<template>`. This suggests it may represent the actual HTML code within some templates (for instance, if you have used Vue Router to navigate through layouts).
   
2. `el-input` element from ECharts has been replaced by `<el-input>` which also uses El-Echarts library. The difference lies in the CSS classes of these elements (`el-input` vs. `<el-input>`) but this doesn't directly involve any significant change or irregularity compared to standard Vuejs components.

Therefore, no further details can be added without additional context about where and how this code might run. I would recommend providing more specifics if available so I could better understand and assist with any potential issues or optimizations you're facing while working with this code.

frontend/src/views/database/redis/conn/index.vue
```diff
@@ -68,9 +68,11 @@
...
+                            <template #suffix>
+                                <CopyButton type="icon" :content="form.password" class="w-30" />
+                            </template>
                             <template #append>
-                                <CopyButton :content="form.password" />
-                                <el-button @click="random" class="p-ml-5">
+                                <el-button @click="random">
                                     {{ $t('commons.button.random') }}
                                 </el-button>
                             </template>
```
wanghe-fit2cloud, line 78: The code appears to be for an interface used in frontend applications.

Regarding potential issues:

- Missing `type` attribute in the `<CopyButton>` tag. A missing "type" (e.g., `"font"` or `"image"`) value could potentially break the button behavior. 

Optimization suggestion regarding `random()`: 
Since you use it multiple times in different cases, you might want consider putting it into a reusable function (`doSomething()`), or extracting its usage out of form submission process entirely. This would not only simplify the code but also allow for easier testing and modification. Also, using async operations may increase performance especially when doing many iterations without delay.

wanghe-fit2cloud (commented):
/lgtm

PR #7313, opened 2024-12-10
feat: Fix Missing CPU and Memory Limit Warnings When Editing Installe…

zhengkunwang223:


frontend/src/views/app-store/installed/detail/index.vue
```diff
@@ -163,13 +182,32 @@ const paramModel = ref<any>({
...
+    limits.value = res.data;
+    limits.value.memory = Number((limits.value.memory / 1024 / 1024).toFixed(2));
+    oldMemory.value = limits.value.memory;
+};
 
 const acceptParams = async (props: ParamProps) => {
+    loadLimit();
     submitModel.value.installId = props.id;
     params.value = [];
     paramData.value.id = props.id;
```
wanghe-fit2cloud, line 213: In the given snippet of code, there seem to be a few areas that could be improved. Some differences and improvements suggested:

1. In `submitModel`, instead of defining both `CPU` and `MEMORY_LIMIT` with same validations and rules under a single key (`params`), consider using different keys as you may have specific constraints on CPU Quota vs. Memory Limit.

```typescript
// ... other definitions...
const submitModel = ref<{[key: string]: any}>({cpuQuota: {}, memoryLimit: {...}}); // use proper object keys for distinct validation & rule sets

```
   
2. The `extensions` array is imported twice (once before extending codemirror and again after). You can import them only once at the beginning so we don't accidentally re-import them unnecessarily later.
3. In JavaScript language definition, it would help if these were properly formatted and structured like they should be per best practices.
4. The documentation isn’t very clear about why `codeMirrorInstance` was mentioned; it seems out-of-line from rest of the code section. Consider placing it at its respective place closer to where it will be used within the component(s).
5. There's some duplication across multiple places which might confuse others reading through this codebase. Ensure each instance has appropriate imports/uses.
6. Lastly, make note of missing `<script lang...>` tags when importing external libraries/components. This prevents errors related to incorrect file inclusion in TypeScript files. 

For optimization suggestion:
Use the correct syntax when importing React components from another folder into typescript to prevent error messages during compiles. For example:

```typescript
import { EditorState } from "@vue/reactivity";
```

Replace "EditorState" in the above line with your actual component name.

This helps keeping your .tsx code more tidy without worrying about unnecessary complexity introduced due to typing mismatching between ts and js codes leading to unexpected runtime errors.

wanghe-fit2cloud (commented):
/lgtm

Issue #7370, opened 2024-12-16
[Bug] There are front-end display issues when the 1Panel App Store is not loaded


Anyexyz:
### Contact Information

Anye

### 1Panel Version

Pro v1.10.21-lts

### Problem Description

There are front-end display issues when the 1Panel App Store is not loaded.
Like this
<img width="539" alt="3800f4e988f7d1c81f90f7a224ffb735" src="https://github.com/user-attachments/assets/9b4baad2-dbf1-494d-ada7-e8f56578e713" />

### Steps to Reproduce

After installing 1Panel, disconnect from the Internet and the front-end bug can be triggered stably.

wanghe-fit2cloud:
Thanks for your feedback! We’ll optimize this in an upcoming version.

wanghe-fit2cloud:
v1.10.22 版本已发布。

PR #7374, opened 2024-12-16
feat: Fix Application Store Style Issue Under Poor Network Conditions

zhengkunwang223:
Refs https://github.com/1Panel-dev/1Panel/issues/7370

### DIFF
--- a/backend/app/service/app.go
+++ b/backend/app/service/app.go
@@ -126,11 +126,8 @@ func (a AppService) PageApp(req request.AppSearch) (interface{}, error) {
 }
 
 func (a AppService) GetAppTags() ([]response.TagDTO, error) {
-	tags, err := tagRepo.All()
-	if err != nil {
-		return nil, err
-	}
-	var res []response.TagDTO
+	tags, _ := tagRepo.All()
+	res := make([]response.TagDTO, 0)
 	for _, tag := range tags {
 		res = append(res, response.TagDTO{
 			Tag: tag,


backend/app/service/app.go
```diff
@@ -126,11 +126,8 @@ func (a AppService) PageApp(req request.AppSearch) (interface{}, error) {
...
-	tags, err := tagRepo.All()
-	if err != nil {
-		return nil, err
-	}
-	var res []response.TagDTO
+	tags, _ := tagRepo.All()
+	res := make([]response.TagDTO, 0)
 	for _, tag := range tags {
 		res = append(res, response.TagDTO{
 			Tag: tag,
```
wanghe-fit2cloud, line 133: The code snippet has several improvements:

1. **Removed Unnecessary Error Handling**: The line `return nil, err` is unnecessary because the `_` variable already captures any error from the call to `tagRepo.All()`.

2. **Optimized Loop Initialization**: Instead of creating a slice with a fixed size and then appending elements, you can directly initialize a new slice and append only needed values during iteration. This avoids the initial allocation size overhead if the loop iterates less than twice.

3. **Use Slice Literal Syntax**: It's cleaner and more idiomatic to use the slice literal syntax `[0]` (`make([]response.TagDTO, 0)`).

4. **Consistent Variable Naming**: While not always necessary for readability, it might be good practice to ensure consistent naming conventions.

Here’s an optimized version of the function:

```go
func (a AppService) GetAppTags() ([]response.TagDTO, error) {
	tags, _ := tagRepo.All()
	res := make([]response.TagDTO, len(tags))
	for i, tag := range tags {
		res[i] = response.TagDTO{
			Tag: tag,
		}
	}
	return res, nil
}
```

These changes improve performance slightly by reducing memory allocations at initialization and making the logic easier to read and maintain.

wanghe-fit2cloud (commented):
/lgtm

PR #7375, opened 2024-12-16
fix: Fixed an issue where the language pack initialization download f…

ssongliu:


backend/utils/http/get.go
```diff
@@ -2,21 +2,30 @@ package http
...
+			TLSHandshakeTimeout:   5 * time.Second,
+			ResponseHeaderTimeout: 10 * time.Second,
+			IdleConnTimeout:       15 * time.Second,
+		},
 	}
-	transport := xpack.LoadRequestTransport()
-	client.Transport = transport
 
 	req, err := http.NewRequestWithContext(context.Background(), "GET", url, nil)
 	if err != nil {
```
wanghe-fit2cloud, line 31: The provided code has several improvements and fixes:

1. **InsecureSkipVerify**: This is set to `true`, which means the SSL certificate will not be verified. Consider setting it to `false` when working with production environments.

2. **Custom Dialer Settings**: The connection timeout values have been set appropriately (60 seconds), but additional checks might be needed based on specific requirements.

3. **Response Header Timeout**: Setting the response header timeout to 10 seconds seems appropriate if long headers are anticipated.

4. **Idle Connection Timeout**: This ensures idle connections do not remain open indefinitely, optimizing resource usage.

5. **Simplified Error Handling**: No error handling for TLS handshake or network timeouts is implemented. Adding these can help catch unexpected failures faster.

6. **TLS Configuration**: Additional TLS configurations can be added as needed for other security settings.

7. **Code Cleanup**: Removed unnecessary imports from external packages, ensuring only necessary ones are used in the context of this function.

8. **Documentation Update**: Mentioning potential side effects due to using an insecure skip verify setting might improve clarity.

Here's the revised snippet with some of these considerations:

```go
import (
	"context"
	"crypto/tls"
	"errors"
	"net"
	"net/http"
	"strings"

	"github.com/1Panel-dev/1Panel/backend/buserr"
)

func GetHttpRes(url string) (*http.Response, error) {
	client := &http.Client{
		Timeout:      time.Second * 300,
		TLSClientConfig: &tls.Config{
			InsecureSkipVerify: false, // Set to false in production
			Certificates:       []tls.Certificate{}, // Add certificates if required
		},
		Transport: &http.Transport{
			DialContext: (&net.Dialer{
				Timeout:   60 * time.Second,
				KeepAlive: 60 * time.Second,
			}).DialContext,
			TLSHandshakeTimeout:   5 * time.Second,    // Reduce if short responses expected
			ResponseHeaderTimeout: 10 * time.Second,  // Adjust based on expected headers
			MaxIdleConnsPerHost:  10,                // Optimize based on server expectations
			IdleConnTimeout:       15 * time.Second,  // Adjust to balance performance with resource management
		},
	}

	req, err := http.NewRequestWithContext(context.Background(), "GET", url, nil)
	if err != nil {
		return nil, errors.Wrap(err, "failed to create HTTP request")
	}

	resp, err := client.Do(req)
	if err != nil {
		return nil, errors.Wrap(err, "failed to make HTTP request")
	}

	return resp, nil
}
```

This version addresses the main concerns while maintaining the integrity of the original functionality.

backend/init/lang/lang.go
```diff
@@ -127,7 +127,7 @@ func downloadLangFromRemote(fileOp files.FileOp) {
 func downloadGeoFromRemote(fileOp files.FileOp, targetPath string) {
 	_ = os.MkdirAll(path.Dir(targetPath), os.ModePerm)
 	pathItem := fmt.Sprintf("%s/geo/GeoIP.mmdb", global.CONF.System.RepoUrl)
-	if err := fileOp.DownloadFileWithProxy(pathItem, targetPath); err != nil {
+	if err := fileOp.DownloadFile(pathItem, targetPath); err != nil {
 		global.LOG.Errorf("download geo ip failed, err: %v", err)
 		return
 	}
```
wanghe-fit2cloud, line 133: There are no irregularities or significant performance issues detected in the provided code snippet. The changes to `downloadLangFromRemote` and `downloadGeoFromRemote` functions from using `DownloadFileWithProxy` to simply `DownloadFile` are straightforward and should not introduce new bugs or require further optimizations.

The only change made here is updating the method signature of calling the `fileOp.DownloadFile` function with an additional argument representing the URL instead of a proxy URL. This simplifies the call since we don't need to pass the proxy configuration anymore.

wanghe-fit2cloud (commented):
/lgtm

Issue #7380, opened 2024-12-16
[Bug] Folder naming capitalization issue when manually creating an arrangement


Anyexyz:
### Contact Information

Anye

### 1Panel Version

Pro： v1.10.21-lts

### Problem Description

When manually creating an orchestration, the folder name contains uppercase letters, which will create two orchestration results (one of which is wrong). The final generated folder name is uppercase (normal), and the container starts normally.
![45d6961933cf9223f230dcfe04d11a97](https://github.com/user-attachments/assets/f938756d-8276-4496-bc05-4754bd41342a)
![42e7bf37f72ff550f3f0d0f3a06144f9](https://github.com/user-attachments/assets/5ce66c39-4a3b-4c13-b46e-9c0225d88e2e)
![7b8948cbe5c5762ba04d01f987811d21](https://github.com/user-attachments/assets/c58593bc-9833-49e0-8b6e-849bbc3d5157)

### Steps to Reproduce

Manually create an orchestration, fill in `AutoBangumi` for the folder name, and fill in
```yaml
networks:
1panel-network:
external: true
services:
AutoBangumi:
image: "ghcr.io/estrellaxd/auto_bangumi:latest"
container_name: AutoBangumi
volumes:
- ./config:/app/config
- ./data:/app/data
ports:
- "7892:7892"
networks:
- 1panel-network
restart: unless-stopped
environment:
- TZ=Asia/Shanghai
```
You can reproduce it

wanghe-fit2cloud:
Thanks for your feedback. We will fix this issue in the next version.

wanghe-fit2cloud:
The v1.10.22 version has been released.

PR #7382, opened 2024-12-17
fix: Fix the issue of compose name validation errors

ssongliu:
Refs #7380


frontend/src/global/form-rules.ts
```diff
@@ -644,6 +658,11 @@ export const Rules: CommonRule = {
...
         trigger: 'blur',
     },
+    composeName: {
+        required: true,
+        validator: checkComposeName,
+        trigger: 'blur',
+    },
     volumeName: {
         required: true,
         validator: checkVolumeName,
```
wanghe-fit2cloud, line 668: The code does not contain significant irregularities or major issues, but there is one suggestion for further improvement:

```javascript
// Replace the existing reg with a more specific regex pattern to allow only alphanumeric characters starting with a letter.
const reg = /^[a-zA-Z][\w-]{0,256}$/;
```

This change enhances the regular expression by using `[\w-]` instead of just `[a-z0-9_]`, which allows underscores and hyphens while ensuring that the string starts with an alphabetic character. This should prevent any invalid composition names due to special characters at the beginning of the input.

Optimization: The current implementation includes multiple conditional checks (`if`) and callbacks throughout the validation logic. You could potentially refactor this into fewer conditions by combining checks where needed. For example:
1. Ensure non-empty values and handle `null` or `undefined` cases centrally early in each rule function.
2. Combine similar logic in the validate functions where feasible.

Here's an updated version with these optimizations considered:

```javascript
const checkEmpty = (rule: any, value: any, callback: any) => {
    if (!value || typeof value === 'undefined' || value == null) {
        return new Error(i18n.global.t('commons.rule.empty'));
    }
};

const validRegex = /^(?!.*[^\w-])([a-zA-Z][-\w\d]){1,256}$/;

const commonRules = {
    image_name: {
        ...checkEmpty,
        message: i18n.global.t('commons.rule.imageName'),
        trigger: ['input', 'blur'],
        validator: (rule: any, value: any, callback: any) => {
            const reg = validRegex;
            if (reg.test(value) && value !== '') {
                callback();
            } else {
                callback(new Error(rule.message));
            }
        },
    },
    compose_name: {
        ...checkEmpty,
        message: i18n.global.t('commons.rule.composeName'),
        trigger: ['input', 'blur'],
        validator: (rule: any, value: any, callback: any) => {
            if (validRegex.test(value)) {
                callback();
            } else {
                callback(new Error(rule.message));
            }
        },
    },
    volume_name: {
        ...checkEmpty,
        message: i18n.global.t('commons.rule.volumeName'),
        trigger: ['input', 'blur'],
        validator: (rule: any, value: any, callback: any) => {
            if (validRegex.test(value)) {
                callback();
            } else {
                callback(new Error(rule.message));
            }
        },
    },
};
```
In this revised approach:
- A helper function `commonRules` stores configuration objects that apply uniformly across all fields; thus reducing redundancy.
- All validations now use the same basic structure (`required`, `validator`) making them easier to manage and less prone to errors when refactoring.
- Combined conditions within `validator` methods reduce unnecessary repetition.

frontend/src/lang/modules/en.ts
```diff
@@ -195,6 +195,8 @@ const message = {
             dbName: `This field mustn't start with underscore character and must consist of English, numbers, and "_" characters with a length of 1-64.`,
             imageName:
                 'This field must consist of English, numbers, ":", "@", "/", ".", "-", and "_" characters with a length of 1-256.',
+            composeName:
+                'Supports non-special characters at the beginning, lowercase letters, numbers, - and _, length 1-256',
             volumeName:
                 'This field must consist of English, numbers, ".", "-", and "_" characters with a length of 2-30.',
             supervisorName:
```
wanghe-fit2cloud, line 202: The changes appear to correct some validation conditions for specific fields in an object literal:

1. **`dbName`**: The message now mentions that underscores cannot be used in the string, which was previously incorrect.
2. **`composeName`**: This field is new and appears to allow for more flexible names with support for non-special characters at the start.

While these corrections enhance clarity regarding allowed character sets and constraints, they might not have been necessary if the previous version already covered those aspects adequately. It would also help to ensure consistency across similar fields by applying a standard pattern or using a regex for validation.

frontend/src/views/container/compose/create/index.vue
```diff
@@ -169,7 +169,7 @@ const form = reactive({
     envFileContent: `env_file:\n  - 1panel.env`,
 });
 const rules = reactive({
-    name: [Rules.requiredInput, Rules.imageName],
+    name: [Rules.requiredInput, Rules.composeName],
     path: [Rules.requiredInput],
     template: [Rules.requiredSelect],
 });
```
wanghe-fit2cloud, line 175: There is an error in this code snippet. The rule for "composeName" should be imported properly, since `Rules.composeName` seems like it's misspelled or not defined correctly. It appears that there may have been a typo when defining the rules array.

To fix this issue:

1. Ensure that the correct import statement exists at the beginning of your module where the code is located (e.g., if you're using ES modules): 
   ```javascript
   import { requiredInput, imageName } from './rules'; // Adjust the path as necessary
   ```

2. Change the `Rules.composeName` to use either the proper spelling (`requiredComposeName`) or the correct import statement:

Here’s how you can address it:

```javascript
import {
  requiredInput,
  imageName,
  requiredComposeName // Add requiredCompositeName here if needed
} from './rules'; // Adjust the path as necessary

const form = reactive({
  envFileContent: `env_file:\n  - 1panel.env`,
});

const rules = reactive({
  name: [requiredInput, requiredComposeName], // Use correct rule from imported ruleset
  path: [requiredInput],
  template: [requiredSelect],
});
```

This change ensures that the component is using valid and correctly named rules, thus improving both functionality and readability in the application.

wanghe-fit2cloud (commented):
/lgtm

PR #7395, opened 2024-12-17
fix: Improve cache cleaning for internationalization

ssongliu:


frontend/src/views/toolbox/clean/index.vue
```diff
@@ -641,12 +639,20 @@ function load18n(label: string) {
...
+        case 'directory':
+        case 'log':
+        case 'cutWebsiteLog':
+        case 'clean':
+        case 'snapshot':
+        case 'ntp':
+            return i18n.global.t('cronjob.' + label);
         case 'container_images':
             return i18n.global.t('clean.images');
         case 'container_containers':
```
wanghe-fit2cloud, line 658: The code appears to have several logical inconsistencies:

1. **Duplicate Case Statements**: The `case 'snapshot'` is duplicated without a comment explaining the purpose. This likely indicates an error that should be corrected.

2. **Redundant Cases**: Some cases appear redundant, such as `"container_shell"` and `"curl"`, which are also present under different labels but do not match each other exactly.

3. **Inconsistent Pattern Matching**: There seems to be pattern matching based on substrings of the label rather than strict matches (`label.includes()`). This might lead to unexpected behavior if the order or structure of strings changes.

### Suggested Optimizations and Improvements

1. **Remove Duplicated Case Statement**:
   ```javascript
   case 'snapshoot': // Note: Typo fixed here
       return i18n.global.t('clean.snapshot');
   ```

2. **Clean Up Redundancies**:
   - Remove `"container_shell"` and `"curl"` since they already cover similar functionalities with different prefixs (e.g., `$SHELL`).

3. **Use Strict Matches for Consistency**:
   If you want to strictly match against specific patterns, consider using regular expressions or more explicit string comparisons instead of substring matching.

Here's what the optimized version could look like:

```javascript
function load18n(label: string) {
    switch (label.toLowerCase()) { // Convert label to lowercase for consistent comparison
        case "cache":
            return i18n.global.t("clean.cache");
        case "snapshot_tm":
            return i18n.global.t("clean.snapshotTmp");
        case "snapshot_local":
            return i18n.global.t("clean.snapshotLocal");
        
        case "docker_log":
        case "nginx_access": 
            return i18n.global.t("cleanup.log.");
        
        case "task_log":
           return i18n.global.t("clean.taskLog");

        case "container_image":
            return i18n.global.t("clean.images");
        case "container_container":
            return i18n.global.t("clean.containers");

        default:
            console.error(`Unknown clean action: ${label}`);
            return null;
    }
}
```

This approach ensures clearer logic and eliminates unnecessary duplication while maintaining consistency.

frontend/src/lang/modules/zh.ts
```diff
@@ -1666,9 +1665,7 @@ const message = {
         systemLogHelper: '系统日志信息、容器构建或镜像拉取等日志信息以及计划任务中产生的日志文件',
         dockerLog: '容器操作日志文件',
         taskLog: '计划任务执行日志文件',
-        shell: 'Shell 脚本计划任务',
-        containerShell: '容器内执行 Shell 脚本计划任务',
-        curl: 'CURL 计划任务',
+        containerShell: '容器内 Shell 脚本',
 
         containerTrash: '容器垃圾',
         images: '镜像',
```
wanghe-fit2cloud, line 1671: The provided code changes appear to be small edits without significant impact on the overall structure or functionality. The primary change is replacing `snapshot` with `snapshotTmp`, which likely represents temporary files used during snapshot upload.

**Irregularities/Issues**: None spotted.

**Optimization Suggestions**: 
- Ensure that file names and identifiers remain consistent throughout the project, especially when they refer to similar concepts (e.g., renaming `snapshot` to another variable name consistently).
- Double-check if any logic depends on the presence of these files, ensuring that the new naming does not inadvertently affect behavior.
- Verify that there are no existing references to the old `snapshot` path that would need updating along with the rename.

Overall, this change should have a minimal effect but good practices suggest consistency in naming conventions across the codebase.

frontend/src/lang/modules/en.ts
```diff
@@ -1784,9 +1783,7 @@ const message = {
             'System log information, container build or image pull log information, and log files generated in scheduled tasks',
         dockerLog: 'Container operation log files',
         taskLog: 'Scheduled task execution log files',
-        shell: 'Shell script scheduled tasks',
         containerShell: 'Container internal Shell script scheduled tasks',
-        curl: 'CURL scheduled tasks',
 
         containerTrash: 'Container Trash',
         images: 'Images',
```
wanghe-fit2cloud, line 1789: The changes appear to be intended to clean up and update translations related to system components. Here is a revised version of the code snippet:

```json
{
  "upgradeHelper": "(Recommended keeping the latest upgrade backup for system rollback)",
  "cache": "System cache files",
  "cacheHelper": "(Proceed with caution, cleaning requires a service restart)",
  "snapshotTmp": "System snapshot upload temporary files",
  "snapshotLocal": "System snapshot creation temporary files",
  "rollback": "Backup files before recovery",
  "logs": [
    "All relevant logs including system log information, container build or image pull log information, and log files generated in scheduled tasks"
  ],
  "dockerLog": "Container operation log files",
  "taskLog": "Scheduled task execution log files",
  "containerShell": "Container internal shell script scheduled tasks",
  "shell": "",
  
  "trash": "Container Trash",
  "images": "Images"
}
```

Changes made:
1. Replaced `curl` with an empty string since it seems that curl-related commands have been removed.
2. Combined the descriptions under the `"logs"` key into a single array for better readability.

wanghe-fit2cloud (commented):
/lgtm

PR #7423, opened 2024-12-18
style(website): Update Nginx Status Page Styling

zhengkunwang223:


frontend/src/views/website/website/nginx/status/index.vue
```diff
@@ -85,3 +60,9 @@ onMounted(() => {
     get();
 });
 </script>
+
+<style scoped>
+.el-col {
+    text-align: center;
+}
+</style>
```
wanghe-fit2cloud, line 68: No differences found that indicate any irregularities or potential issues with this template for displaying Nginx status metrics within a web page. It maintains its original structure and contents which could be suitable without making major changes since 2021.

However, if the requirement is to optimize it further:

- To make the elements more visually appealing by aligning them closer together (e.g. using `float` instead of positioning). This is not relevant to this specific case but it's usually beneficial for responsive design.
- Use CSS Flexbox or Grid for the layout to ensure all components have appropriate width/height while also respecting responsiveness.

The current structure works very well with existing content, so no major changes are needed at this time.

frontend/src/components/layout-content/index.vue
```diff
@@ -21,36 +21,32 @@
...
+                        <div v-else class="flex flex-wrap gap-4 sm:justify-between">
+                            <div class="flex gap-2 flex-wrap items-center justify-start">
+                                <slot name="leftToolBar" v-if="slots.leftToolBar"></slot>
+                                <slot name="buttons" v-if="slots.buttons"></slot>
+                            </div>
+                            <div class="flex flex-wrap gap-3" v-if="slots.rightToolBar || slots.rightButton">
+                                <slot name="rightToolBar"></slot>
+                                <slot name="rightButton"></slot>
                             </div>
                         </div>
```
wanghe-fit2cloud, line 52: The major changes between the two versions you provided occur at line `21`, which is causing an inconsistency in styling. Specifically, the `<div>` element inside a card has both rounded corners (using CSS for box-shadow) and borders around it using display flex with border-radius set to 4px instead of being specified directly on that same div.

To correct these discrepancies:

- Set the CSS properties (`box-shadow` and other relevant styles) within `.content-container__title`.
- Remove unnecessary margin property from `.content-container__title`. The content containers should be responsive if needed, not always have margins.
- Ensure all instances of `.content-container .btn-primary.btn-danger` are correctly aligned without padding.
- Update back-to-front styling consistently across various components to maintain a consistent layout structure throughout this module.

These adjustments will help improve the coherence and look of the overall design.

agent/app/service/nginx.go
```diff
@@ -94,13 +95,34 @@ func (n NginxService) GetStatus() (response.NginxStatus, error) {
...
+	writing, err := strconv.Atoi(resArray[13])
+	if err == nil {
+		status.Writing = writing
+	}
+	waiting, err := strconv.Atoi(resArray[15])
+	if err == nil {
+		status.Waiting = waiting
+	}
 	return status, nil
 }
```
wanghe-fit2cloud, line 128: There are two main points of discrepancy in this code:

In the function `Nginx`, there was an expected return type (`ResponseNginxStatus`) instead of `(error)` in line that reads 

```makefile
status := response.GetServerStats()
```

This leads to a different return type in comparison between previous versions.

To resolve this inconsistency, you may want to update the function with the correct return statement or adjust it accordingly. Remember to keep your functions consistent so they all have the same return structure.

wanghe-fit2cloud (commented):
/lgtm

PR #7439, opened 2024-12-19
feat: merge code from dev

ssongliu:


agent/app/service/firewall.go
```diff
@@ -308,21 +308,58 @@ func (u *FirewallService) OperateForwardRule(req dto.ForwardRuleOperate) error {
...
 	sort.SliceStable(req.Rules, func(i, j int) bool {
+		if req.Rules[i].Operation == "remove" && req.Rules[j].Operation != "remove" {
+			return true
+		}
+		if req.Rules[i].Operation != "remove" && req.Rules[j].Operation == "remove" {
+			return false
+		}
 		n1, _ := strconv.Atoi(req.Rules[i].Num)
 		n2, _ := strconv.Atoi(req.Rules[j].Num)
 		return n1 > n2
```
wanghe-fit2cloud, line 365: This code does not contain any identified errors. However, there is an improvement to optimize it:

Instead of sorting `rules` and then comparing the numbers after sorting which increases time complexity from O(n log n), we can directly compare both elements using equality operator. 

Additionally, you might want to use string replace instead of substring if target port has "/" character. It would make the function more readable.

Here's how I suggest optimizing it:

```go
func (u *FirewallService) OperateForwardRule(req dto.ForwardRuleOperate) error {

  rules, _ := client.ListForward()

  var filteredRules []dto.ForwardRule

  for i := 0; i < len(rules); i++ {
    // Remove empty Target IP
    if !strings.ContainsAny(req.Rules[i].TargetIP, ".:") {
      continue
    }

    ruleToRemove, exists := findRuleInList(i, rules[i])

    if exists && filterRules(r[ruleIndex]) {
      filteredRules = append(filteredRules, r[ruleIndex])
      i-- /* go one less index */
      delete(rules, i)
    }
  }

  fmt.Println("Number of Rules before filtering:", len(rules))
  fmt.Println("Number of Rules after filtering:", len(filteredRules))

  return nil
}

```

And here the method `filterRules()` can be implemented similar to that shown below. Note this changes depend on your specific implementation details so please adjust accordingly based on what you're expecting.

```golang
func filterRules(rule ForwardRule) bool {
	return rule.Operator == "add" ||
		rule.Operation == "changeValue" &&
			strings.HasSuffix(rule.NewValue, "/tcp") &&
			strings.HasPrefix(rule.OldName, "/") &&
			len(strings.Fields(rule.OldName)) == 6 &&
			re.MatchString(strings.Replace(allProtocols[r.policy], "/", "\\/", -1), rule.Port)
}
```
For all operations (add/replace/remove) in forward rule:

```go
for i := range req.Rules {
    if strings.Compare(filters[r][i-1].OldName(), newValues[filters[r][i][j]]) != 0 
        || re.FindAllIndex(newValues[filters[r][i][j]], filters[filters[r][i-1]][i+j].NewKey(), -1)[0] >= len(newValues) {
            err := u.AddOrUpdateForwardRule(&dto.FirewallServiceAddOrUpdateForward{RequestPolicy: requestPolicy.ID(),
                Operation:   constants.Add,
                Name:       r.Name,
                Policy:     r.Policy,
                OldValues:  dts.FromMapToStruct(dts.Value(r),
                    map[string]string{
                        "name":              r.Name,
                        "operation":          constants.Update,
                        "old_name":           "",
                        "target_ip":          r.TargetIP,                        
                        "protocol":  protocolFromStr(r.Protocol),
                        "port":               strings.TrimSpace(strconv.Itoa(int(float64(r.Port))), ""),
                     })).Update(requestData.CreateEntity().(*dto.EntityCreate)).Execute().Err()
             if err != nil {
                 return
             } 
    }
}

```

frontend/src/views/app-store/index.vue
```diff
@@ -35,6 +48,26 @@ const buttons = [
...
+            canUpdate.value = false;
+        })
+        .finally(() => {
+            syncing.value = false;
+        });
+};
+
 const search = () => {
     SearchAppInstalled({ update: true, page: 1, pageSize: 100 })
         .then((res) => {
```
wanghe-fit2cloud, line 73: The given code seems to be related to searching or updating installed applications using an API. There doesn't seem to be any glaring inconsistencies, potential issues, or optimizations that need immediate consideration.

However, there's one suggestion I would make – consider refactoring the code structure to make it more clear and easier to follow. For instance:

```typescript
import i18n from '@/lang';
```

Could be placed below `const messages` or at the top of the file:

```typescript

import { SearchAppInstalled } from '@/api/modules/app';

const sync = () =>
{
//...
}
```

To improve usability, you might also add some documentation explaining why this line does what it does:

```bash
import { SearchAppInstalled } from '<YourComponentName>'

let showButton = useRef(true);

const searchApps = async() => {...} ;
```
It helps users understand how each variable / function relates to other parts of the component/UI without having to refer back extensively to the actual source file which can lead to confusion sometimes. This is especially valuable as versioning evolves over time.

agent/app/service/dashboard.go
```diff
@@ -518,3 +529,15 @@ func loadXpuInfo() []dto.XPUInfo {
...
+	conn, err := network.Dial("udp", "8.8.8.8:80")
+
+	if err != nil {
+		return "IPNotFound"
+	}
+	defer conn.Close()
+
+	localAddr := conn.LocalAddr().(*network.UDPAddr)
+	return localAddr.IP.String()
+}
```
wanghe-fit2cloud, line 543: I apologize for the confusion earlier but currently I am unable to execute or review any code snippets. Please try again later when you have actual source code provided along with some specific queries about it.

wanghe-fit2cloud (commented):
/lgtm

wanghe-fit2cloud (commented):
/lgtm

PR #7458, opened 2024-12-20
fix(system): Fix Cloudflare Proxy Access Issue with 1Panel When Secur…

zhengkunwang223:
Refs https://github.com/1Panel-dev/1Panel/issues/7457

### DIFF
--- a/backend/init/router/router.go
+++ b/backend/init/router/router.go
@@ -28,9 +28,9 @@ var (
 )
 
 func toIndexHtml(c *gin.Context) {
+	c.Writer.Header().Set("Content-Type", "text/html; charset=utf-8")
 	c.Writer.WriteHeader(http.StatusOK)
 	_, _ = c.Writer.Write(web.IndexByte)
-	c.Writer.Header().Add("Accept", "text/html")
 	c.Writer.Flush()
 }
 
@@ -131,9 +131,9 @@ func setWebStatic(rootRouter *gin.RouterGroup) {
 				handleNoRoute(c)
 				return
 			}
+			c.Writer.Header().Set("Content-Type", "text/html; charset=utf-8")
 			c.Writer.WriteHeader(http.StatusOK)
 			_, _ = c.Writer.Write(web.IndexByte)
-			c.Writer.Header().Add("Accept", "text/html")
 			c.Writer.Flush()
 		})
 	}


backend/init/router/router.go
```diff
@@ -131,9 +131,9 @@ func setWebStatic(rootRouter *gin.RouterGroup) {
 				handleNoRoute(c)
 				return
 			}
+			c.Writer.Header().Set("Content-Type", "text/html; charset=utf-8")
 			c.Writer.WriteHeader(http.StatusOK)
 			_, _ = c.Writer.Write(web.IndexByte)
-			c.Writer.Header().Add("Accept", "text/html")
 			c.Writer.Flush()
 		})
 	}
```
wanghe-fit2cloud, line 139: The given code appears to be part of a Gin web application. It defines two functions: `toIndexHtml` and `setWebStatic`. Here's a brief analysis:

### Code Differences and Potential Issues:

**File 1: main.go (or similar)**

```go
package main

import (
    "github.com/gin-gonic/gin"
    ...
)

func main() {
    router := gin.Default()

    // Set up static files routes
    setWebStatic(router)

    // Define other routes
    setupRoutes(router)

    // Start the server
    if err := router.Run(":8080"); err != nil {
        panic(err)
    }
}

// Function definitions...
```

**File 2: utils/web/handlers.go**

```go
var (

	IndexByte []byte = []byte("<html><body>Hello, World!</body></html>")
)

func toIndexHtml(c *gin.Context) {
	c.Writer.Header().Set("Content-Type", "text/html; charset=utf-8") // Added content type header
	c.Writer.WriteHeader(http.StatusOK)
	_, _ = c.Writer.Write(IndexByte)
	c.Writer.Flush()
}

func setWebStatic(rootRouter *gin.RouterGroup) {
	rootRouter.Use(func(c *gin.Context) { // Middleware to add headers
		if !c.Request.MethodAllowed() || len(c.Request.URL.Path) == 0 {
			c.JSON(405, gin.H{"error": "Method Not Allowed"})
			return
		}
		c.Set("requestPath", c.Request.URL.Path)
		c.Next()
	})
	rootRouter.Any("*/*", func(c *gin.Context) {
		parts := c.Request.URL.Path.split("/") // Assuming Path is a string function for splitting paths
		if len(parts) <= 1 {
			toIndexHtml(c) // Serve index page when no specific route matches
		} else {
			staticDir := "./www" // Adjust this path as needed
			file, err := ioutil.ReadFile(path.Join(staticDir, strings.Join(parts[1:], "/")))
			if err != nil {
				http.Error(c.Writer, err.Error(), http.StatusInternalServerError) // Handle file not found error
			 return
			}
			contentType := mime.TypeByExtension(filepath.Ext(file.Name()))
			c.Data(http.StatusOK, contentType, file)
			return
		}
	})

    // Additional middleware and routing logic can go here...
}

// Helper functions...
```

### Optimizations and Suggestions:

1. **Remove Redundant Content Type Header**: The line `c.Writer.Header().Add("Accept", "text/html")` can be removed since it doesn't serve any additional purpose except adding an unnecessary key-value pair.

2. **Middleware Implementation**: 
   - Use Gin middleware to centralize common request handling tasks like setting status codes and logging.
   - Consider using context variables instead of querying parameters within handler functions, especially in performance-critical parts of your app.

3. **Error Handling**:
   - When reading from the filesystem (`ioutil.ReadFile`) and serving data over HTTP, ensure you handle errors appropriately and log them for debugging purposes.

4. **Performance Improvements**:
   - If dealing with multiple requests per second, consider parallelizing file reads if applicable.
   - Pre-compute some constants or values that do not change during runtime.

Overall, the code looks well-structured and should work correctly for its intended purpose as long as all referenced packages are properly imported.

wanghe-fit2cloud (commented):
/lgtm

PR #7545, opened 2024-12-24
feat:  Add node service detection

ssongliu:


frontend/src/components/dialog-pro/index.vue
```diff
@@ -0,0 +1,69 @@
...
+
+const dialogVisible = computed({
+    get() {
+        return props.modelValue;
+    },
+    set(value: boolean) {
+        emit('update:modelValue', value);
+    },
+});
+</script>
```
wanghe-fit2cloud, line 69: Based on the provided snippet, here are some observations and suggested improvements:

1. **Size Property Naming**: `size` property is redundant since it's already stored as an instance variable named `size`.
2. **Slots Implementation**: The `<template>` tags do not match their `<slot>` directives, specifically because `<template>` elements should be placed inside the `<div>` element for inline slot processing instead of placing them outside `<slot>s`.
3. **Dialog Width Management**: Use `$attrs` to access the width directly when creating this component rather than setting via `.width="size"`.

Here's the refined code based on these points:

```html
<div class="drawer-pro" @click="$emit('close')">
    <div :class="'drawer-' ~ $attrs.width || "normal"' ref="drawer"> <!-- Adjusted line -->
      <!-- Content goes here -->
    </div> 
</div>
```

This version addresses most concerns listed above while maintaining a consistent syntax across all components within this package if it were included laterally, like in `el-container`.

agent/app/service/setting.go
```diff
@@ -59,3 +65,52 @@ func (u *SettingService) Update(key, value string) error {
...
+		return nil
+	}
+
+	global.CONF.System.BaseDir, _ = settingRepo.GetValueByKey("BaseDir")
+	global.CONF.System.Version, _ = settingRepo.GetValueByKey("SystemVersion")
+	global.CONF.System.EncryptKey, _ = settingRepo.GetValueByKey("EncryptKey")
+	global.CONF.System.CurrentNode, _ = settingRepo.GetValueByKey("CurrentNode")
+	global.CONF.System.MasterAddr, _ = settingRepo.GetValueByKey("MasterAddr")
+	return nil
+}
```
wanghe-fit2cloud, line 116: It seems that there are several issues with this code snippet.

**Issues**

1. **Type mismatches**
   - The first class definition (`SettingService`) does not correctly extend from the interface `ISettingService`.
   
2. **Interface methods and implementation of update method missing**
   - It is important to implement all the methods specified in the interface, otherwise it will lead to undefined behavior.

**Suggestions**

I would make these changes:

```go
type SettingService struct{}

// GetSettingInfo retrieves a setting object's information based on its key.
func (s *SettingService) GetSettingInfo(key string) (*App.GetSettingData, error) {
  // Implementation to retrieve an app SettingData
}

// Update updates an existing setting based on provided key-value pairs.
func (s *SettingService) Update(key string, value string) error {
  dataToUpdate, ok := App.GetSettings[key]
  if !ok {
    return fmt.Errorf("no such setting found")
  }

  // TODO: Implement logic to update database

  return nil
}
```

These implementations should fix the aforementioned bugs and improve readability and maintainability.

Also, don't forget to add docstrings around functions and variable comments. This makes your code more accessible and easier for others who may read / modify your work.

core/app/service/setting.go
```diff
@@ -208,20 +222,6 @@ func (u *SettingService) UpdatePort(port uint) error {
...
-			if len(oldMasterPort) != 0 {
-				if err := xpack.UpdateMasterAddr(strings.ReplaceAll(masterAddr.Value, oldMasterPort, fmt.Sprintf("%v", port))); err != nil {
-					global.LOG.Errorf("update master addr from db failed, err: %v", err)
-					return
-				}
-			}
-		}
 	}()
 	return nil
 }
```
wanghe-fit2cloud, line 227: There does not appear to be any obvious issues with this code snippet from an English perspective. It appears it checks if a provided port number is valid using a scan function and updates settings accordingly through `settingRepo`. However, since it only mentions reloading database information for updating the port, more context would provide more specific guidance on improvements.

wanghe-fit2cloud (commented):
/lgtm

Issue #7600, opened 2024-12-30
[Bug] Creating a new uppercase application will replace the original lowercase application


Anyexyz:
### Contact Information

Anye

### 1Panel Version

Pro v1.10.22-lts

### Problem Description

![image](https://github.com/user-attachments/assets/e93fe8ee-71ee-4617-afa5-6ac6410efcf4)
![image](https://github.com/user-attachments/assets/bafd40ca-4eea-436f-9b45-62f6338898ac)
Creating a new uppercase application will replace the original lowercase application.

### Steps to Reproduce

Install gitea in the app store, set the name to `gitea`, keep other configurations as default, and confirm the installation.
Install gitea again, set the name to `Gitea`, modify the port and other content so that there is no conflict, and confirm the installation.

Then you will find that both gitea applications are installed, but the first installed gitea stops running.

zhengkunwang223:
Thank you for your feedback. We will address this issue in a future version.

wanghe-fit2cloud:
v1.10.23-lts 版本已发布。

PR #7601, opened 2024-12-30
fix: Fix Application Startup Failure Caused by Case-Sensitive Applica…

zhengkunwang223:
Refs https://github.com/1Panel-dev/1Panel/issues/7600

### DIFF
--- a/frontend/src/views/app-store/detail/install/index.vue
+++ b/frontend/src/views/app-store/detail/install/index.vue
@@ -173,7 +173,7 @@ const installData = ref<InstallProps>({
 });
 const open = ref(false);
 const rules = ref<FormRules>({
-    name: [Rules.appName],
+    name: [Rules.paramSimple],
     params: [],
     version: [Rules.requiredSelect],
     containerName: [Rules.containerName],


frontend/src/views/app-store/detail/install/index.vue
```diff
@@ -173,7 +173,7 @@ const installData = ref<InstallProps>({
 });
 const open = ref(false);
 const rules = ref<FormRules>({
-    name: [Rules.appName],
+    name: [Rules.paramSimple],
     params: [],
     version: [Rules.requiredSelect],
     containerName: [Rules.containerName],
```
wanghe-fit2cloud, line 179: The difference in this code snippet is that the `appName` rule from `"Rules"` has been replaced with `"paramSimple"`. It's unclear what these rules are defined as, but replacing one with another could have different behaviors or constraints depending on their implementation. Ensure both "appName" and "paramSimple" rules behave similarly to avoid unexpected behavior in the application.

wanghe-fit2cloud (commented):
/lgtm

PR #7629, opened 2025-01-03
feat(appstore): Add MD5 Checksum for Custom Application Store Packages

zhengkunwang223:


core/utils/files/files.go
```diff
@@ -169,3 +171,17 @@ func Stat(path string) bool {
...
+		return "", err
+	}
+	defer file.Close()
+	hash := md5.New()
+
+	if _, err = io.Copy(hash, file); err != nil {
+		return "", err
+	}
+	return hex.EncodeToString(hash.Sum(nil)), nil
+}
```
wanghe-fit2cloud, line 187: The given code seems to be correct and there are no apparent issues. However, it's worth mentioning that the `md5` function from the standard library is not recommended due to security concerns. It uses MD4 algorithm which was replaced by MD5 in version 4 of the specification but some programs do not support this old version anymore. Instead of using this library, you can opt for an alternative approach like third-party libraries such as "crypto/tls" with its built-in MD5 hashing functionality.

Additionally, consider refactoring your imports if necessary to ensure proper version compatibility and reduce clutter.

Here are some suggestions:

1. Refactor imports (if needed)

2. Use different hashing algorithms

3. Implement proper type hints where appropriate

Remember that these changes would affect future code versions when merging new commits into production branches.

agent/app/service/app_install.go
```diff
@@ -414,25 +414,25 @@ func (a *AppInstallService) Update(req request.AppInstalledUpdate) error {
...
-				global.LOG.Errorf(buserr.WithErr(constant.ErrUpdateBuWebsite, err).Error())
+				global.LOG.Error(buserr.WithErr(constant.ErrUpdateBuWebsite, err).Error())
 				return
 			}
 			if err := nginxCheckAndReload(nginxInstall.SiteConfig.OldContent, config.FilePath, nginxInstall.Install.ContainerName); err != nil {
-				global.LOG.Errorf(buserr.WithErr(constant.ErrUpdateBuWebsite, err).Error())
+				global.LOG.Error(buserr.WithErr(constant.ErrUpdateBuWebsite, err).Error())
 				return
 			}
 		}()
```
wanghe-fit2cloud, line 438: Sorry, but you've shared a snippet of PHP code which isn't clear enough to analyze the differences between the two versions that you're contrasting it against your version. Could you please share more details on the current application context or additional information about the specific lines/diagrams being shown? That would help me accurately assess what parts need improvement or clarification.

PR #7631, opened 2025-01-03
feat: Increase version restrictions for node switching

ssongliu:


frontend/src/layout/components/Sidebar/index.vue
```diff
@@ -195,6 +220,7 @@ function getCheckedLabels(json: Node): string[] {
 
 const search = async () => {
     const res = await getSettingInfo();
+    version.value = res.data.systemVersion;
     const json: Node = JSON.parse(res.data.xpackHideMenu);
     if (json.isCheck === false) {
         json.children.forEach((child: any) => {
```
wanghe-fit2cloud, line 226: There seems to be two main sections where potential improvements are needed for clarity and consistency:

1. In each function (`changeNode`), there's a call "if (command == 'local')"
but it should probably look more like this: "`return;`"
because in JavaScript, you can usually just put `break;` at the bottom.

Also, regarding style and formatting suggestions:

```
import { elButton, SvgIcon } from '@/components/button';

```  
This could potentially be:
```
import IconButton from '@/components/button/IconButton';
/*...*/

...
<span>
{`${loadCurrentName()}`}
</span>    
```

frontend/src/utils/version.ts
```diff
@@ -0,0 +1,42 @@
...
+    }
+    if (start !== -1) {
+        numbers.push(version.slice(start));
+    }
+    return numbers;
+}
+
+function isDigit(char: string): boolean {
+    return /^\d$/.test(char);
+}
```
wanghe-fit2cloud, line 42: The code does not appear to have any immediate syntax errors. However, there are some improvements that could be made:

1. The `extractNumbers` function can be optimized by only extracting digits when they're on their own line instead of concatenating them with null strings.

Here's an updated version:
```javascript
function extractNumbers(version: string): string[] {
    const numbers: string[] = [];
    let startAtLastChar = false;

    for (let i = 0; i < version.length; i++) {
        const char = version[i];

        // If digit at current position matches the last non-digit character,
        // ensure it has its starting index set correctly before appending the number part.
        if (!startAtLastChar && isDigit(char)) {
            startAtLastChar = { startAtThisPosition: i };
                if (numbers[0]) {
                    console.log('StartAtFirstPos', startAtFirstPos)
            } else if (numbers && numbers[numbers.length-1].length === 0) {
                console.log('startOnNewLine')
                        numbers.push('');
            }
        }

        // Only append a number part when we've reached end of current line
        if (char !== ',' && !isDigit(char)) break;
        
        if (!startAtLastChar) continue;

        if (typeof startAtLastChar.startAtThisPosition == 'undefined') continue ;

```

To summarize the key improvements from these suggestions:
```lua
* Extract just one group of consecutive characters between commas as "digits" instead of trying to find groups of numbers across multiple lines or spaces 
 * Make sure that each number starts itself (on separate lines), ensuring you handle both leading zeros properly
    ```

File: frontend/src/lang/modules/en.ts [hunk 3]
```diff
@@ -1743,6 +1743,10 @@ const message = {
         upgradeNow: 'Upgrade now',
         source: 'Download source',
         hasNewVersion: 'New version Available',
+        versionHigher:
+            'Detected that node {0} version is higher than the main node, switching is not supported at this time. Please upgrade the main node system version and try again!',
+        versionLower:
+            'Detected that node {0} version is lower than the main node, switching is not supported at this time. Please upgrade the system version of this node and try again!',
 
         about: 'About',
         project: 'Project Address',
```
wanghe-fit2cloud, line 1752: Based on my knowledge up to 2021-09-01, the provided code appears fine with no apparent issues or need for changes. It seems quite standardized within its current context of a project addressing node versions among other things and offers clear options like upgrading now, downloading source files, etc. However, I would still recommend running it through some basic linting tools to ensure there might be specific improvements that can enhance its efficiency or legibility (if you're using this kind of code).

frontend/src/layout/components/Sidebar/index.vue
wanghe-fit2cloud, line 226: There is no notable difference in syntax, variable names, or function usage between the given code snippets. However, there might be more meaningful ways to optimize these scripts based on current best practices and coding standards such as using `@props`, `ref()`, `Vue.use()` etc. Please feel free to mention anything specific you want me to investigate. Also note that this check doesn't account for semantic changes like breaking APIs, new dependencies used instead of existing ones, or changes affecting how data structures are being manipulated.

The most important thing to note here is if there's any typo present which could potentially lead to bugs without user awareness. For instance, consider variables named `logOutApi`, they should ideally be renamed to avoid confusion with vue api (`axios`, `localStorage`) hence, it would become `logoutHandler` or similar.

Remember, while reviewing JS code, always ensure it adheres to current standards and security policies to make sure it performs as intended across environments including both production and testing phases.

frontend/src/utils/version.ts
wanghe-fit2cloud, line 42: The provided code snippet contains several inconsistencies and errors that require attention before being released. Here's an overview of these issues:

1. Improper handling of `isNaN` vs using conditional operators like ternary operator.
2. Lack of validation on input strings, ensuring they contain only alpha numeric characters instead of alphanumeric combinations.

For instance, to address these concerns, you would need to implement explicit validations such as checking the inputs for alphabetic/numeric values rather than relying solely on regular expression matching. 

Additionally, implementing basic type safety checks could prevent runtime errors during compile time.

Here’s a simplified approach with improved error handling and additional optimizations:

```typescript
import { strict as assert } from 'assert';

/**
 * Compares two version strings based on how numerically each number part should be compared when concatenated together,
 * considering negative sign prefixing as necessary.
 * 
 * @param version1 The first version string representing the major/minor/patch numbers of interest.
 * @param version2 The second version string also representing the same information.
 * @returns true if there is no mismatch between version parts (e.g., if both versions can be safely added directly),
 *          false otherwise due to invalid input or numerical anomalies leading to incorrect comparisons.
 */
function isValidCompareVersion(version1: string, version2: string): boolean {
    const partsA = extractPartsFromVersions(version1);
    const partsB = extractPartsFromVersions(version2);

    // Validate input types
    if (!Array.isArray(partsA) || !Array.isArray(partsB)) {
        throw new Error("Inputs must be arrays");
    }

    // Compare valid elements by concatenating them together
    for (const parts of [partsA, partsB]) {
        if (isValidElementsByCombiningElements(partsWithMajorMinorsToArrays(parts))) {
            return true; // No mismatches found between consecutive sections after combining their individual parts
        }
    }

    return false; // Some section(s) had mismatches
}

// Helper functions used in comparison logic

/**
 * Extracts all digit characters from a given string, returns empty array if none available.
 * @param str String containing digits to process.
 * @returns An array of extracted digits.
 */
function extractDigits(str: string): string[] {
    return str.match(/\d+/g)?.map((digit) => digit.trim()) ?? [];
}

function isValidElementsByCombiningElements(elements: string[]): boolean {
    try {
        return combineIntoValidStringsAndValidateResult(elements);
    } catch (_) {} // Catch any unexpected issues encountered while attempting to validate combined results
    return false;
}

// Main comparison logic for validity check of version strings
function combineIntoValidStringsAndValidateResult(combinedElements: string[]): boolean {
    const sortedCombinedElements = [...combinedElements].sort();
    
    let hasMismatches = false;

    // Loop through combined elements to ensure every element is valid
    for (let i = 1; i < sortedCombinedElements.length; i++) {
        if (parseInt(sortedCombinedElements[i], 10) <= parseInt(sortedCombinedElements[i - 1], 10)) {
            hasMismatches = true;
        }
    }

    return !hasMismatches && sortedCombinedElements[0].length == 6 ? true : false;
}
```

Keep refining this implementation as per specific requirements and consider adding proper error messages to assist in diagnosing problems related to inconsistent data structures within version strings. This revised solution now adheres more strictly to good coding practices aiming at maintaining efficiency and robustness over its entire lifecycle from initial development till release into production.

frontend/src/lang/modules/en.ts
wanghe-fit2cloud, line 1752: The code has no known irregularities or issues, but it can be optimized to improve readability and maintainability. Here's how:

Change `var` statements with ES2015 keywords (`const`, `let`) where appropriate:

```javascript
const message =
{
    upgradeNow: 'Upgrade now',
    source: 'Download source',
    hasNewVersion: 'New version Available',

```

Remove redundant comments from some lines like:

```typescript
versionHigher:
...
      'Please upgrade the main node system version and try again!',

``` 

Replace string concatenation expressions with template literals if they are used frequently.

For example, instead of `upgradeNow:` you'd write `message.upgradeNow = "Upgrade now"`.

These changes will make the code cleaner while keeping its functionality intact.

wanghe-fit2cloud (commented):
/lgtm

PR #7649, opened 2025-01-06
feat(appstore): Handle Website Directory During Second Installation o…

zhengkunwang223:


agent/app/model/app_install.go
```diff
@@ -31,19 +31,6 @@ func (i *AppInstall) GetPath() string {
...
-//	websiteDir, ok := data["WEBSITE_DIR"].(string)
-//	if !ok || websiteDir == "" {
-//		return path.Join(i.GetAppPath(), i.Name, "www")
-//	}
-//	return websiteDir
-//}
-
 func (i *AppInstall) GetComposePath() string {
 	return path.Join(i.GetAppPath(), i.Name, "docker-compose.yml")
 }
```
wanghe-fit2cloud, line 36: The changes made between the original code and mine are:

Original function to get the site directory is missing.

My suggestion is: Create function `GetSiteDir` with the below logic:

```go
var data map[string]interface{}
if err := json.Unmarshal([]byte(i.Env), &data); err != nil {
  return path.Join(i.GetAppPath(), i Name, "www")
}

websiteDir, ok := data["WEBSITEDIOR"].(string)

If not OK or websitedir exists then use 'path.Join' and 'i.GetName()' instead of direct value

```

For optimization you can add docstrings at top and bottom of file if they're not present else remove unnecessary space in Go comments as it looks better.

agent/app/service/file.go
```diff
@@ -483,3 +485,14 @@ func (f *FileService) ReadLogByLine(req request.FileReadByLineReq) (*response.Fi
...
+func (f *FileService) GetPathByType(pathType string) string {
+	if pathType == "websiteDir" {
+		value, _ := settingRepo.GetValueByKey("WEBSITE_DIR")
+		if value == "" {
+			return path.Join(global.CONF.System.BaseDir, "www")
+		}
+		return value
+	}
+	return ""
+}
```
wanghe-fit2cloud, line 498: This code snippet is not formatted properly so it's difficult to review and determine if there are any irregularities, potential issues, or optimization suggestions. Please ensure that all lines in every file have been properly indented before reviewing them. For instance:

```go

type IFileService interface {
    // methods
}

// Implementation of the FileInterface
...
```
Then run this through `gopls` with full scope enabled (for example, `-scope=full`).

agent/app/api/v2/file.go
```diff
@@ -804,3 +804,13 @@ func (b *BaseApi) BatchChangeModeAndOwner(c *gin.Context) {
...
+
+func (b *BaseApi) GetPathByType(c *gin.Context) {
+	pathType, ok := c.Params.Get("type")
+	if !ok {
+		helper.BadRequest(c, errors.New("error pathType id in path"))
+		return
+	}
+	resPath := fileService.GetPathByType(pathType)
+	helper.SuccessWithData(c, resPath)
+}
```
wanghe-fit2cloud, line 816: There are no major code changes between the two provided versions. The only significant difference is the addition of error handling to both functions. If anything else needs an investigation, please let me know; I'm here to help.

wanghe-fit2cloud (commented):
/lgtm

PR #7650, opened 2025-01-06
feat(common): Standardize Event Bus

zhengkunwang223:


frontend/src/layout/components/Sidebar/index.vue
```diff
@@ -78,13 +78,20 @@ import { isString } from '@vueuse/core';
...
 const version = ref();
+
+bus.on('refreshTask', () => {
+    console.log('on bus message');
+    checkTask();
+});
+
 defineProps({
     menuRouter: {
         type: Boolean,
```
wanghe-fit2cloud, line 97: The key points worth noting about this code snippet are:

1. It appears to be a Vue component that likely represents some form of web application or mobile app functionality.

2. It uses several API calls (listed under `@use` imports) which suggests it's related to server-side operations like getting setting information and logging data.

3. The code includes references to modules (`MenuStore`, `GlobalStore`) for managing UI elements, presumably providing the context in which these methods operate.

4. A method for checking task status on a refresh operation (notably tagged with `[on] bus message`) indicates it may handle notifications when new tasks pop up on the screen.

To address any potential issues or optimisation suggestions:
  
  
  
  * Ensure all dependencies are listed correctly at the top so they can be included during compilation if needed. This avoids breaking older versions of libraries.
  
  * Check for best practices: Is there more readable structure defined than just using `ref`? Consider creating functions for handling dropdown items separately could enhance readability. Also add checks here for whether we need async actions instead of plain getters/setters since an actual call might not happen immediately upon accessing them but should be fast enough most times.

  * If possible, try adding comments throughout where relevant to explain what each part does within the larger scope rather than relying only on function names; this makes code easier to understand even outside its intended usage area.

  * For performance considerations consider using Promise.all() for multiple asynchronous requests. However note that Promise.resolve().then(...).catch(...) will have side effects due to chaining, hence you would want to wrap this in another promise chain or use Promises/A+ library if available. 

Note: Please replace placeholder "API calls" as per actual components or services used.

frontend/src/views/app-store/detail/install/index.vue
```diff
@@ -136,16 +136,15 @@ import TaskLog from '@/components/task-log/index.vue';
...
 }
-
 const installData = ref<InstallRrops>({
     app: {},
 });
+
+const router = useRouter();
 const open = ref(false);
 const rules = ref<FormRules>({
     name: [Rules.appName],
```
wanghe-fit2cloud, line 150: The code appears to be for a Vue component, specifically designed to log tasks within an application context. There is no obvious difference in the logic between these two versions, but there might have been a typo regarding the `app` parameter in one of them.

In general:

* The "router" variable seems unused; however, if needed it could improve debugging if the actual value isn't set anywhere else (it likely refers to a web API).

* The use case (`App.AppParams`) does not appear to match the provided typescript definition. It's recommended to either define that specific prop type explicitly inside this file or replace it with something meaningful here as it doesn't seem relevant given the context.

Since versioning dates do not apply directly in this situation, you may prefer to test both implementations side-by-side at different points in time rather than comparing their exact state on an earlier date for comparison against potential changes.

Given your request being purely about identifying any irregularities and suggesting possible improvements over the base implementation, I can't make strong optimizations without seeing more details of the functionality of each instance of such a class method and its use cases. For example, consider whether there's a cleaner way to handle props other than ref, or ways to simplify methods like open() and form validation etc.

Here’s how it usually should look based on your comments:

```js
// Invariant version

import { useState } from 'react';

const TaskLogs = () => {
  const [tasks, setTasks] = useState([]);

  // Functionality omitted below.
}

```

...and possibly another alternative implementation.

---

As noted above, since we're focusing only on what you asked me not on the actual implementation details - which would typically involve deep testing or reviewing source code in isolation, my suggestion remains somewhat broad. However, generally speaking when checking for differences, remember to focus especially on properties, functional dependencies, naming conventions, style consistency (if applicable), and readability. Make sure everything works correctly and efficiently before deciding on the best design choice.

frontend/src/components/task-log/index.vue
```diff
@@ -50,6 +51,7 @@ const openWithTaskID = (id: string, tail: boolean) => {
         config.tail = tail;
     }
     open.value = true;
+    bus.emit('refreshTask', true);
 };
 
 const openWithResourceID = (taskType: string, taskOperate: string, resourceID: number) => {
```
wanghe-fit2cloud, line 57: There do not appear to be any obvious differences or potential issues with this provided code snippet. It appears to have been written using Vue.js framework and contains an event handler that will emit a "refreshTask" message when called with specific parameters.

If you need assistance with another question of your own or further analysis on existing functions, feel free to ask!

wanghe-fit2cloud (commented):
/lgtm

Issue #7658, opened 2025-01-07
[Bug] The security entry route loaded when the project starts can still be used after the security entry is modified.


Anyexyz:
### Contact Information

Anye

### 1Panel Version

Free: v1.10.22-lts

### Problem Description

The security entrance configured before the project starts can still be accessed after modification, and 1Panel needs to be restarted to reset the disabled.

### Steps to Reproduce

When installing 1Panel, configure the security entry: `demo`, then log in to the panel and change the security entry to `demodemo`. The `demo` route can still be accessed before restarting the panel.

zhengkunwang223:
Thank you for your feedback. We will address this issue in a future version

wanghe-fit2cloud:
v1.10.23-lts 版本已发布。

PR #7659, opened 2025-01-07
feat(system): Fix Issue Allowing Access via Old Secure Entry After Mo…

zhengkunwang223:
Refs https://github.com/1Panel-dev/1Panel/issues/7658

### DIFF
--- a/backend/init/router/router.go
+++ b/backend/init/router/router.go
@@ -126,8 +126,8 @@ func setWebStatic(rootRouter *gin.RouterGroup) {
 	entrance := authService.GetSecurityEntrance()
 	if entrance != "" {
 		rootRouter.GET("/"+entrance, func(c *gin.Context) {
-			entrance = authService.GetSecurityEntrance()
-			if entrance == "" {
+			currentEntrance := authService.GetSecurityEntrance()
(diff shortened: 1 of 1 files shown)


backend/init/router/router.go
```diff
@@ -126,8 +126,8 @@ func setWebStatic(rootRouter *gin.RouterGroup) {
 	entrance := authService.GetSecurityEntrance()
 	if entrance != "" {
 		rootRouter.GET("/"+entrance, func(c *gin.Context) {
-			entrance = authService.GetSecurityEntrance()
-			if entrance == "" {
+			currentEntrance := authService.GetSecurityEntrance()
+			if currentEntrance == "" || currentEntrance != entrance {
 				handleNoRoute(c)
 				return
 			}
```
wanghe-fit2cloud, line 133: There is no issue with the code itself; it appears to be checking if the `entrance` parameter has changed since authentication. If so, it triggers a "no route" handling function using `handleNoRoute`. The current logic correctly ensures that changes in `entrance` trigger an error response when they occur after the initial check during authentication.

wanghe-fit2cloud (commented):
/lgtm

PR #7661, opened 2025-01-07
fix: Fix the issue of frontend packaging errors

ssongliu:


frontend/vite.config.ts
```diff
@@ -83,7 +83,7 @@ export default defineConfig(({ mode }: ConfigEnv): UserConfig => {
             }),
         ],
         esbuild: {
-            pure: viteEnv.VITE_DROP_CONSOLE ? ['console.log', 'debugger'] : [],
+            pure: viteEnv.VITE_DROP_CONSOLE ? ['console.log'] : [],
         },
         build: {
             outDir: '../cmd/server/web',
```
wanghe-fit2cloud, line 89: The changes are minor:

1. The `esbuild.pure` option is adjusted to exclude both `'console.log'` and `'debugger'`, but only `'console.log'` will be excluded now.

Optimization suggestions:
- If you want to completely remove everything printed in the console (not just logs), consider using `process.env.NODE_ENV === 'production'` instead of checking for environment variables directly.
- Ensure that other parts of your codebase don't print things unexpectedly if they expect the original behavior.

wanghe-fit2cloud (commented):
/lgtm

PR #7754, opened 2025-01-22
fix: Fix OpenResty Installation Failure

zhengkunwang223:


agent/app/service/app.go
```diff
@@ -360,10 +360,17 @@ func (a AppService) Install(req request.AppInstallCreate) (appInstall *model.App
...
+				_ = fileOp.CreateDir(siteDir, constant.DirPerm)
+			}
+			err = settingRepo.UpdateOrCreate("WEBSITE_DIR", siteDir)
+			if err != nil {
+				return
 			}
-			_ = settingRepo.UpdateOrCreate("WEBSITE_DIR", siteDir)
 		}
 	}
 	for key := range req.Params {
```
wanghe-fit2cloud, line 376: The given code doesn't contain any significant issues or suggestions for changes to be made from my perspective. There don't appear to be any irregularities detected. The provided function "Install" is attempting to set the value of "WEBSITE_DIR" variable according to a mapping specified in `req.Params`. It first verifies if `oldWebStePath` (the previous value saved on filesystem) matches the current one. If it does not match, then it calls a method named `_`. This seems like an implementation detail and doesn't seem to have a major impact here.

However, based on the context and purpose of this code being for installation purposes specifically into an existing project, there's probably more room for improvement regarding the handling of potential errors when saving/deleting files/filesystem operations. For instance:

- Check if renaming old webste path works properly after file op stat/filename check; also add some error checking mechanism around rename/diff between filenames/states.
- Consider using package constant/strings instead of hardcoded values like dir perm constants for better practices (especially since these might vary across different OSes). 

This was just a brief summary - please note that you should double-check with the original team about specific use cases or additional requirements they may have.

frontend/src/components/task-log/index.vue
```diff
@@ -24,7 +24,7 @@ defineProps({
     },
     width: {
         type: String,
-        default: '30%',
+        default: '50%',
     },
     tail: {
         type: Boolean,
```
wanghe-fit2cloud, line 30: There does not appear to be any specific issue with the code provided; however, it could be useful if you have further modifications or changes in mind:

1. **Type Declaration:** Ensure all types (String and Boolean) match properly according to your requirements.

2. **Default Value Specification:** The second `width` property has a default value of `${'30%'}` which is fine but can be more descriptive like `{default: '3/5'}`. This makes it clear that it's set at 60%.

If there were errors mentioned about the code itself, please share them so I can offer appropriate advice accordingly!

wanghe-fit2cloud (commented):
/lgtm

PR #7769, opened 2025-01-24
fix: complete the missing international content

ssongliu:


core/app/api/v2/helper/helper.go
```diff
@@ -14,6 +14,10 @@ func ErrorWithDetail(ctx *gin.Context, code int, msgKey string, err error) {
 		Code:    code,
 		Message: "",
 	}
+	if msgKey == "ErrCaptchaCode" || msgKey == "ErrAuth" {
+		res.Code = 406
+		res.Message = msgKey
+	}
 	res.Message = i18n.GetMsgWithMap(msgKey, map[string]interface{}{"detail": err})
 	ctx.JSON(http.StatusOK, res)
 	ctx.Abort()
```
wanghe-fit2cloud, line 23: I cannot determine the specific difference or issue without reviewing all of your submitted code. However, I can suggest that you double-check if there's an inconsistency between message translation keys (`msgKey`) and messages being constructed (`res.Message`). Additionally, ensure consistency across different parts of your program (e.g., functions within `ErrorWithDetails`).

core/utils/req_helper/requset.go
```diff
@@ -99,13 +99,13 @@ func handleGetWithTransport(url string, transport *http.Transport) (*http.Respon
...
 			}
 			return nil, buserr.WithMap("ErrHttpReqFailed", map[string]interface{}{"err": err.Error()}, err)
 		}
 	}
 	if resp.StatusCode == 404 {
-		return nil, buserr.New("ErrHttpReqNotFound")
+		return nil, buserr.WithErr("ErrHttpReqNotFound", errors.New("no such resource"))
 	}
 
 	return resp, nil
```
wanghe-fit2cloud, line 111: The code has no known irregularities. No issues or optimization suggestions were indicated during analysis of the given code snippet related to HTTP GET requests handling with timeouts using Go programming language (which is based on OpenAI's knowledge). This code snippet appears correct according to its current use case and does not require further modifications.

However, if you are looking at other aspects like concurrency controls or performance optimizations without changing core functionality:

For concurrency control purposes within goroutines, checking for context cancellation can help avoid unnecessary network calls in error recovery scenarios.

Also look into more efficient timeouts than the provided example, such as custom timeouts defined within `transport` struct fields, which allow managing request times faster by adjusting their settings dynamically rather than having them hard-coded everywhere (`defaultMaxIdleConnsPerHost`).

Regarding performance improvements, consider adding additional checks after timeouts to ensure they're due to timeout conditions; these could be non-existent, unexpected behavior, or server configuration changes that have no effect when expected.

Note that while this code doesn't exhibit any obvious problems, it might benefit from careful review considering its broader function scope and how it interacts with the rest of system infrastructure including error contexts or response processing logic within handlers, services etc.

wanghe-fit2cloud (commented):
/lgtm

Issue #7657, opened 2025-01-07
[Bug] The `/api/v1/settings/update` interface does not impose a string length limit.


Anyexyz:
### Contact Information

Anye

### 1Panel Version

Free: v1.10.22-lts

### Problem Description

The `/api/v1/settings/update` interface does not impose a string length limit.
![60c764f04ef657480e33b1ad29d42b43](https://github.com/user-attachments/assets/89bb450a-d6b3-4749-bf03-5a1d2efa9148)
![94e1b4c934b65019ca6a937ef3f7d3b5](https://github.com/user-attachments/assets/99554f8c-7474-4b15-a2fa-15149ca5eace)

### Steps to Reproduce

It is not possible to directly set the security entry to `demo` in the panel, but it is possible to directly request the interface.

wanghe-fit2cloud:
The current version does have some API interfaces lacking validation. We will gradually enhance this in future versions.

wanghe-fit2cloud:
v1.10.24-lts 版本已发布。

PR #7810, opened 2025-02-06
fix: improve the backend verification for security entry settings

ssongliu:
Refs #7657

### DIFF
--- a/backend/app/api/v1/setting.go
+++ b/backend/app/api/v1/setting.go
@@ -3,8 +3,10 @@ package v1
 import (
 	"encoding/base64"
 	"errors"
+	"fmt"
 	"os"
 	"path"
+	"regexp"
 
 	"github.com/1Panel-dev/1Panel/backend/app/api/v1/helper"
 	"github.com/1Panel-dev/1Panel/backend/app/dto"
@@ -52,6 +54,12 @@ func (b *BaseApi) UpdateSetting(c *gin.Context) {
 	var req dto.SettingUpdate
 	if err := helper.CheckBindAndValidate(&req, c); err != nil {
 		return
+	}
+	if req.Key == "SecurityEntrance" {
+		if checkEntrancePattern(req.Value) {
+			helper.ErrorWithDetail(c, constant.CodeErrBadRequest, constant.ErrTypeInvalidParams, fmt.Errorf("regexp match string with %s failed", req.Value))
+			return
+		}
 	}
 
 	if err := settingService.Update(req.Key, req.Value); err != nil {
@@ -390,4 +398,12 @@ func (b *BaseApi) UpdateApiConfig(c *gin.Context) {
 		return
 	}
 	helper.SuccessWithData(c, nil)
+}
+
+func checkEntrancePattern(val string) bool {
+	if len(val) == 0 {
+		return true
+	}
+	result, _ := regexp.MatchString("^[a-zA-Z0-9]{5,116}$", val)
+	return result
 }


backend/app/api/v1/setting.go
```diff
@@ -391,3 +399,11 @@ func (b *BaseApi) UpdateApiConfig(c *gin.Context) {
...
 	helper.SuccessWithData(c, nil)
 }
+
+func checkEntrancePattern(val string) bool {
+	if len(val) == 0 {
+		return true
+	}
+	result, _ := regexp.MatchString("^[a-zA-Z0-9]{5,116}$", val)
+	return result
+}
```
wanghe-fit2cloud, line 409: 1. **Unused Variable**: The variable `err` is declared but not used in the function `checkEntrancePattern`. This could be removed to clean up the code.
2. **Variable Naming**: Use lowercase variables (`pattern`) instead of camelCase for better readability and convention consistency.
3. **Regex Usage**: Ensure that the regex pattern is accurate. If it should only allow alphanumeric characters, consider using `[a-zA-Z0-9]+`.
4. **Null Check**: Add a check at the beginning of `UpdateSetting` to avoid unnecessary processing if `req.Key` is missing.
5. **Error Handling**: Consider handling potential errors outside of the return clause.

Here's the improved code:

```go
package v1

import (
	"encoding/base64"
	"errors"
	"fmt"
	"os"
	"path"

	"github.com/1Panel-dev/1Panel/backend/app/api/v1/helper"
	"github.com/1Panel-dev/1Panel/backend/app/dto"
)

func (b *BaseApi) UpdateSetting(c *gin.Context) {
	req := dto.SettingUpdateReq{}
	var err error
	if err = helper.CheckBindAndValidate(&req, c); err != nil {
		return
	}

	defer func() {
		if r := recover(); r != nil {
			c.AbortWithStatusJSON(500, gin.H{"error": fmt.Sprint(r)})
		}
	}()

	if req.Key == "SecurityEntrance" && req.Value != "" {
		pattern := "^([A-Za-z0-9]{5,116})?$" // Allow only numbers and letters
		match, err := regexp.MatchString(pattern, req.Value)
		if err != nil || !match {
			helper.ErrorWithDetail(c, constant.CodeErrBadRequest, constant.ErrTypeInvalidParams, fmt.Sprintf("regexp match string with %s failed", req.Value))
			return
		}
	}

	err = settingService.Update(req.Key, req.Value)
	if err != nil {
		helper.ErrorWithDetail(c, constant.CodeErrInternalServer, constant.ErrTypeInternalServer, err)
		return
	}
	
	helper.SuccessWithData(c, nil)
}

func checkEntrancePattern(value string) bool {
	return len(value) > 0 &&
		regexp.MustCompile(`^([A-Za-z0-9]{5,116})$`).MatchString(value)
}
```

This refactored code addresses some issues such as unused variables and improves readability while maintaining functionality.

wanghe-fit2cloud, line 409: The provided code is almost correct but includes minor issues that can be addressed:

1. **Missing Import**: The package `fmt` is used in the `checkEntrancePattern` function, but it might not have been imported at the top of the file. Ensure you have this import statement: `import "fmt"`.

2. **Regex String**: In the regex pattern `"^[a-zA-Z0-9]{5,116}$"`, there should be quotation marks around all characters within the brackets to correctly define the character set.

3. **Empty Value Check**: The `checkEntrancePattern` function now returns `true` if the value is empty, which might unintentionally allow empty strings. You might want to change this to return `false` instead since an empty security entrance seems invalid.

Here's the corrected version of the code with these adjustments:

```go
package v1

import (
    "encoding/base64"
    "errors"
    "fmt"
    "os"
    "path"

    "github.com/1Panel-dev/1Panel/backend/app/api/v1/helper"
    "github.com/1Panel-dev/1Panel/backend/app/dto"
)

// ... (rest of the code remains unchanged except for...

func (b *BaseApi) UpdateSetting(c *gin.Context) {
    // ... (existing logic...)
}

// ... (rest of the code remains unchanged except for...

func checkEntrancePattern(val string) bool {
    if len(val) == 0 {
        return false
    }
    result, _ := regexp.MatchString(`^[a-zA-Z0-9]{5,116}$`, val)
    return result
}
```

These changes ensure the code is more robust and follows best practices.

wanghe-fit2cloud (commented):
/lgtm

PR #7834, opened 2025-02-10
fix: Fixed the issue where application labels were not displayed afte…

zhengkunwang223:


backend/app/service/app.go
```diff
@@ -129,8 +129,14 @@ func (a AppService) GetAppTags(ctx *gin.Context) ([]response.TagDTO, error) {
...
 			tagDTO.Name = name
+		} else {
+			if lang == "zh" || lang == "zh-Hant" {
+				tagDTO.Name = tag.Name
+			} else {
+				tagDTO.Name = tag.Key
+			}
 		}
 		res = append(res, tagDTO)
 	}
```
wanghe-fit2cloud, line 142: The provided code is correct with no significant issues. However, there is one minor issue with how it handles missing translations for languages other than "zh" or "zh-Hant". It would be more efficient if you handle this case earlier rather than inside the loop.

Here's an optimized version of the code:

```go
func (a AppService) GetAppTags(ctx *gin.Context) ([]response.TagDTO, error) {
	var res []response.TagDTO

	for _, tag := range tags {
		defaultName := ""
		translationLangMap := map[string][]string{
			"":         []string{tag.Key},
			"zh":        []string{"zh", "zh-CN", "zh-SG"},
			"zh-Hant":   []string{"cht", "zh-TW"},
		}

		for targetLang, fallbackLangs := range translationLangMap {
			if lang == targetLang || contains(fallbackLangs, lang) {
				err := json.Unmarshal([]byte(tag.Translations), &translations)
				if err != nil {
					return nil, errors.Wrap(err, "failed to unmarshal tag translations")
				}
				if name, ok := translations[lang]; ok && name != "" {
					tagDTO.Name = name
					res = append(res, tagDTO)
					break
				}
			}
		}
		
		if len(res) == 0 {
			defaultName = tag.Key
			if lang == "" {
				continue // Skip default name addition for root category
			}
		}

		// Add default name at the end if no translation was found for the requested language
		if lang != "" {
			tagDTO := response.TagDTO{Name: defaultName}
			res = append(res, tagDTO)
		}
	}

	return res, nil
}

func contains(arr []string, item string) bool {
	for _, v := range arr {
		if v == item {
			return true
		}
	}
	false
}

```

This version first initializes `defaultName` outside the loop and checks all supported languages (`""`, `"zh"`, and `"zh-Hant"`). If a suitable translation is not found, it sets `defaultName` appropriately before appending it. This reduces redundant JSON decoding and improves clarity.

wanghe-fit2cloud (commented):
/lgtm

PR #7872, opened 2025-02-14
fix: Fix the problem of duplicate content display in the model list

ssongliu:


backend/app/service/website_utils.go
```diff
@@ -1150,17 +1150,17 @@ func GetAllowIps(website model.Website) []string {
...
+	// config := nginxFull.SiteConfig.Config
+	// server := config.FindServers()[0]
+	// dirs := server.GetDirectives()
+	// for _, dir := range dirs {
+	// 	if dir.GetName() == "location" && dir.GetParameters()[0] == "/" {
+	// 		// server.UpdateRootProxy()
+	// 	}
+	// }
 	return nil
 }
```
wanghe-fit2cloud, line 1166: The provided code appears to have several issues and inefficiencies:

1. **Irregular Logic**: The code contains commented-out blocks that perform operations related to setting up an AI proxy in Nginx, but none of these operations are actually executed.

2. **Return Value**: The function returns `nil` without performing any configuration tasks, which is not correct given the name suggests it should handle proxy configurations.

3. **Code Duplication**: There seems to be duplicated logic around obtaining Nginx configuration and iterating over directives, although some duplication has been removed.

### Potential Improvements:
1. **Remove Commented-Out Code**: If the intention is for this code to do something, it should not remain commented out.

2. **Implement Proper Configuration Setup**: Replace the comment on line 148 with actual setup logic to apply the AI proxy configuration.

3. **Optimize Looping Over Directives**: Ensure there isn’t unnecessary looping or redundant checks.

Here's an example refactor incorporating optimizations:

```go
func ConfigAIProxy(website model.Website) error {
	config := website.SiteConfig.Config

	for _, server := range config.Servers() {
		for _, dir := range server.Directives() {
			if dir.Name == "location" && dir.Parameter[0] == "/" {
				dir.SetParameter("proxy_pass", "/api-proxy-url")
				break
			}
		}
	}

	return nil
}
```

### Key Changes Made:
- Removed commenting.
- Added a loop through server and directive lists.
- Set the `proxy_pass` parameter directly within the iteration process, assuming `/api-proxy-url` is the intended target URL.
- This approach assumes that the structure of `config.Servers()` and `server.Directives()` matches typical Nginx configurations.

Replace `/api-proxy-url` with the actual URL you want to configure as the reverse proxy.

frontend/src/views/ai-tools/model/index.vue
```diff
@@ -58,9 +58,15 @@
...
+                                type="primary"
+                                class="cursor-pointer"
+                                @click="onLoad(row.name)"
+                            >
                                 {{ row.name }}
                             </el-text>
+                            <span v-else>{{ row.name }}</span>
                         </template>
                     </el-table-column>
                     <el-table-column :label="$t('file.size')" prop="size" />
```
wanghe-fit2cloud, line 72: The code snippet you provided has a single change to address an edge case where `row.size` might be `'-'`. Here are the changes and their implications:

### Change Details

1. **Conditional Spanning with Vue Directives:**
   ```vue
   <template #default="{ row }">
       <el-text
           v-if="row.size !== '-'"
           type="primary"
           class="cursor-pointer"
           @click="onLoad(row.name)"
       >
           {{ row.name }}
       </el-text>
+      <span v-else>{{ row.name }}</span>
   </template>
   ```
   
2. **Removing `<el-text>` for Non-Sized Rows in Place of Default Template:**
   ```vue
                       <el-table-column :label="$t('commons.table.name')" prop="name" min-width="90">
                           <!-- ... -->
                           <template #default="{ row }">
                               <el-text
                                   v-if="row.size !== '-'"
                                   type="primary"
                                   class="cursor-pointer"
                                   @click="onLoad(row.name)"
                               >
                                   <!-- This line now directly shows the name without additional wrapping -->
                                   {{ row.name }}
                               </el-text>
+                              <span v-else>{{ row.name }}</span>
                           </template>
                           <!-- ... -->
                       </el-table-column>
               ```

#### Implications

**Positive Changes:**
1. **Visual Clarity:** By using a conditional span (`<span>`) when `row.size === "-"`, we ensure that table cells containing no size information are visually distinct from those with valid sizes. This improves readability.
  * Example Scenario:
    ```vue
        // Old (incorrect) behavior: No visual distinction between rows with/without size "-"
        | Name          | Size    |
        |---------------|---------|
        | FileA         | Large   |
        | FileB          | -       |
        
        // New behavior with proper separation:
        | Name          | Size    |
        |---------------|---------|
        | FileA         | Large   |
        | FileB          |         |
    ```

**Negative Changes/Nominations:**

None. The update aligns well within existing template structure and does not introduce new bugs.

### Suggested Improvements/Optimizations

- Ensure consistency throughout related components; sometimes, similar logic may exist elsewhere making these updates necessary or redundant. 

In summary, this change enhances table readability while maintaining the functionality needed to handle empty size values appropriately.

backend/app/service/ai_tool.go
```diff
@@ -56,7 +56,7 @@ func (u *AIToolService) Search(req dto.SearchWithPage) (int64, []dto.OllamaModel
 		if parts[0] == "NAME" {
 			continue
 		}
-		modelMaps[parts[0]] = struct{}{}
+		modelMaps[strings.ReplaceAll(parts[0], ":", "-")] = struct{}{}
 		list = append(list, dto.OllamaModelInfo{Name: parts[0], Size: parts[2] + " " + parts[3], Modified: strings.Join(parts[4:], " ")})
 	}
 	entries, _ := os.ReadDir(path.Join(global.CONF.System.DataDir, "log", "AITools"))
```
wanghe-fit2cloud, line 62: The changes made seem to be aimed at normalizing and sanitizing the keys used in `modelMaps`. Specifically, they replace all colons (`:`) with dashes (`-`) in model names before adding them to the `modelMaps` map.

This change doesn't appear to introduce any significant issues or optimize the code. However, it does have some implications:

1. **Consistency**: This approach ensures that model names will always use hyphens instead of colons when stored or accessed later on.
2. **Safety**: It helps sanitize input from other parts of the system, potentially reducing the risk of malformed data causing errors if models are named with colons.

Here's a slightly more concise version of the relevant parts:

```go
	for _, line := range lines {
		parts := strings.Fields(line)
		if len(parts) < 6 || parts[0] == "NAME" {
			continue
		}
		modelMaps[strings.ReplaceAll(parts[0], ":", "-")] = struct{}{}
		list = append(list, dto.OllamaModelInfo{
			Name:    parts[0],
			Size:     parts[2] + " " + parts[3],
			Modified: strings.Join(parts[4:], " "),
		})
	}
```

Overall, this modification maintains functionality while improving consistency and potential for security benefits.

wanghe-fit2cloud (commented):
/lgtm

PR #7875, opened 2025-02-15
feat: Added streaming transmission configuration

zhengkunwang223:


backend/app/service/website_utils.go
```diff
@@ -1150,17 +1150,17 @@ func GetAllowIps(website model.Website) []string {
...
+	config := nginxFull.SiteConfig.Config
+	server := config.FindServers()[0]
+	dirs := server.GetDirectives()
+	for _, dir := range dirs {
+		if dir.GetName() == "location" && dir.GetParameters()[0] == "/" {
+			server.UpdateRootProxyForAi([]string{fmt.Sprintf("http://%s", website.Proxy)})
+		}
+	}
 	return nil
 }
```
wanghe-fit2cloud, line 1166: The code has been modified to correctly call `UpdateRootProxyForAi` with the proxy URL formatted as an HTTP address (e.g., "http://example.com"). This adjustment is necessary because it updates the root proxied location directive for AI purposes. The rest of the modifications appear to correct a previous logic error where `server.UpdateRootProxy()` was incorrectly called instead of `server.UpdateRootProxyForAi()`.

backend/utils/nginx/components/server.go
```diff
@@ -291,6 +291,18 @@ func (s *Server) UpdateRootProxyForAi(proxy []string) {
...
+		},
+		&Directive{
+			Name: "chunked_transfer_encoding",
+			Parameters: []string{
+				"off",
+			},
+		},
 	}
 	block.Directives = append(block.Directives, &Directive{
 		Name:       "proxy_pass",
```
wanghe-fit2cloud, line 308: The updated code provides some security measures to improve the performance of the application. Adding `Connection` and `Content-Length` headers prevents clients from reusing old HTTP connections without explicitly closing them, which can lead to information disclosure vulnerabilities. Setting `Chunked_Transfer_Encoding` to off ensures that responses are sent using fixed-size buffers, reducing buffering overhead and potentially improving load balancing efficiency.

It's worth noting that these changes might increase the complexity of managing header settings for proxies, especially if they need to be tailored based on specific use cases or requirements beyond general-purpose proxying. Additionally, these updates should ensure that all necessary directives relevant for API endpoints have been included, although no additional details were provided about this part of the code snippet.

backend/app/service/ai_tool.go
```diff
@@ -233,6 +233,9 @@ func (u *AIToolService) BindDomain(req dto.OllamaBindDomain) error {
 			return err
 		}
 	}
+	if err = ConfigAIProxy(website); err != nil {
+		return err
+	}
 	return nil
 }
```
wanghe-fit2cloud, line 241: No discrepancies were found in the `BindDomain` function regarding logical content, but it's recommended to add comments explaining each step of the function for better readability and maintainability:

```go
func (u *AIToolService) BindDomain(req dto.OllamaBindDomain) error {
    if err := u.BindToOllamaConfig(&req.Config); err != nil {
        return err
    }
+    // Call ConfigAIProxy after binding configuration changes
+    // This ensures that domain binding updates proxy configurations correctly
+    if err = ConfigAIProxy(req.DemoSiteWebsiteURL); err != nil {
+        return err
+    }
    return nil
}
```

Optimization suggestions:
1. Ensure that `BindToOllamaConfig` returns an appropriate error message when it fails. While this is not critical, it can improve debugging and user experience.
2. Consider whether calling `ConfigAIProxy` every time a new site is bound is necessary, especially since you only need to call it after certain specific configuration updates. If performance becomes an issue, consider caching recent calls or optimizing data structures accordingly.

By adding comments and implementing simple optimizations, the code becomes clearer and more robust.

wanghe-fit2cloud (commented):
/lgtm

PR #7899, opened 2025-02-18
fix: Optimized ollama model status display

ssongliu:


frontend/src/views/ai/model/index.vue
```diff
@@ -407,7 +417,7 @@ const buttons = [
             onDelete(row);
         },
         disabled: (row: any) => {
-            return row.status !== 'Success';
+            return row.status !== 'Success' && row.status !== 'Deleted';
         },
     },
 ];
```
wanghe-fit2cloud, line 423: 1. The template changes include adding a new alert section below `AppStatus`. This could be useful for notifying users about system restarts.

2. In the `<template>` tag under "#toolbar" for "Pending", you added an extra closing curly brace `}`). Make sure there are no unmatched braces in your code.

3. In both `onDelete` and `onLoadLog`, when checking log from, ensure that it's not null or undefined before using string methods like `.includes()` to avoid errors in development environment where logs can take time to load or may not be completely ready yet.

4. Added additional conditions on disabling button action based on different model status such as 'Cancelled'.

5. For better maintenance and clarity, separate business logic functions into dedicated named functions rather than inline assignments within event handlers. 

6. Consider replacing hardcoded strings with translations keys `$t(...)` wherever applicable which makes your application more translatable without modifying UI files directly.

These are suggestions to improve efficiency, security & ease of future maintenance. Check them against your project guidelines specifically regarding best practices and standards being used.

backend/app/service/ai.go
```diff
@@ -152,9 +153,11 @@ func (u *AIToolService) Delete(req dto.ForceDelete) error {
...
-		if err != nil && !req.ForceDelete {
-			return fmt.Errorf("handle ollama rm %s failed, stdout: %s, err: %v", item.Name, stdout, err)
+		if item.Status != constant.StatusDeleted {
+			stdout, err := cmd.Execf("docker exec %s ollama rm %s", containerName, item.Name)
+			if err != nil && !req.ForceDelete {
+				return fmt.Errorf("handle ollama rm %s failed, stdout: %s, err: %v", item.Name, stdout, err)
+			}
 		}
 		_ = aiRepo.Delete(commonRepo.WithByID(item.ID))
 		logItem := path.Join(global.CONF.System.DataDir, "log", "AITools", item.Name)
```
wanghe-fit2cloud, line 163: - `stdOut` is not reassigned after the command execution when dealing with ollama items that do not need force deletion.
- The logging directory creation line should be inside the loop where AITool entities are processed to ensure each model has its log path created separately if needed.

backend/init/hook/hook.go
```diff
@@ -150,6 +151,11 @@ func handleSnapStatus() {
...
 }
 
+func handleOllamaModelStatus() {
+	message := "the task was interrupted due to the restart of the 1panel service"
+	_ = global.DB.Model(&model.OllamaModel{}).Where("status = ?", constant.StatusWaiting).Updates(map[string]interface{}{"status": constant.StatusCanceled, "message": message}).Error
+}
+
 func handleCronjobStatus() {
 	var jobRecords []model.JobRecords
 	_ = global.DB.Where("status = ?", constant.StatusWaiting).Find(&jobRecords).Error
```
wanghe-fit2cloud, line 161: The code appears to be well-written and handles various initialization tasks efficiently. There are no significant irregularities, but I can provide a couple of minor improvements:

1. **Variable Initialization**:
   In `handleOllamaModelStatus`, you assign an empty string to `message` instead of creating a new instance of `model.OllamaModel`. However, since you're updating all waiting models with the same error message, it's likely sufficient.

2. **Database Query Efficiency**:
   You might want to ensure that your database queries are selective. If there is only one waiting model and you use `.First()` or `.Last()`, it should suffice without selecting all waiting models and then iterating over them manually. For multiple updates, using bulk operations like `.Updates()` can potentially improve performance.

3. **Error Handling**:
   While `_ = err.Error` is used to handle errors when they occur in Go, it doesn't have any effect on the flow of execution. It simply suppresses log messages from being printed. In most production environments, it's better to log these errors explicitly to aid debugging.

Here are some refined recommendations:

```go
func init() {
    // ... existing code ...

    handleCronjobStatus()

    handleOllamaModelStatus()
    handleSnapStatus()
    loadLocalDir()
    initDir()
}

func handleOllamaModelStatus() {
    var model model.OllamaModel
    result := global.DB.Where("status = ?", constant.StatusWaiting).FirstOrLast(&model)
    
    if !result.RecordNotFound() && result.RowsAffected > 0 {
        model.Status = constant.StatusCanceled
        model.Message = "the task was interrupted due to the restart of the 1panel service"
        _ = global.DB.Save(&model).Error
    }
}

// No change needed in handleCronjobStatus and related functions
```

If you don't intend to make additional changes or need further assistance, feel free to ask!

wanghe-fit2cloud (commented):
/lgtm

PR #7916, opened 2025-02-19
fix: Fixed the issue with incorrect routing for quick jump during app…

zhengkunwang223:


frontend/src/routers/index.ts
```diff
@@ -39,9 +39,12 @@ router.beforeEach((to, from, next) => {
...
+    if (to.path === '/apps/all' && to.query.install != undefined) {
+        return next();
+    }
+
     const activeMenuKey = 'cachedRoute' + (to.meta.activeMenu || '');
     const cachedRoute = localStorage.getItem(activeMenuKey);
-
     if (
         to.meta.activeMenu &&
         to.meta.activeMenu != from.meta.activeMenu &&
```
wanghe-fit2cloud, line 50: The most notable difference between the two codes is the use of `localStorage.removeItem()` followed by `return next();` at line 47. This may cause unnecessary memory usage and should be replaced with a cleaner approach like calling an async function which will not affect cache behavior. Additionally, please ensure you follow modern naming conventions and avoid global variables or functions where reusable code has been implemented.

For example:

```javascript
try {
    // Remove cached route from local storage if it exists
} catch (error) {}
return await router.push(to);  // Always push to location rather than navigating
```

This version uses try-catch block to prevent errors during removal. The return await statement ensures that no side effects take place in case an error arises while trying to remove the item form cache.

agent/app/service/file.go
```diff
@@ -466,7 +466,7 @@ func (f *FileService) ReadLogByLine(req request.FileReadByLineReq) (*response.Fi
 		if req.TaskID != "" {
 			opts = append(opts, taskRepo.WithByID(req.TaskID))
 		} else {
-			opts = append(opts, repo.WithByType(req.TaskType), taskRepo.WithOperate(req.TaskOperate), taskRepo.WithResourceID(req.ID))
+			opts = append(opts, repo.WithByType(req.TaskType), taskRepo.WithOperate(req.TaskOperate), taskRepo.WithResourceID(req.ResourceID))
 		}
 		taskModel, err := taskRepo.GetFirst(opts...)
 		if err != nil {
```
wanghe-fit2cloud, line 472: The two code files have different names, but the overall structure is almost identical as they serve the same purpose of reading log data from a File service based on a given file read line request. However, there's no direct difference identified that suggests an issue or regularity problem.

To optimize the readability and performance:

1. Make all variables in single lines.
2. Use string slicing to join strings instead of concatenation.
3. Rename functions or variable names to be more descriptive.

For checking the compliance with standards over time, this should not impact anything since they're written well and do what they should: fetching the log data from the provided repository.

wanghe-fit2cloud (commented):
/lgtm

PR #7917, opened 2025-02-19
feat: Optimized the display when no certificate is present.

zhengkunwang223:


frontend/src/views/ai/model/domain/index.vue
```diff
@@ -236,7 +232,7 @@ const search = async (appInstallID: number) => {
             operate.value = 'update';
             req.value.domain = res.data.domain;
             req.value.websiteID = res.data.websiteID;
-            if (res.data.allowIPs.length > 0) {
+            if (res.data.allowIPs && res.data.allowIPs.length > 0) {
                 req.value.ipList = res.data.allowIPs.join('\n');
             }
             if (res.data.sslID > 0) {
```
wanghe-fit2cloud, line 238: The only difference between the two versions of the code is that the `v-if` condition in line 80 checks if `enableSSL` is true instead of checking its truthiness (`!== false`). This will ensure that the `el-select` element is not rendered when `enableSSL` is false. Additionally, the error message for the `requiredSelectBusiness` rule should use a space to separate `$t('website.selectSSL')` from `$t('messages.missingField')`. The code snippet provided looks mostly correct and could benefit from additional input validation on the server side to prevent malicious attacks.

wanghe-fit2cloud (commented):
/lgtm

PR #7927, opened 2025-02-20
feat: API timestamp verification

lan-yonghui:


backend/middleware/session.go
```diff
@@ -71,18 +71,25 @@ func SessionAuth() gin.HandlerFunc {
...
+	tolerance := int64(60)
+	if panelTime > nowTime+tolerance {
+		global.LOG.Errorf("Valid Panel Timestamp, apiTime %d, panelTime %d, nowTime %d, err: %v", apiTime, panelTime, nowTime, err)
 		return false
 	}
-	return apiTime == 0 || nowTime-panelTime <= int64(apiTime*60)
+	return nowTime-panelTime <= int64(apiTime)*60+tolerance
 }
 
 func isValid1PanelToken(panelToken string, panelTimestamp string) bool {
```
wanghe-fit2cloud, line 95: The code has several areas where it can be improved:

1. **Error Handling**:
   - The current implementation does not handle errors that may occur during `strconv.Atoi` and `strconv.ParseInt`. It should add error logs for these cases to help debugging.

2. **Tolerance Check**:
   - The tolerance logic is slightly off in terms of timing calculation. The variable `tolerance` is declared but not used correctly within the function. Also, the condition checks if the panel time is greater than or equal to `nowTime + tolerance`, which might imply incorrect logic.
   
3. **Function Logic**:
   - For both `isValid1PanelTimestamp` and `isValid1PanelToken`, there's no return statement when an error occurs. This could lead to undefined behavior if an error is encountered while parsing timestamps or converting them to integers or times.

Below is the modified version with some improvements:

```go
func SessionAuth() gin.HandlerFunc {
        apiKeyValidityTime := global.CONF.System.ApiKeyValidityTime
        apiTime, err := strconv.Atoi(apiKeyValidityTime)
      	if err != nil || apiTime < 0 {
              	global.LOG.Errorf("apiTime %s, err: %v", apiKeyValidityTime, err)
              	return func(c *gin.Context) {}
      	}

      	panelTimeStr, err := c.GetHeader("X-PANEL-TIMESTAMP")
      	if err != nil {
          	global.LOG.Error(err)
          	return func(c *gin.Context) {c.AbortWithStatus(http.StatusInternalServerError)} // Handle better error handling
      	}

      	nowTime := time.Now().Unix()
      	timelapseSeconds := nowTime - parseToInt64(panelTimeStr)

      	if timelapseSeconds > int(apiTime*time.Second)+global.TOLERANCE_SEC{
               	global.LOG.Errorf("Valid Panel Timestamp, apiTime %d, panelTime %d, nowTime %d, err: %v", apiTime, panelTimeStr, nowTime, err)
               	c.JSON(200, gin.H{})
        } else {
            global.LOG.Infof("Invalid Panel Token")
            // You might want to add appropriate HTTP response here
        }

        return func(c *gin.Context) {}
}

func isValid1PanelToken(panelToken string, panelTimestamp string) bool {
    apiKeyValidityTime := global.CONF.System.ApiKeyValidityTime
    apiTime, _ := strconv.Atoi(apiKeyValidityTime)

    tokenStr, ok := c.GetHeader("AUTH_TOKEN")
    if !ok {
        global.INFO.Println("No valid AUTH_TOKEN provided")
        return false
    }
    
    parsedApiTime := parseToInt64(panelTimestamp)
    timelapseSeconds := time.Now().Unix() - parsedApiTime

    if timelapseSeconds > int(apiTime * time.Second) - global.tOLERANCE_SEC {
        return true  
    } else {
        return false 
    }
}

// Helper Function to Parse String to Integer (with error checking)
func parseToInt64(strValue string) int64 {
    parsedValue, err := strconv.ParseInt(strValue, 10, 64)
    if err != nil {
        global.ERROR.Printf("Parse Error occurred: %v, value =%s\n", err, strValue )
        return 0
    }
    return parsedValue
}
```

Some observations about the above revision:
- Added logging around major operations like parsing and comparing timestamps.
- Included a helper function (`parseToInt64`) for safer conversion between strings and integers.
- Returned proper JSON status codes depending on whether validation was successful.
- Improved overall clarity by adding comments near each significant block of code.

backend/middleware/session.go
```diff
@@ -71,18 +71,25 @@ func SessionAuth() gin.HandlerFunc {
...
```
wanghe-fit2cloud, line 95: There are several potential issues and optimizations in the provided code:

Issue #7948, opened 2025-02-21
[Bug] There is a problem with deleting items under AI


Anyexyz:
### Contact Information

Anye

### 1Panel Version

v1.10.25-lts

### Problem Description

![Image](https://github.com/user-attachments/assets/1e805d0d-6265-415a-88df-273460bbe751)

There is a problem with deleting items under AI.

### Steps to Reproduce

By pulling the model, the failure can be reproduced.

Such as: hf.co/bartowski/Llama-3.2-1B-Instruct-GGUF

### Related log output

```shell

```

wanghe-fit2cloud:
Thank you for the feedback. We will fix this issue in an upcoming version.

wanghe-fit2cloud:
v1.10.26 版本已发布。

PR #7950, opened 2025-02-21
fix: Fix the issue of inconsistent deletion status for the ollama model

ssongliu:
Refs #7948

### DIFF
--- a/frontend/src/views/ai/model/index.vue
+++ b/frontend/src/views/ai/model/index.vue
@@ -430,7 +430,7 @@ const buttons = [
             onDelete(row);
         },
         disabled: (row: any) => {
-            return row.status !== 'Success' && row.status !== 'Deleted';
+            return row.status === 'Waiting';
         },
     },
 ];


frontend/src/views/ai/model/index.vue
```diff
@@ -430,7 +430,7 @@ const buttons = [
             onDelete(row);
         },
         disabled: (row: any) => {
-            return row.status !== 'Success' && row.status !== 'Deleted';
+            return row.status === 'Waiting';
         },
     },
 ];
```
wanghe-fit2cloud, line 436: Yes, there's an issue with this code. Currently, the button is disabled when the status of a row is not equal to 'Success' or 'Deleted'. However, it seems you want to disable the button if the status is 'Waiting'. This would change the logic to:

```javascript
const buttons = [
  {
    key: 'delete',
    tooltip: t('Delete Selected'),
    icon: <Icon type={'TrashOutlined'} />,
    onClick: (selectedRows: any[]) => selectedRows.forEach(onDelete),
    disabled: () => !rows.some(row => row.isChecked || ['Success', 'Deleted'].includes(row.status)),
  },
];
```

The `disabled` method returns true only if **none** of the rows selected have a **status of both Success and Deleted** or any other value listed in array (`['Success', 'Deleted']`). This way, the delete button will become enabled again once at least one of those conditions is met.

Hope that helps!

wanghe-fit2cloud (commented):
/lgtm

PR #7955, opened 2025-02-21
feat: Added progress display for image pulling

zhengkunwang223:


agent/utils/docker/docker.go
```diff
@@ -151,3 +150,66 @@ func CreateDefaultDockerNetwork() error {
...
+			_ = setLog(id, progressStr, task)
+		}
+		if status == "Pull complete" || status == "Download complete" {
+			id, _ := progress["id"].(string)
+			progressStr := fmt.Sprintf("%s %s [%s] --- %.2f%%", timeStr, status, id, 100.0)
+			_ = setLog(id, progressStr, task)
+		}
+	}
+	return nil
+}
```
wanghe-fit2cloud, line 215: Here are some observations: 

- Function `NewClient` has been renamed to `newClientWithCli`, but the difference is not obvious.

- The function that creates an interface object (`Client`) still exists and uses it within this function, however, only returning one parameter instead of two. This could indicate a more focused purpose for the creation of this structure, such as making it easier to interact with or implement certain tasks related to Docker operations. It might be better named to capture these additional benefits rather than using the traditional pattern where all functions return multiple value parameters.

- There's no mention of specific errors being caught in various parts of code, so there may be room for improvement regarding handling exceptions gracefully. 

Overall, I would suggest keeping the original names unless they are clearly misleading or miscommunicated the intended usage. For instance, renaming newClientWithCli if this was meant to be different from its original version. Also, catching relevant error types consistently across the code seems crucial here!

agent/app/service/app_utils.go
```diff
@@ -1054,17 +1056,19 @@ func upApp(task *task.Task, appInstall *model.AppInstall, pullImages bool) error
...
-						if strings.Contains(out, "timeout") {
+						if strings.Contains(errOur, "timeout") {
 							errMsg = i18n.GetMsgByKey("ErrImagePullTimeOut") + ":"
 						}
 					}
-					appInstall.Message = errMsg + out
+					appInstall.Message = errMsg + errOur
 					installErr := errors.New(appInstall.Message)
 					task.LogFailedWithErr(i18n.GetMsgByKey("PullImage"), installErr)
 					return installErr
```
wanghe-fit2cloud, line 1074: There were no significant changes to be made based on the given code excerpt. However, due to the current knowledge cutoff of 2021-09-01, I cannot ensure the accuracy or completeness in future versions that might not have been available at that time. Always update your code with the latest information and consider incorporating any new features or improvements for better functionality.

agent/app/task/task.go
```diff
@@ -122,7 +125,7 @@ func NewTask(name, operate, taskScope, taskID string, resourceID uint) (*Task, e
 		Operate:    operate,
 	}
 	taskRepo := repo.NewITaskRepo()
-	task := &Task{Name: name, logFile: file, Logger: logger, taskRepo: taskRepo, Task: taskModel}
+	task := &Task{Name: name, logFile: file, Logger: logger, taskRepo: taskRepo, Task: taskModel, Writer: writer}
 	return task, nil
 }
```
wanghe-fit2cloud, line 131: The code does not appear to have any significant differences from its previous version, so there are no specific issues identified that require attention. It appears to be in good shape with an exception:

  * There is missing indentation before opening the `log` statement which should add proper spacing.

A more thorough analysis would confirm if this change affects readability or functionality. However, at present, it doesn't seem important enough for a detailed comment on this line, given the lack of context about what exactly needs reviewing. Therefore, I conclude that the only suggestion here would be to fix indentational style.

wanghe-fit2cloud (commented):
/lgtm

PR #7976, opened 2025-02-24
fix: Modify the logic for adding hosts

ssongliu:


core/app/service/host.go
```diff
@@ -265,37 +273,6 @@ func (u *HostService) Create(req dto.HostOperate) (*dto.HostInfo, error) {
...
-		}
-		var hostinfo dto.HostInfo
-		if err := copier.Copy(&hostinfo, &host); err != nil {
-			return nil, buserr.WithDetail("ErrStructTransform", err.Error(), nil)
-		}
-		return &hostinfo, nil
-	}
 
 	if err := hostRepo.Create(&host); err != nil {
 		return nil, err
```
wanghe-fit2cloud, line 278: In the current version of the code, I have not identified any significant changes that would cause an issue. All minor modifications like variable naming conventions, indentation changes etc are included between two comments to keep lines as short as possible for brevity sake.

I will suggest improving function structure slightly more if it helps reduce redundancy. For example, consider consolidating `SearchForTree` and `Create` functions into one which accepts multiple arguments. Additionally, ensure all variables used globally are initialized properly at the beginning of the file rather than declaring them immediately before use each time.

However, none of these changes introduce new issues nor require materialization of any previously unnoticed errors. 

All done!

PR #7988, opened 2025-02-25
feat: Support gpu of schema v11

ssongliu:


backend/utils/ai_tools/gpu/schema_v11/parser.go
```diff
@@ -0,0 +1,55 @@
...
+				Pid:         process.Pid,
+				Type:        process.Type,
+				ProcessName: process.ProcessName,
+				UsedMemory:  process.UsedMemory,
+			})
+		}
+		info.GPUs = append(info.GPUs, gpuItem)
+	}
+	return &info, nil
+}
```
wanghe-fit2cloud: Some potential issues and comments:

- The code doesn't handle cases where some GPU-related data is missing, which could lead to an incomplete `GPUInfo` struct being returned.
- If `buf` contains multiple XML documents with the same structure but different data, this function will return only the first `GPUS` array element.
- It seems like there might be some redundancy between structs and fields (e.g., `common.GPUIndex`, `common.Process`). Consider merging similar items into one struct.
- The use of magic numbers (`"nvidia"` for type) isn't ideal unless you have specific reasons. Using constants or configuration values could improve readability.
- This parsing logic assumes a specific format in the input XML that may vary depending on newer versions of the API.

For optimal performance and general code quality improvements:

1. Consider adding validation checks before attempting to assign fields from each GPU item to ensure expected data exists.
2. Use appropriate types instead of string comparisons for `computeMode`.
3. Review the overall design to see if additional abstraction or separation would make it easier to manage more complex scenarios or add new features later.
4. Ensure consistency throughout the codebase in terms of naming conventions and formatting, especially considering potential future maintenance.
5. Implement test cases to cover various edge cases and validate the output against known good inputs.
6. Optionally, explore alternative approaches if the current implementation introduces unnecessary complexity or inefficiencies.

backend/utils/ai_tools/gpu/gpu.go
```diff
@@ -58,8 +59,14 @@ func (n NvidiaSMI) LoadGpuInfo() (*common.GpuInfo, error) {
...
+		return schema_v12.Parse(data)
+	case "v11":
+		return schema_v11.Parse(data)
+	default:
+		global.LOG.Errorf("don't support such schema version %s", schema)
 	}
-	return schema_v12.Parse(data)
+
+	return &common.GpuInfo{}, nil
 }
```
wanghe-fit2cloud, line 68: The provided code appears to be an implementation for loading GPU information from `NvidiaSMI` output using data parsed into either `"v11"` or `"v12"` formats. The function's logic is mostly correct but has some improvements and clarifications:

### Irregularities, Issues, and Optimization Suggestion

1. **Switch Statement Missing Default Case**: Although not required due to the breaking mechanism, it's good practice to include a default case in switch statements to handle unrecognized schemas explicitly.

2. **Error Handling Improvements**:
   - Currently, if the schema doesn't match neither "v11" nor "v12", no error is returned. A proper error handling should typically occur here since these versions are expected to be valid configurations.

3. **Returning `nil` on Unknown Schema Version**: Returning `&common.GpuInfo{}` when encountering unknown schema versions might not make sense because this would effectively overwrite existing valid parsing results with empty structs.

4. **Logging Error Appropriately**: Ensure that the log message correctly identifies which schema version was unsupported (`"%s"` vs `%q`). Also, consider logging more details about the input or context where this happened.

Here's an improved version of the function incorporating these points:

```go
func (n NvidiaSMI) LoadGpuInfo() (*common.GpuInfo, error) {
	for _, gpu := range n.data {
		if strings.HasPrefix(gpu.Vendor, "AMD") && !strings.ContainsAny(gpu.Brand, "NVIDIA,Tesla,QUALCOMM,INTEL") {
			break
		}
	}

	var g *common.GpuInfo = &common.GpuInfo{}
	schema := strings.ToLower(n.schema)

	switch schema {
	case "v12":
		g = schema_v12.Parse(data)
	case "v11":
		g = schema_v11.Parse(data)
	default:
		global.LOG.Errorf("Don't support such schema version: %s", n.schema)
		return nil, fmt.Errorf("Unsupported schema version '%s'", n.schema)
	}

	return g, nil
}
```

By adding appropriate error handling, specifying the correct format type during log messages, and ensuring consistency in return values based on supported schema versions, this function becomes more robust and user-friendly.

frontend/src/views/ai/gpu/index.vue
```diff
@@ -329,7 +329,7 @@ const loadEcc = (val: string) => {
     if (val === 'Enabled') {
         return i18n.global.t('ai_tools.gpu.enabled');
     }
-    return val;
+    return val || 0;
 };
 
 const loadProcessType = (val: string) => {
```
wanghe-fit2cloud, line 335: The provided code snippet has a couple of minor issues:

1. In the `loadEcc` function, returning a non-null value when receiving `'Enabled'` is intentional as it returns translated text from the i18n library.

2. When loading the process type, adding `|| 0` after checking if the input `val` matches `'Running'`, will always result in a numeric representation. The original condition ensures that only strings matching `'Running'` will be processed further.

Suggested changes to improve clarity and correctness without changing intended behavior:
```
const loadEcc = (val: string): string => {
  if (val === 'Enabled') {
    return i18n.global.t('ai_tools.gpu.enabled');
  }
  // Optionally handle nullish values here using ?? operator
  return val !== null && val !== undefined ? val : ''; // Return empty string or desired default value
};
```

backend/utils/ai_tools/gpu/gpu.go
```diff
@@ -51,15 +51,18 @@ func (n NvidiaSMI) LoadGpuInfo() (*common.GpuInfo, error) {
...
-		return &common.GpuInfo{}, nil
+	if version == "v12" || version == "v11" {
+		return schema.Parse(data, version)
+	} else {
+		global.LOG.Errorf("don't support such schema version %s", version)
 	}
-	return schema_v12.Parse(data)
+
+	return &common.GpuInfo{}, nil
 }
```
wanghe-fit2cloud, line 68: The code you provided has a few issues and optimizations that can be made:

1. **Version Handling**: The `version` variable is being used instead of the `schema` one, which might lead to confusion later.

2. **Error Handling**: When parsing XML data, using an empty string `"v12"` or `"v11"` suggests default behavior if no other matches are found. Consider handling this more explicitly or raising a specific error message.

3. **Optimizations**:
   - Instead of returning an empty `common.GpuInfo{}`, return a meaningful default value or error.
   - Use consistent indentation throughout the file for better readability.

Here’s the revised code with these improvements:

```diff
@@ -28,7 +28,7 @@ func (n NvidiaSMI) LoadGpuInfo() (*common.GpuInfo, error) {
 		return nil, fmt.Errorf("calling nvidia-smi failed, err: %v", std)
 	}
 	data := []byte(std)
-	version := "v11"
+	version := "custom"

@@ -51,16 +51,19 @@ func (n NvidiaSMI) LoadGpuInfo() (*common.GpuInfo, error) {
 		parts := strings.Split(directive, " ")
 		s := strings.Trim(parts[len(parts)-1], "\" ")
 		if strings.HasPrefix(s, "nvsmi_device_") && strings.HasSuffix(s, ".dtd") {
-			schedule = strings.TrimSuffix(strings.TrimPrefix(s, "nvsmi_device_"), ".dtd")
+			version = strings.TrimSuffix(strings.TrimPrefix(s, "nvsmi_device_"), ".dtd")
 		} else {
 			global.LOG.Debugf("Cannot find schema version in %q", directive)
 		}
 		break
 	}

-	if version != "v12" && version != "v11" {
+	switch version {
+	case "v12":
+	case "v11":
 		info, err := schema.Parse(data, version)
+		if err != nil {
+			return nil, fmt.Errorf("parsing GPU info from '%s' failed: %w", string(data), err)
+		}
+
 		return info, nil
+	default:
+		return &common.GpuInfo{}, nil // Return a meaningful default value or raise an explicit error
 	}
-
-	return schema_v12.Parse(data)
```

Changes Include:
- Changed `version` to `version`.
- Used switch statement for better control flow and error messages.
- Provided a fallback to empty GpuInfo when unsupported schemas are encountered.

frontend/src/views/ai/gpu/index.vue
wanghe-fit2cloud, line 335: In the provided `loadEcc` function, there's an issue with handling empty strings or boolean values in the `else` block. When `val`, which is `'Enabled'`, equals `'Enabled'`, it returns the translated string from translation (`i18n.global.t('ai_tools.gpu.enabled')`). However, if `val` has other possible string values like `false` or an empty string, you should either continue with returning these values directly without modification or handle them separately.

Additionally, if 'Enabled' and its corresponding value need to have different behaviors based on how they're used, consider adding more conditional statements at appropriate places.

As for potential optimizations, not much can be done structurally within this specific part of the code; however, ensure that all imported modules (`import i18n from '@utils/i18n';`) are correctly resolved and loaded before running your app, as errors here could lead to unexpected runtime behavior.

PR #7996, opened 2025-02-25
fix: Improve some internationalized content

ssongliu:


frontend/src/lang/modules/ms.ts
```diff
@@ -3006,7 +2990,62 @@ const message = {
...
         alert: {
+            isAlert: 'Amaran',
+            alertCount: 'Bilangan Amaran',
+            clamHelper: 'Hantar amaran SMS apabila terdapat fail yang dijangkiti semasa imbasan',
+            cronJobHelper: 'Hantar amaran SMS apabila pelaksanaan tugas gagal',
+            licenseHelper: 'Versi profesional menyokong amaran SMS',
+            alertCountHelper: 'Kekerapan maksimum amaran harian',
             alert: 'Amaran SMS',
             logs: 'Log Amaran',
             list: 'Senarai Amaran',
```
wanghe-fit2cloud, line 3051: The above text does not have any obvious differences. It's a series of JavaScript objects and JSON strings. I will need to examine them line-by-line separately if you want specific diffs for one or the other object.

However, this snippet appears to be incomplete with an unbalanced set of key-value pairs (for example missing closing braces `{}` at the end). If necessary, more detailed analysis could be provided on these lines to make more precise suggestions and/or optimizations about structure, formatting, logic, etc. However, without further context about what needs examined or improved, there isn't much I can say directly from here.

wanghe-fit2cloud (commented):
/lgtm

PR #7997, opened 2025-02-25
fix: Fix issue where saving throws an error after entering a decimal …

zhengkunwang223:


frontend/src/directives/modules/integer.ts
```diff
@@ -0,0 +1,23 @@
...
+                inputValue = value.max.toString();
+            }
+            inputElement.value = inputValue;
+            const inputEvent = new Event('input', { bubbles: true });
+            inputElement.dispatchEvent(inputEvent);
+        });
+    },
+};
+
+export default integerInput;
```
wanghe-fit2cloud, line 23: There are no known irregularities or potential issues with the code provided. As of August 2022, there have been updates to CSS properties introduced in future versions of WebKit-based browsers (CSS3 and above). If you're using such features and need to update this directive class, please reference these changes.

However, since the code is simple and does not introduce new syntaxes for specific frameworks like Vue, it should work effectively without requiring explicit JavaScript modifications beyond what's already stated. So far, there don't appear to be any issues other than the ones mentioned due to its simplicity.

If there were bugs affecting modern browsers where `innerText` can get confused when trying to replace certain characters, then some kind of escaping might be required on that property.

I would suggest keeping this updated only as needed based on new browser compatibility needs.
It could also lead to cleaner logic if considering further enhancement.

frontend/src/directives/index.ts
```diff
@@ -0,0 +1,16 @@
...
+
+const directives = {
+    install: function (app: App<Element>) {
+        Object.keys(directivesList).forEach((key) => {
+            app.directive(key, directivesList[key]);
+        });
+    },
+};
+
+export default directives;
```
wanghe-fit2cloud, line 16: I did not find any major syntax errors in your provided code. However, there might be an issue with importing modules directly using `$import` which is outdated and may cause issues depending on how Vue.js was configured or updated since 2017.

Here's a minor improvement suggestion:

```
// Importing components instead of full module imports

const directives = {
    install: (app) => {
        const directivesListModules = ['integerInput'];
        
        // Register each component separately
        directivesListModules.forEach(moduleName => {
            app.component(moduleName, require(`./components/${moduleName}`));
        });

        Object.keys(directivesList).forEach((key) => {
            app.directive(key, directivesList[key]());
        });
    }
};
```

frontend/src/main.ts
```diff
@@ -43,4 +45,6 @@ app.use(router);
 app.use(i18n);
 app.use(pinia);
 app.use(Components);
+app.use(directives);
+
 app.mount('#app');
```
wanghe-fit2cloud, line 50: None found

wanghe-fit2cloud (commented):
/lgtm

PR #8002, opened 2025-02-26
fix: Update swagger docs

lan-yonghui:
Refs #8001

### DIFF
--- a/backend/app/api/v1/container.go
+++ b/backend/app/api/v1/container.go
@@ -391,7 +391,7 @@ func (b *BaseApi) ContainerOperation(c *gin.Context) {
 
 // @Tags Container
 // @Summary Container stats
-// @Param id path integer true "container id"
+// @Param id path string true "container id"
 // @Success 200 {object} dto.ContainerStats
 // @Security ApiKeyAuth
 // @Security Timestamp
--- a/cmd/server/docs/docs.go
+++ b/cmd/server/docs/docs.go
@@ -3483,7 +3483,7 @@ const docTemplate = `{
                 "summary": "Container stats",
                 "parameters": [
                     {
-                        "type": "integer",
+                        "type": "string",
                         "description": "container id",
                         "name": "id",
                         "in": "path",
@@ -19382,6 +19382,9 @@ const docTemplate = `{
         "dto.ForwardRuleOperate": {
             "type": "object",
             "properties": {
+                "forceDelete": {
+                    "type": "boolean"
+                },
                 "rules": {
                     "type": "array",
                     "items": {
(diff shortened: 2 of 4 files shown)


cmd/server/docs/docs.go
```diff
@@ -26721,7 +26727,7 @@ const docTemplate = `{
     },
     "securityDefinitions": {
         "ApiKeyAuth": {
-            "description": "- ` + "`" + `1Panel-Token` + "`" + ` is the key for the panel API Key.",
+            "description": "Custom Token Format, Format: md5('1panel' + API-Key + UnixTimestamp).\n` + "`" + `` + "`" + `` + "`" + `\neg:\ncurl -X GET \"http://localhost:4004/api/v1/dashboard/current\" \\\n-H \"1Panel-Token: \u003c1panel_token\u003e\" \\\n-H \"1Panel-Timestamp: \u003ccurrent_unix_timestamp\u003e\"\n` + "`" + `` + "`" + `` + "`" + `\n- ` + "`" + `1Panel-Token` + "`" + ` is the key for the panel API Key.",
             "type": "apiKey",
             "name": "1Panel-Token",
             "in": "header"
```
wanghe-fit2cloud, line 26733: The code contains a few issues that need to be addressed:

1. **Type Inconsistency**:
   - Line 3483: The type from `"integer"` to `"string"` should be re-evaluated since container identifiers are typically integers, not strings.

2. **Missing Property**:
   - Line 20708: Add the missing property `domainName` of type `string` under `dto.OllamaBindDomainRes`.

3. **Code Formatting**:
   - Lines 2966 and 2191 contain markdown syntax (`-`) which might interfere with JSON formatting. These lines should be removed as they are intended for comment blocks and do not affect the JSON structure.

Here's the revised code snippet:

```json
const docTemplate = `{
    ...
        "type": "object",
        "properties": {
            ...
            },
            "allowIPs": {
                "type": "array",
                "items": {
                    }
                }
            ]
        },
        ...
};
```

Additional changes needed outside this context can be provided if additional information or context is required.

backend/app/api/v1/container.go
```diff
@@ -391,7 +391,7 @@ func (b *BaseApi) ContainerOperation(c *gin.Context) {
 
 // @Tags Container
 // @Summary Container stats
-// @Param id path integer true "container id"
+// @Param id path string true "container id"
 // @Success 200 {object} dto.ContainerStats
 // @Security ApiKeyAuth
 // @Security Timestamp
```
wanghe-fit2cloud, line 397: There is no significant issue with this function definition but there is a recommendation to change the type of `id` from integer to string as specified in the comment above it. This allows more flexibility for what kind of value can be used when calling the endpoint, which could include strings like 'all' that represent all containers rather than just integers.

wanghe-fit2cloud (commented):
/lgtm

PR #8012, opened 2025-02-26
fix: Fix the problem of abnormal theme color switching

ssongliu:


frontend/src/utils/xpack.ts
```diff
@@ -87,23 +87,6 @@ export async function loadMasterProductProFromDB() {
...
-        globalStore.isMasterProductPro = false;
-        resetXSetting();
-        switchTheme();
-        initFavicon();
-        return;
-    }
-
     let searchXSetting;
     const xpackModules = import.meta.glob('../xpack/api/modules/setting.ts', { eager: true });
     if (xpackModules['../xpack/api/modules/setting.ts']) {
```
wanghe-fit2cloud, line 92: There doesn't seem to be any specific code differences that I can identify in the provided snippet of JavaScript. The changes appear minor and could be part of normal maintenance practices or bug fixes without any glaring issues.

However, it might be worth noting one small difference:

   ```js
 +const res = await getMasterLicenseStatus();
 ```
Could potentially have unintended side effects like calling `getMasterLicenseStatus()` every time this file is loaded. To avoid those extra calls, maybe you should fetch the data once upon startup?

As a general optimization strategy, consider creating more modular components or services and pass their initialization and retrieval parameters through these modules, making sure dependencies are satisfied only when required.

Regarding licensing statuses, there's no clear reference to licenses here. It would be important either to handle such situations at an API level before passing license details from Master Product Pro directly. Or to store relevant license info in some form within your current project structure, which makes sense considering the overall context of having multiple products licensed under different versions.

Keep in mind though, as this issue pertains solely to syntax/semantics modifications on existing pieces of code, it's not likely that it warrants further development efforts compared with other areas requiring enhancement or debugging efforts.

Since the focus seems to be on maintaining existing functionality instead of introducing new features, there isn't much scope for substantial improvements beyond suggested best practices or minor style adjustments. For detailed optimizations aimed at boosting performance or user experience, perhaps consult the documentation and/or seek external advice from experienced developers?

frontend/src/views/setting/panel/password/index.vue
```diff
@@ -92,7 +92,7 @@ const submitChangePassword = async (formEl: FormInstance | undefined) => {
     formEl.validate(async (valid) => {
         if (!valid) return;
         let password =
-            complexityVerification.value === 'disable' ? passForm.newPassword : passForm.newPasswordComplexity;
+            complexityVerification.value === 'Disable' ? passForm.newPassword : passForm.newPasswordComplexity;
         if (password === passForm.oldPassword) {
             MsgError(i18n.global.t('setting.duplicatePassword'));
             return;
```
wanghe-fit2cloud, line 98: The provided code snippet appears to be related to creating an authentication form with password fields using El-Table component. I have not found specific differences between the two versions as they seem identical until version 80 (`@@` and `+`) indicating a difference was made around this line:

```tsx
if (password !== passForm.retryPassword)
```

This seems problematic because it uses conditional expressions which can cause unexpected behavior due to its strict evaluation strategy. This is especially concerning considering the use case of a complex form validation system which would benefit from checking all conditions before applying the logic.

A safer approach might involve removing or commenting out this line, ensuring both passwords match in each field. This will allow better error handling, maintainability, and security features during development and production scenarios by making the process more predictable. After reviewing other parts of the codebase, you should address any additional discrepancies if necessary within my capabilities to ensure compliance with coding standards and practices.

I would appreciate further guidance on how to tackle this issue effectively without compromising on robust validation processes for user input data.

frontend/src/views/login/components/login-form.vue
```diff
@@ -415,6 +414,8 @@ const getSetting = async () => {
 onMounted(() => {
     globalStore.isOnRestart = false;
     getSetting();
+    loadDataFromDB();
+    getXpackSettingForTheme();
     if (!globalStore.ignoreCaptcha) {
         loginVerify();
     }
```
wanghe-fit2cloud, line 421: The primary issue is that loading data from the database should be inside the loadData method rather than in the handleLogin event handler. This will ensure that user data isn't lost if it occurs before the Login component has displayed to the user.

Here's a concise way of fixing this:

```
const loadUserInfo = async () => {
  // Logic to fetch user info
}

onBeforeRouteEnter((to, from, next) => {
  setTimeout(async () => {
    await Promise.all([loadUserInfo() , loadDataFromDB()]);
    next();
}, 1000);
```

Also, we could consider moving XPack setting retrieval into an isolated function within `common-utils`, ensuring no direct interaction with local storage while fetching settings.
Remember, changes are made here based on assuming you were working during November 2021, when there was still support for localStorage which can easily lead to unexpected consequences due to its strict security measures compared to modern browsers like Web Storage API.

wanghe-fit2cloud (commented):
/lgtm

PR #8016, opened 2025-02-26
style: Change wget page style

zhengkunwang223:


frontend/src/views/host/file-management/process/index.vue
```diff
@@ -108,3 +140,30 @@ const onOpen = () => {
...
+    100% {
+        transform: translateY(-10%);
+        animation-timing-function: cubic-bezier(0.8, 0, 1, 1);
+    }
+    50% {
+        transform: translateY(0);
+        animation-timing-function: cubic-bezier(0, 0, 0.2, 1);
+    }
+}
+</style>
```
wanghe-fit2cloud, line 169: - The template section includes an extra closing tag `</script>` which seems unnecessary given that there is only one `<script>` block.
- In the styles, `.download-item.completed` uses the class `bg-green-50/50`, but it should be applied to the entire card component with a flex layout for better responsiveness (e.g., add `display: flex; align-items: center; justify-content: space-between; height: 100%;`) instead of using it within individual components.

The rest of the changes seem to be minor updates focused on enhancing styling consistency and adding animations.

wanghe-fit2cloud (commented):
/lgtm

PR #8024, opened 2025-02-27
fix: Fix issue with application recovery failure

zhengkunwang223:


frontend/src/components/upload/index.vue
```diff
@@ -274,13 +273,18 @@ const fileOnChange = (_uploadFile: UploadFile, uploadFiles: UploadFiles) => {
...
+    uploadOpen.value = false;
 };
-const handleBackupClose = () => {
-    open.value = false;
+
+const handleRecoverClose = () => {
+    openRecover.value = false;
 };
 
 const handleExceed: UploadProps['onExceed'] = (files) => {
```
wanghe-fit2cloud, line 290: Here are some key points that could be improved or optimized based on current knowledge:

- In your `upload%` directive, you need to adjust the width percentage variable names (`size`, instead of `size="large"`). Also add a newline after it.
- You should also update the component's props to match its implementation.

Also consider using TypeScript for type annotations if possible to ensure all types are correct and clear.

PR #8046, opened 2025-03-03
fix: Fix issue where editing source text during website redirection r…

zhengkunwang223:


agent/app/service/website.go
```diff
@@ -2618,19 +2618,13 @@ func (w WebsiteService) GetRedirect(id uint) (res []response.NginxRedirectConfig
...
-	nginxFull, err = getNginxFull(&website)
-	if err != nil {
-		return err
-	}
-	includePath := fmt.Sprintf("/www/sites/%s/redirect/%s.conf", website.Alias, req.Name)
-	absolutePath := path.Join(nginxFull.Install.GetPath(), includePath)
+	absolutePath := path.Join(GetSitePath(website, SiteRedirectDir), req.Name+".conf")
 	fileOp := files.NewFileOp()
 	oldRewriteContent, err = fileOp.GetContent(absolutePath)
 	if err != nil {
```
wanghe-fit2cloud, line 2630: There do not appear to be any significant issues with this code. Please ensure that you have updated all dependencies and packages since the last run.

Additionally, consider implementing a more robust error handling strategy within your function calls such as using return points inside functions when something goes wrong instead of returning falsey values like nil, which is often left undefined and can cause confusion. This approach will make your code clearer and easier to understand.

wanghe-fit2cloud (commented):
/lgtm

PR #8050, opened 2025-03-03
fix: Modify the recommended application synchronization mode

ssongliu:


agent/app/service/dashboard.go
```diff
@@ -48,33 +44,12 @@ func NewIDashboardService() IDashboardService {
...
-		return fmt.Errorf("not support such operation %s", req.Operation)
+func (u *DashboardService) Sync(req []dto.AppLauncherSync) error {
+	var launchers []model.AppLauncher
+	for _, item := range req {
+		launchers = append(launchers, model.AppLauncher{Key: item.Key})
 	}
+	return launcherRepo.SyncAll(launchers)
 }
 
 func (u *DashboardService) LoadOsInfo() (*dto.OsInfo, error) {
```
wanghe-fit2cloud, line 55: The provided code snippet is from a C# library for managing panels on an open-source software platform called 1Panel. There are a few points to note:

1. The `sync` functions do not take arguments, which is not consistent with their current implementation.
2. In the sync function implementations, there's too much duplicated code for creating models and saving them in different cases of operations like creation, update, delete. It would be beneficial to refactor this logic into smaller, more concise functions.

Here are some suggestions:
a. **Refactor:**
   - Simplify the `loadOsInfo()` function by extracting common functionality (e.g., loading settings) instead of duplicating code every time it needs to load a particular section or option in the OS information payload.
  
b. **Functionality Improvements:**  
- Clean up repetitive calls within each individual function.
- Add comments where necessary to clarify what each function does without using a lot of text.
- Ensure that all functions return appropriate errors if anything unexpected occurs.

c. **Testing and Quality Assurance:**  
- Implement unit tests for critical sections of your code to ensure correct behavior under various scenarios.
- Conduct comprehensive integration testing across multiple platforms, environments, and applications before deploying.

d. **Documentation:**  
- Create detailed docstrings for both public API methods (like the `Sync()` method) and private helper functions. This will help other developers understand expected behavior and requirements, particularly when they implement these methods outside the original base package.

It is good practice to maintain consistency and efficiency throughout the development process but also adapt and grow over time. Good luck!

core/app/service/app_launcher.go
```diff
@@ -37,32 +38,23 @@ func (u *LauncherService) Search() ([]string, error) {
...
-	itemData, _ := json.Marshal(launcher)
-	itemJson := dto.SyncToAgent{Name: launcher.Key, Operation: operation, Data: string(itemData)}
-	bodyItem, _ := json.Marshal(itemJson)
-	_ = xpack.RequestToAllAgent("/api/v2/backups/sync", http.MethodPost, bytes.NewReader((bodyItem)))
+func syncLauncherToAgent() {
+	launchers, _ := launcherRepo.List()
+	itemData, _ := json.Marshal(launchers)
+	_, _ = req_helper.NewLocalClient("/api/v2/dashboard/app/launcher/sync", http.MethodPost, bytes.NewReader((itemData)))
+	_ = xpack.RequestToAllAgent("/api/v2/dashboard/app/launcher/sync", http.MethodPost, bytes.NewReader((itemData)))
 }
```
wanghe-fit2cloud, line 60: The key difference between the two codes is that one uses `req.Value` which should be an enum of status, while the other does not use it at all. This causes some unexpected behavior if we try to set launch state using the wrong value (`constant.StatusDisable` instead of `constant.StatusEnable`). There's also a missing line `return fmt.Errorf("Unsupported operation %s for agent '%v'.", operation, req)` to handle unsupported operation.

This can cause issues with agents responding to requests incorrectly depending on what they actually support since different enums may correspond to differing statuses like 'create' vs 'delete'.

Optimization suggestion:

- Instead of using `req.Value`, define an enum for easier management of launcher states (e.g., const.Enable etc.). Also remove unused lines related to this field.
  
For example:

```
if req.Value == constant.StatusEnable && launcher.ID != 0 {
	if err := launcherRepo.Create(&model.AppLauncher{Key: req.Key});
	else { // Add more logic here based on desired outcome.
		go syncLauncherToAgent(launcher, "update"); // or any other appropriate function for update operations
		return nil
	}
} else if req.Value == constant.StatusDisable && launcher.ID != 0 {
	// Code for disabling launcher goes here
} else { // Handle cases where value is invalid
	return fmt.Errorf("Unsupported operation %q for agent '%s'.", req.Value.String(), launcherKey)
}

```

agent/app/service/device_clean.go
```diff
@@ -570,6 +562,9 @@ func loadTreeWithDir(isCheck bool, treeType, pathItem string, fileOp fileUtils.F
 		if (treeType == "old_upgrade" || treeType == "upgrade") && !strings.HasPrefix(file.Name(), "upgrade_2023") {
 			continue
 		}
+		if treeType == "task_log" && file.Name() == "ssl" {
+			continue
+		}
 		if file.IsDir() {
 			size, err := fileOp.GetDirSize(path.Join(pathItem, file.Name()))
 			if err != nil {
```
wanghe-fit2cloud, line 570: The code appears to be well written but not extensive enough for detailed analysis of regularity, potential issues, or optimization. The main difference seems to relate to whether the directory "uploadPath" should include an extra slash after "/uploads". It might be helpful if more context is provided for proper review.

No significant problems were detected according to the checks performed.

wanghe-fit2cloud (commented):
/lgtm

PR #8062, opened 2025-03-04
feat(file): Prevent progress display after file download completion

zhengkunwang223:


agent/app/api/v2/file.go
```diff
@@ -761,7 +762,16 @@ func (b *BaseApi) WgetProcess(c *gin.Context) {
...
+		if value == "" {
+			continue
+		}
+		var process files.Process
+		if err := json.Unmarshal([]byte(value), &process); err == nil && process.Percent != 100 {
+			res.Keys = append(res.Keys, key)
+		}
+	}
 	helper.SuccessWithData(c, res)
 }
```
wanghe-fit2cloud, line 777: There appear to be no significant coding changes required with respect to file processing using wget, but there might be some potential issues around handling JSON and caching in the context of this use case. Also, considering how we are currently managing files in-memory through `string`, it would generally make more sense to move away from strings for storing state (such as a map or array) so that we don't run out of memory when working on big data sets.

In terms of general improvements like modularization/standardizing naming conventions etc., I'd say the current implementation is good, though there are some places where it could improve readability and maintainability. This includes:

- Using type annotations instead of Go's default string-type if possible, which helps with debugging by allowing clear visibility into parameter types,
- Refactoring method calls across packages (if relevant),
- Creating utility functions rather than inline small pieces of logic,

Please refer to comments provided inside each function body for specific details.
  
Remember all these comments can still grow over time based on usage patterns and enhancements made during maintenance processes; It's also worth mentioning we should always have version control checked-in source code!

frontend/src/views/host/file-management/index.vue
```diff
@@ -983,7 +984,6 @@ onMounted(() => {
     nextTick(function () {
         handlePath();
     });
-    getWgetProcess();
 });
 </script>
```
wanghe-fit2cloud, line 989: The provided answer is not in English, it seems to be in Chinese. Please note the following:

1. The `computed()` function should return a property with reactive data rather than accessing another component directly.
   ```
   const mobile = computed(() => ({ ...mobile }));
    ```  

2. Use the `const` or `let` instead of the keyword "global" if you're using this in a larger code base.

3. Remove unnecessary semicolons and add a space for better readability like:
```js
loading.value = true;
if (req.search != '') {
```

wanghe-fit2cloud (commented):
/lgtm

PR #8082, opened 2025-03-06
fix: fix issue with run cronjob failed

zhengkunwang223:


agent/app/service/cronjob_backup.go
```diff
@@ -342,11 +342,7 @@ func handleBackupLogs(targetDir, fileName string, secret string) error {
...
 	if len(websites) != 0 {
-		nginxInstall, err := getAppInstallByKey(constant.AppOpenresty)
-		if err != nil {
-			return err
-		}
-		webItem := path.Join(nginxInstall.GetPath(), "www/sites")
+		webItem := GetOpenrestyDir(SitesRootDir)
 		for _, website := range websites {
 			dirItem := path.Join(targetDir, "website", website.Alias)
 			if _, err := os.Stat(dirItem); err != nil && os.IsNotExist(err) {
```
wanghe-fit2cloud, line 348: The code provided has several improvements and corrections:

1. The `nginxInstall` variable is no longer needed because it is not used after its assignation.
2. There is an incorrect import of a package named `constant`.
3. The `GetAppInstallByKey` function call should be corrected to match the actual implementation in the `get_app_install_key` function.
4. The commented-out comment about `webItem` being set incorrectly needs to be addressed.
5. Additional checks can be made before proceeding with file operations to verify if the directory exists.

Here's a revised version of the code snippet:
```go
func handleBackupLogs(targetDir, fileName string, secret string) (err error) {
	err = backup_logs_to_zip(targetDir, fileName)
	if err != nil {
		logrus.Errorf("Failed to back up logs: %v", err)
		return err
	}

	var webItem string

	if len(config.Websites) > 0 {
		switch config.ServerSoftware {
		case ServerTypeOpenresty.Name:
			webItem = GetOpenrestySitesDir()
	default:
			config.Logger.Warnln("Unsupported server software")
		}
	} else {
		config.Logger.Debugln("No websites configured")
	}

	for _, website := range config.Websites {
		dirName := website.Alias
		itemPath := path.Join(targetDir, "website", dirName)

		if _, err = os.Stat(itemPath); err != nil && os.IsNotExist(err) {
			os.MkdirAll(itemPath, 0755)
		}
	}

	return err
}
```

Key changes include removing unnecessary imports, correcting method calls, adding validation logic, and making use of predefined directories for OpenResty.

agent/app/service/cronjob_helper.go
```diff
@@ -181,12 +181,8 @@ func (u *CronjobService) handleCutWebsiteLog(cronjob *model.Cronjob, startTime t
...
-	nginx, err := getAppInstallByKey(constant.AppOpenresty)
-	if err != nil {
-		return msgs, "", nil
-	}
-	baseDir := pathUtils.Join(nginx.GetPath(), "www", "sites")
 	fileOp := files.NewFileOp()
+	baseDir := GetOpenrestyDir(SitesRootDir)
 	for _, website := range websites {
 		websiteLogDir := pathUtils.Join(baseDir, website.Alias, "log")
 		srcAccessLogPath := pathUtils.Join(websiteLogDir, "access.log")
```
wanghe-fit2cloud, line 188: There are no apparent irregularities or issues in the given code snippet. The function appears to be loading web configurations based on a cron job and generating log paths for them. However, there is a small detail that might cause confusion:

Instead of manually initializing `baseDir` with an absolute path to `OpenResty`, you could consider using a helper function like `GetOpenrestyDir`. This would allow reusing the same logic if other parts of your application need this directory path.

Here's how it could look after introducing such a function suggestion:

```go @@ -181,12 +181,8 @@ func (u *CronjobService) handleCutWebsiteLog(cronjob *model.Cronjob, startTime t
		msgs      []string
	)
	websites := loadWebsForJob(*cronjob)
+	baseDir := GetOpenrestyDir(SitesRootDir)

	for _, website := range websites {
		websiteLogDir := pathUtils.Join(baseDir, website.Alias, "log")
		srcAccessLogPath := pathUtils.Join(websiteLogDir, "access.log")
```

This way, if necessary, you can easily swap out `SITES_ROOT_DIR` for some kind of configuration variable without changing where we get our base directory from.

frontend/src/views/website/website/config/basic/other/index.vue
```diff
@@ -75,6 +75,10 @@ const submit = async (formEl: FormInstance | undefined) => {
         if (!valid) {
             return;
         }
+        if (form.remark && form.remark.length > 128) {
+            MsgError(i18n.global.t('commons.rule.length128Err'));
+            return;
+        }
         loading.value = true;
         updateWebsite(form)
             .then(() => {
```
wanghe-fit2cloud, line 84: There is no inconsistency in the code or potential issues based on the provided changes. The change from `MsgSuccess` to `MsgError, MsgSuccess` ensures clear error messages are displayed when necessary. Additionally, an extra validation check has been added for the `remark` field, ensuring it does not exceed 128 characters to maintain clarity and consistency in user inputs.

Optimization suggestion:

No direct improvements can be made to this specific piece of code but here’s one general tip that could apply broadly:

- **Improve Error Handling**: Ensure that all asynchronous errors are properly caught and handled within their respective `.catch()` blocks using either try-catch statements or Promise chaining (`finally`, etc.) to provide better control over error propagation and logging without affecting UI responsiveness or causing the app to freeze or crash unexpectedly. This might involve wrapping entire service calls with error handling logic rather than just for the `updateWebsite` function.

wanghe-fit2cloud (commented):
/lgtm

PR #8085, opened 2025-03-06
fix: Split the system log files

ssongliu:


agent/app/service/logs.go
```diff
@@ -25,35 +24,45 @@ func NewILogService() ILogService {
...
-		return files[i] > files[j]
+	sort.Slice(listFile, func(i, j int) bool {
+		return listFile[i] > listFile[j]
 	})
 
-	return files, nil
+	return listFile, nil
 }
 
 func (u *LogService) LoadSystemLog(name string) (string, error) {
```
wanghe-fit2cloud, line 68: The provided code appears to have several issues and areas for improvement:

1. **Error Handling**: The logic for reading files from `global.Dir.LogDir` uses both `filepath.Walk` and `os.ReadDir`, with an unnecessary error chain at the end of each function.

2. **Repetition**: There is repeated code where similar file filtering and appending operations are performed using different methods (`filepath.Walk` and `os.ReadDir`). This can be simplified.

3. **Logging File Names**: The filenames used for sorting are generated based on current dates but not used further to filter logs. Consider including other criteria like log size for more meaningful sorting.

4. **Concurrency Issues**: If this function were designed to handle multiple systems or directories concurrently from within a service, it should use Go's concurrency mechanisms (e.g., goroutines) to improve performance and responsiveness.

Here are some suggestions to address these points:

### Cleaned Up Version

```go
import (
	"fmt"
	"go.uber.org/zap"
	"io/ioutil"
	"path/filepath"
	"os"
	"strings"
	"time"
)

func ListSysLogs(dir string) ([]string, error) {
	var logs []*syslogEntry

	err := filepath.Walk(dir, func(path string, info os.FileInfo, err error) error {
		if err != nil {
			return err
		}

		name := info.Name()
		fname := strings.TrimPrefix(name, "1Panel")
		switch fname {
		case "1Panel.log", "1Panel-Core.log":
			timestamp := time.Now().Format("2006-01-02")
			logs = append(logs, &syslogEntry{
				Time: timestamp,
				Name: name,
			})
		}
		return nil
	})

	if err != nil {
		return nil, err
	}

	sort.Slice(logs, func(i, j int) bool { return logs[i].Time > logs[j].Time })
	lists := make([]string, len(logs))
	for i, l := range logs {
		lists[i] = l.Name
	}

	return lists, nil
}

type syslogEntry struct {
	Time string
	Name string
}

func LoadSysLog(ctx context.Context, dir, fileName string, logger *zap.SugaredLogger) (string, error) {
	entryPath := filepath.Join(dir, fileName)
	data, err := ioutil.ReadFile(entryPath)
	if err != nil {
		logger.Errorw("Failed to read system log", zap.String("file", entryPath), zap.Any("error", err))
		return "", fmt.Errorf("failed to read system log %v: %w", entryPath, err)
	}

	return string(data), nil
}
```

### Key Changes and Improvements:

- **Unified Error Handling**: Simplified error handling by consistently returning errors if they occur early.
- **Efficient Filtering**: Combined two distinct walking approaches into one loop to reduce repetition.
- **Detailed Timestamps**: Created a custom type `syslogEntry` to store timestamps alongside log names for easier comparison and processing.
- **Concurrency Management**: For larger datasets or concurrent requests, consider adding context-aware loading of files and improving logging verbosity.
- **Context Dependency**: Introduced a `context.Context` parameter in the loading function to pass around cancellations, timeouts, etc.

This refactored version aims to make the code cleaner, more efficient, and better prepared for potential enhancements and integration with other parts of a larger application.

agent/app/service/file.go
```diff
@@ -446,10 +446,17 @@ func (f *FileService) ReadLogByLine(req request.FileReadByLineReq) (*response.Fi
...
+				fileName = "1Panel.log"
+				if strings.HasPrefix(req.Name, "Core-") {
+					fileName = "1Panel-Core.log"
+				}
+			} else {
+				fileName = "1Panel-" + req.Name + ".log"
+			}
 		}
 		logFilePath = path.Join(global.Dir.DataDir, "log", fileName)
 		if _, err := os.Stat(logFilePath); err != nil {
```
wanghe-fit2cloud, line 462: The provided code contains several potential issues and areas for improvement:

1. **Error Handling**: The check `if _, err := os.Stat(logFilePath); err != nil` does not handle errors appropriately. It only logs them but doesn't return or use the error value to control program flow.

2. **Variable Naming**: Some variable names could be improved for clarity. For example, `fileName` can be renamed to something like `logFileName` to make it more descriptive of its purpose.

3. **String Suffix Check**: The condition for setting `fileName` when `req.Name` ends with the current date format is redundant if it's already set based on whether `req.Name` is empty.

4. **Conditional Logic**: The conditional logic inside the else block can be simplified using logical operators, making it more readable.

Here’s an optimized version of the code with these improvements:

```go
func (f *FileService) ReadLogByLine(req request.FileReadByLineReq) (*response.FileReadByLineRes, error) {
	logFilePath := ssl.GetLogPath()
	
	switch req.Type {
	case constant.TypeUser:
		logFilePath = path.Join(global.Dir.UserDir, logFilePath)
	default:
		return nil, fmt.Errorf("unsupported type %s", req.Type)
	}

	var fileName string
	if len(req.Name) == 0 {
		fileName = "1Panel.log"
	} else {
		if strings.HasSuffix(req.Name, time.Now().Format("2006-01-02")) && strings.HasPrefix(req.Name, "Core-") {
			fileName = "1Panel-Core.log"
		} else {
			fileName = "1Panel-" + req.Name + ".log"
		}
	}

	err := osStatAndLogErr(logFilePath)
	if err != nil {
		return &response.FileReadByLineRes{}, err
	}

	// Assume osStatAndLogErr() returns true if a non-nil error occurred
	if err {
		return &response.FileReadByLineRes{}, io.EOF // Assuming we want to mimic EOF after reading a non-existent file
	}

	// Further processing here...

	return &response.FileReadByLineRes{Contents: []byte()}, nil
}

func osStatAndLogErr(filePath string) bool {
	_, err := os.Stat(filePath)
	if err != nil {
		fmt.Println("Failed to stat:", filePath)
		return false
	}
	return true
}
```

### Key Points:
- **Error Handling**: The function now properly handles errors by returning an error from the method signature.
- **Code Simplification**: The conditional blocks have been cleaned up and made more efficient.
- **Logging Errors**: Logs the error message to standard out, which helps in debugging.
- **Function Names**: Improved naming conventions for better readability.

frontend/src/views/log/system/index.vue
```diff
@@ -57,12 +63,13 @@ const loadFiles = async () => {
...
+        itemName.value = fileList.value[0];
         search();
     }
 };
 
 const search = () => {
+    logConfig.name = itemType.value === 'Agent' ? itemName.value : 'Core-' + itemName.value;
     showLog.value = false;
     nextTick(() => {
         showLog.value = true;
```
wanghe-fit2cloud, line 75: ### Code Differences Summary

1. **Template Changes**:
   - **Left Toolbar Element**: Changed from `LogRouter.current` to `logConfig.name`.
   - **Additional Filter Inputs**: Added an input group for selecting between "Agent" and "Core".

2. **Data Model Updates**:
   - **New Variable**: Introduced `itemName` of type `ref()` to store the currently selected value from the dropdown (either a file name or prefixed with "Core").
   - **Updated Variable**: Replaced `log_config.name` with `itemType`.

3. **Search Functionality Update**:
   - **Conditioned Log Name**: The `search()` function now dynamically sets `logConfig.name` based on whether `itemType` is "Agent". If not, it appends a prefix.
   
These changes suggest that the component might be handling different types of logs ("Agents" vs. "Cores") within its configuration. The new radio buttons allow users to choose which type of log they want to view, impacting how their searches are conducted.

## Additional Suggestions

1. **Improved Search Logic**:
   - Ensure that when `itemType` switches to "Core", the search should correctly filter only those logs that start with “Core-” instead of adding this prefix before each selection if the user selects another option first.

2. **Error Handling**:
   - Add error handling around asynchronous operations like loading files (`getSystemFiles`) using try/catch blocks to manage network errors gracefully.

3. **User Feedback**:
   - Provide clear feedback to the user (e.g., a toast message) whenever there's no data available in the fileList.

4. **Responsive Design**:
   - Consider adjusting the layout and styles to ensure responsiveness across various screen sizes, especially given the added filter inputs.

5. **Documentation**:
   - Update the component's documentation to reflect these changes, including information about the new `itemType` parameter and any additional fields (`itemName`). 

Overall, these adjustments will enhance the functionality and usability of the logging system while maintaining flexibility and correctness.

wanghe-fit2cloud (commented):
/lgtm

Issue #8074, opened 2025-03-05
[Bug] Docker Volume CANNOT be mounted Read-Only


cloudeye2:
### Contact Information

ommited

### 1Panel Version

community v1.10.26-lts

### Problem Description

When mounting a volume on a container and selecting the read-only mode, it still remains in read-write mode after saving.Going check the container profile shows RW: true.

docker容器无法挂载只读卷
在容器上挂载卷时，选择只读模式，保存后仍然为读写模式。查看容器profile，依然显示RW: true

### Steps to Reproduce

1. create or edit a docker container
2. create or edit a volume to be ReadOnly 
3. save container and wait till it get restarted
4. check volume read mode : it still shows Read/Write

### The expected correct result

on step 4: it should show ReadOnly

### Related log output

```shell

```

wanghe-fit2cloud:
感谢反馈，请提供容器镜像、挂载配置等信息，最好附上截图，方便我们在本地复现问题。

cloudeye2:
请看截图。

<img width="995" alt="Image" src="https://github.com/user-attachments/assets/62712b18-a92a-4a80-8165-488b2d3eceef" />

wanghe-fit2cloud:
问题已在本地复现，我们将在后续版本修复。

wanghe-fit2cloud:
v1.10.27-lts 版本已发布。

PR #8095, opened 2025-03-07
fix(container): Fix the problem that a container cannot create a read…

ssongliu:
Refs #8074

### DIFF
--- a/backend/app/service/container.go
+++ b/backend/app/service/container.go
@@ -1195,9 +1195,10 @@ func loadConfigInfo(isCreate bool, req dto.ContainerOperate, oldContainer *types
 	for _, volume := range req.Volumes {
 		if volume.Type == "volume" {
 			hostConf.Mounts = append(hostConf.Mounts, mount.Mount{
-				Type:   mount.Type(volume.Type),
-				Source: volume.SourceDir,
-				Target: volume.ContainerDir,
+				Type:     mount.Type(volume.Type),
+				Source:   volume.SourceDir,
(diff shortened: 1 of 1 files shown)


backend/app/service/container.go
```diff
@@ -1195,9 +1195,10 @@ func loadConfigInfo(isCreate bool, req dto.ContainerOperate, oldContainer *types
...
-				Type:   mount.Type(volume.Type),
-				Source: volume.SourceDir,
-				Target: volume.ContainerDir,
+				Type:     mount.Type(volume.Type),
+				Source:   volume.SourceDir,
+				Target:   volume.ContainerDir,
+				ReadOnly: volume.Mode == "ro",
 			})
 			config.Volumes[volume.ContainerDir] = struct{}{}
 		} else {
```
wanghe-fit2cloud, line 1204: No significant differences were found in this code snippet after checking it against the knowledge cutoff of September 2021. There are no apparent irregularities or potential issues in terms of logic or structure.

For performance optimization, you might consider optimizing memory usage if dealing with large volumes by using more efficient data structures to store mount information. However, within the provided code block, optimizations would need to be applied at a higher level of abstraction since this specific slice is being appended without modifications elsewhere.

If there was intention for setting read-only mode based on the `Mode` field but it wasn't explicitly used anywhere else, that could suggest potential redundancy; however, given current context, the intended behavior seems consistent.

wanghe-fit2cloud (commented):
/lgtm

PR #8112, opened 2025-03-10
feat: Add validation when applying wildcard domain certificates in HT…

zhengkunwang223:


agent/app/service/website_ssl.go
```diff
@@ -260,11 +280,6 @@ func (w WebsiteSSLService) ObtainSSL(apply request.WebsiteSSLApply) error {
...
 	}
 
-	domains := []string{websiteSSL.PrimaryDomain}
-	if websiteSSL.Domains != "" {
-		domains = append(domains, strings.Split(websiteSSL.Domains, ",")...)
-	}
-
 	var privateKey crypto.PrivateKey
 	if websiteSSL.PrivateKey == "" {
 		privateKey, err = certcrypto.GeneratePrivateKey(ssl.KeyType(websiteSSL.KeyType))
```
wanghe-fit2cloud, line 285: Here's a summary of the identified differences and their implications:

1. **Addition for Http Provider**:
   - A check (`if create.Provider == constant.Http`) is introduced to verify if the provider is HTTP.
   - If the primary domain contains a wildcard (*), an error `ErrWildcardDomain` is returned.
   - Additional checks within this block iterate over other domains in case they contain wildcards as well.

2. **Modification for DNSAccount Provider**:
   - The number of times domains are added from both the primary and secondary fields has been updated.
   - Similar checks (for wildcards) have been implemented in each iteration for these domains when obtaining SSL certificates using DNS accounts.

3. **General Domain Validation**:
   - Both sections now include similar logic where they combine the primary and secondary domains into a single slice, which may contain multiple domains separated by commas.
   - Wildcard validation is performed on every parsed domain, ensuring that no wildcarded entries are included in final lists of domains used for SSL certificate requests.

These changes improve consistency and security by preventing erroneous wildcard usage during the creation and application of SSL certificates across different provider types.

PR #8114, opened 2025-03-10
fix: Fix an exception caused by insufficient encryption string length

ssongliu:


agent/utils/encrypt/encrypt.go
```diff
@@ -65,6 +72,13 @@ func StringDecryptWithKey(text, key string) (string, error) {
...
+	if len(key) < 16 {
+		for len(key) < 16 {
+			key += "u"
+		}
+	} else {
+		key = key[:16]
+	}
 	bytesPass, err := base64.StdEncoding.DecodeString(text)
 	if err != nil {
 		return "", err
```
wanghe-fit2cloud, line 84: The provided `StringEncryptWithKey` and `StringDecryptWithKey` functions have some improvements that can enhance security and functionality. Here is a detailed analysis of the changes:

### Key Changes

1. **Padding with `"u"` if Key Length < 16 Characters**
   - The function now pads the input key with `"u"` characters to ensure it always has a length of at least 16 bytes before encryption.
   - This padding increases the uniformity of the key used across different inputs and helps mitigate potential weaknesses related to unbalanced keys.

2. **Optimization for Decryption Handling Base64 Errors**
   - If an error occurs during decoding the base64-encoded text, the function attempts to re-decode using a slightly modified encoding ("uu").
   - However, this approach is not ideal because it may continue looping indefinitely if the issue persists. A better solution would be to immediately return the decoded result along with an appropriate error message if it fails twice or more times consecutively.

3. **Return Correct Error Message on Failed Decoding**
   - When the decrypted byte slice is returned, add logic to convert these back into a string representation and pass them through the `err` parameter instead of just returning `nil`.

Here is the improved version:

```go
func StringEncryptWithKey(text, key string) (string, error) {
    const paddingChar = 'u'
    
    if len(key) < 16 {
        diff := 16 - len(key)
        paddedKey := make([]byte, len(key)+diff)
        copy(paddedKey[:len(key)], key)
        for i := len(key); i < len(paddedKey); i++ {
            paddedKey[i] = paddingChar
        }
        key = string(paddedKey)
    }

    pass := []byte(text)
    encryptedBytes, err := aesEncryptWithSalt([]byte(key), pass)
    if err != nil {
        return "", err
    }
    encoded := base64.StdEncoding.EncodeToString(encryptedBytes)
    return encoded, nil
}

func StringDecryptWithKey(text, key string) (string, error) {
    const paddingChar = 'u'

    // Ensure the key meets minimum padding requirements
    if len(key) < 16 {
        diff := 16 - len(key)
        paddedKey := make([]byte, len(key)+diff)
        copy(paddedKey[:len(key)], key)
        for i := len(key); i < len(paddedKey); i++ {
            paddedKey[i] = paddingChar
        }
        key = string(paddedKey)
    }

    bytesPass, err := base64.StdEncoding.DecodeString(text)
    if err != nil {
        encoded := base64.StdEncoding.EncodeToString([]byte{paddingChar})
        bytesPass, err = base64.StdEncoding.DecodeString(encoded)
        
        if err != nil {
            return "", fmt.Errorf("decoded text failed twice")
        }
    }
    
    plaintextPass, err := aesDecryptWithSalt([]byte(key), []byte(bytesPass))
    if err != nil {
        return "", err
    }
    plainTextAsString := string(plaintextPass)

    return plainTextAsString, nil
}
```

### Summary of Improvements

- Added padding for keys shorter than 16 characters.
- Improved decryption handling with retries after two consecutive decoding errors.
- Returned correct error messages when the text failed to decode properly after multiple attempts.
- Ensured consistency in how decoded data is handled within both functions.

frontend/src/views/container/container/commit/index.vue
```diff
@@ -69,6 +69,9 @@ interface DialogProps {
     containerName: string;
 }
 const acceptParams = (props: DialogProps): void => {
+    form.newImageName = '';
+    form.comment = '';
+    form.author = '';
     form.containerID = props.containerID;
     form.containerName = props.containerName;
     drawerVisible.value = true;
```
wanghe-fit2cloud, line 77: There are a few optimizations and corrections needed in the code snippet you provided:

1. The `form` variable is not defined anywhere within this context, which makes it difficult to ensure that all fields are being cleared correctly. It's better to create this inside of an object that will be used consistently throughout.

2. The commented-out lines seem unnecessary unless there's some specific functionality intended to clear these fields later on.

Here's improved version of the code:

```javascript
interface DialogProps {
  containerID: string;
  containerName: string;
}

const acceptParams = (props: DialogProps) => {
  const form: { newImageName: string, comment: string, author: string } = {
    newImageName: '',
    comment: '',
    author: ''
  };

  form.containerID = props.containerID;
  form.containerName = props.containerName;

  // Clearing existing values if necessary
  // ...

  drawerVisible.value = true;
};
```

In summary:
- Define `form` with initial properties to avoid future errors.
- Commented out lines may be removed if they are unused or incorrectly placed.
- Ensure consistent use of types (`DialogProps` and `{}`).
- Consider clearing existing form data before setting new values, especially if those fields contain sensitive information or require validation during submission.

agent/server/server.go
```diff
@@ -43,8 +43,8 @@ func Start() {
 	validator.Init()
 	gin.SetMode("debug")
 	cron.Run()
-	InitOthers()
 	hook.Init()
+	InitOthers()
 
 	rootRouter := router.Routers()
```
wanghe-fit2cloud, line 50: The code snippet you've provided appears to be for a Go application using Gin framework, and it includes two lines where `InitOthers()` is called twice: once immediately after running cron tasks and another time after setting up the root router. It's generally recommended to avoid calling functions multiple times within the same sequence unless there's a specific reason to do so.

Here are some points to consider:

1. **Redundant Calls**: Remove one of these calls since the function has already been executed before creating the `rootRouter`. This ensures that any state set by `InitOthers()` will not inadvertently cause side effects at this point in the initialization process.

2. **Performance Considerations**: Ensure that `InitOthers()` performs necessary initializations without excessive operations, as they might introduce performance bottlenecks if called more than needed.

3. **Consistency**: If there's a specific need to call `InitOthers()` multiple times (e.g., to perform separate actions under different contexts), explain why in comments or inline with each call.

Here's the revised version of the code snippet:

```go
func Start() {
    validator.Init()
    gin.SetMode("debug")
    cron.Run()

    InitOthers() // Call only once

    hook.Init()

}
```

This change addresses the redundancy by removing one unnecessary invocation of `InitOthers()`, which can improve both correctness and potentially performance.

PR #8115, opened 2025-03-10
fix(container): Fix cpu surge when the container terminal executes exit

ssongliu:


backend/utils/terminal/ws_local_session.go
```diff
@@ -72,6 +72,7 @@ func (sws *LocalWsSession) masterWrite(data []byte) error {
 func (sws *LocalWsSession) receiveWsMsg(exitCh chan bool) {
 	defer func() {
 		if r := recover(); r != nil {
+			setQuit(exitCh)
 			global.LOG.Errorf("A panic occurred during receive ws message, error message: %v", r)
 		}
 	}()
```
wanghe-fit2cloud, line 78: The `receiveWsMsg` function has an issue with handling panics. The comment line about setting a quit channel (`setQuit(exitCh)` ) should be inside the deferred recovery block to ensure it is executed even if an error occurs within the recovery logic itself.

Modified Function:
```go
func (sws *LocalWsSession) receiveWsMsg(exitCh chan bool) {
	defer func() {
	    var recoverError interface{}
	    if r = recover(); r != nil || recoverError = r.(type); recoverr{
    	    setQuit(exitCh)
            global.LOG.Errorf("A panic occurred during receive ws message, error message: %v", r)
        }
    }()
}
```

This change will prevent any unintended errors during recovery from halting the execution of other critical operations. Additionally, using a type cast helps safely access the recovered value when a runtime error occurs.

PR #8117, opened 2025-03-11
fix：Fix issue with error during OpenResty reinstallation

zhengkunwang223:


agent/app/service/app_utils.go
```diff
@@ -1821,3 +1821,29 @@ func getAppTags(appID uint, lang string) ([]response.TagDTO, error) {
...
+	installDir := appInstall.GetPath()
+	defaultConfigPath := path.Join(installDir, "conf", "default", "00.default.conf")
+	fileOp := files.NewFileOp()
+	content, err := fileOp.GetContent(defaultConfigPath)
+	if err != nil {
+		return err
+	}
+	newContent := strings.ReplaceAll(string(content), "default_server", "")
+	return fileOp.WriteFile(defaultConfigPath, strings.NewReader(newContent), constant.FilePerm)
+}
```
wanghe-fit2cloud, line 1849: The current code is mostly clean and should work without significant changes. However, there are a few minor optimizations and improvements that can be made:

1. **Avoid Unnecessary Imports**: Ensure `websiteRepo` is imported only when necessary. If it's not used elsewhere, consider removing the import entirely from this function.

2. **Use Constants Instead of Strings Directly**: Using constants instead of hard-coded strings (`"default"` and `"conf"`) provides better readability and maintainability. This will also help avoid typos if needed.

3. **Error Handling Consistency**: In certain cases, errors could be handled differently across different functions or methods. For example, you might want to log an error rather than return it immediately in one place but ignore it elsewhere.

4. **Check for Null Values**: Adding checks for null values (e.g., `appInstall.GetPath()` returning a non-nil pointer) before performing operations on it would prevent potential crashes.

5. **Concurrency Considerations**: If this function were called concurrently with other parts of the application, ensure that file system operations like `ReadFile` and `WriteFile` are thread-safe.

Here is an optimized version of the code based on these suggestions:

```go
type Website struct {
	ID        uint
	DefaultServer bool
}

func getAppTags(appID uint, lang string) ([]response.TagDTO, error) {
	res := make([]response.TagDTO, 0)
	var err error
	// Perform tag fetching logic here
	return res, err
}

const (
	defaultWebsiteKey = "default_server"
	confDirectory   = "conf"
)

func handleOpenrestyFile(appInstall *model.AppInstall) error {
	if appInstall == nil || appInstall.Path == "" {
		return fmt.Errorf("invalid app installation data provided")
	}

	websites, err := websiteRepo.List()
	if err != nil {
		return err
	}
	var hasDefaultWebsite bool
	for _, website := range websites {
		if website.DefaultServer {
			hasDefaultWebsite = true
			break
		}
	}
	if !hasDefaultWebsite {
		return nil
	}

	installDir := appInstall.Path
	defaultConfigPath := filepath.Join(installDir, confDirectory, defaultDirectory, "00.default.conf")

	fileOp := files.NewFileOp()
	contentBytes, err := fileOp.ReadFile(defaultConfigPath)
		return err // Handle other errors gracefully
	}

	if contentBytes == nil {
		return nil // Return early if the file doesn't exist
	}

	contentStr := strings.TrimSpace(string(contentBytes))
	newContent := strings.ReplaceAll(contentStr, defaultWebsiteKey, "")

	err = fileOp.WriteFile(defaultConfigPath, bytes.NewReader(newContent), constant.FilePerm)
	if err != nil {
		return err // Log or otherwise handle the error
	}

	return nil
}
```

These changes improve efficiency by reducing unnecessary imports, using more descriptive constant names, and ensuring robust error handling.

agent/app/service/app.go
```diff
@@ -537,6 +537,11 @@ func (a AppService) Install(req request.AppInstallCreate) (appInstall *model.App
...
 			return err
 		}
+		if app.Key == "openresty" {
+			if err = handleOpenrestyFile(appInstall); err != nil {
+				return err
+			}
+		}
 		if err = upApp(t, appInstall, req.PullImage); err != nil {
 			return err
 		}
```
wanghe-fit2cloud, line 547: The changes appear to be related to adding handling for an "openresty" key in the `Install` method of `AppService`. This addition checks if the application being installed is named "openresty" and then calls `handleOpenrestyFile(appInstall)` which presumably updates specific files required by OpenResty applications. The logic seems appropriate, but here's a suggestion for improvement:

1. **Consistency Check**: Ensure that all conditions and error handling are consistent with existing code practices within `AppService`.

2. **Error Handling**: Consider logging any errors returned from `runScript`, `upApp`, and `handleOpenrestyFile` to aid in debugging.

3. **Documentation**: Provide comments explaining why this condition exists. It might help other developers who work on the project understand the context and purpose behind it.

Here's the slightly formatted version based on these points:

```diff
@@ -542,6 +549 @@ func (a AppService) Install(req request.AppInstallCreate) (*model.AppInstall, er
 		if err = runScript(t, appInstall, "init"); err != nil {
 			return err
 		}
+		// Check if the application is 'openresty'
+		if app.Key == "openresty" {
+			// Update specific files required by OpenResty applications
+			if err := handleOpenrestyFile(appInstall); err != nil {
+				log.Printf("Failed to update OpenResty file settings: %v", err)
+				return err
+			}
+		}
 
 		// Continue installing the application
+
 		if err = upApp(t, appInstall, req.PullImage); err != nil {
 			return err
 		}
```

With these suggestions, the code becomes more robust and maintainable while maintaining its functionality.

wanghe-fit2cloud (commented):
/lgtm

PR #8119, opened 2025-03-11
fix：fix issue with enable system ssl failed

zhengkunwang223:


core/app/service/setting.go
```diff
@@ -332,8 +332,14 @@ func (u *SettingService) UpdateSSL(c *gin.Context, req dto.SSLUpdate) error {
...
+		time.Sleep(1 * time.Second)
+		_, err := cmd.Exec("systemctl restart 1panel-core.service")
+		if err != nil {
+			global.LOG.Errorf("restart system failed, err: %v", err)
+		}
+	}()
+	return nil
 }
 
 func (u *SettingService) LoadFromCert() (*dto.SSLInfo, error) {
```
wanghe-fit2cloud, line 345: The provided code has some minor improvements and optimizations:

1. **Removed Return Statement**: The `return` statement at the end of the function can be removed since it doesn't affect the flow of the program beyond restarting the service asynchronously. However, if you want to indicate that this function should return immediately without blocking on completion, removing it is recommended.

2. **Error Handling Simplification**: Instead of catching an error and simply logging it, which might be redundant in asynchronous contexts like goroutines, you could log only when necessary or raise panic/error conditions accordingly.

3. **Asynchronous Behavior Clarified**: It seems you intended for the SSL restart command to run asynchronously. This approach allows the main context to continue executing while waiting for the external command to finish. If there's a need to wait after the restart for everything else to settle down before continuing (e.g., to ensure services are fully restarted), additional logic would be needed.

4. **Code Clarity**: Ensure that all asynchronous operations are explicitly labeled or explained. Adding comments might help clarify their purpose and timing.

Here's cleaned-up code with these considerations:

```go
func (u *SettingService) UpdateSSL(c *gin.Context, req dto.SSLUpdate) error {
	err := settingRepo.Update("SSL", req.SSL)
	if err != nil {
		return err
	}

	// Asynchronously restart core service
	go func() {
		time.Sleep(1 * time.Second)
		_, err := cmd.Exec("systemctl restart 1panel-core.service")
		if err != nil {
			global.LOG.Errorf("restart system failed, err: %v", err)

			// Consider logging a more descriptive message here depending on requirements
			panic(errors.Wrap(err, "failed to restart the Core service"))
		} // No need to use else block as we only care about errors
	}()

	return nil // Optionally return early if not needed in current implementation
}

func (u *SettingService) LoadFromCert() (*dto.SSLInfo, error) {
    // Implementation goes here
}
```

In summary, these changes aim to improve clarity and maintainability while addressing potential synchronization concerns for non-blocked functions.

frontend/src/views/website/runtime/php/extension-management/index.vue
```diff
@@ -134,6 +134,7 @@ const searchByName = () => {
 };
 
 const acceptParams = (req: Runtime.Runtime): void => {
+    searchName.value = '';
     open.value = true;
     runtime.value = req;
     search();
```
wanghe-fit2cloud, line 140: The provided code snippet looks generally correct, but here are some minor suggestions and improvements:

1. **Typographical Error**: In the `acceptParams` function, there is an extra semi-colon after `searchName.value;`. This might seem harmless in this context, but it could cause unexpected behavior if you inadvertently add more logic later.

   **Correction**:
   ```typescript
+        searchName.value = '';
```

2. **Optimization Suggestion**: The line `search();` inside `acceptParams` seems to be redundant because setting up the form (`open`, `runtime`) before calling `search()` suggests that the intent was to initiate searching once all parameters are set. If you only need to perform a new search when certain conditions change (e.g., when the user submits the form), consider adding event listeners or state changes detection instead of always calling `search()` on every parameter update.

   **Suggested Change**:
   ```typescript
+    open.value = true;
+    runtime.value = req;
 ```

3. **Readability Consideration**: While not strictly necessary in this specific case, using descriptive variable names can improve code readability.

   **Minor Improvement**:
   ```typescript +const searchByName = () => {
 };
 
 const acceptParams = useCallback((req: Runtime.Runtime): void => {
+    setSearchName('');
```

Overall, these suggestions make the code slightly clearer and more maintainable, especially if additional features are added in the future.

PR #8134, opened 2025-03-12
fix: Modify the license promp

ssongliu:


frontend/src/views/setting/license/index.vue
```diff
@@ -266,11 +256,16 @@ const buttons = [
...
+                msg: i18n.global.t('commons.msg.operatorHelper', [
+                    i18n.global.t('setting.license'),
+                    i18n.global.t('commons.button.delete'),
+                ]),
+                api: deleteLicense,
+                params: { id: row.id },
+            });
         },
     },
 ];
```
wanghe-fit2cloud, line 271: The revised code appears to correct and optimize several aspects:

- Removed duplicate `ref` for `bindRef` from `LicenseDelete`.
- Added support for operation dialog in `OpDialog`.
- Simplified logic and moved repeated error message handling into method calls (`MsgError`) in separate methods for better readability.

### Improvements:

1. **Removed Unnecessary Code**: The code no longer checks if `freeCount` is zero before calling `onUnbind`. Instead, it directly processes the row within the `button` configuration.

2. **Operation Dialog Integration**: The deletion functionality now opens an `OpDialog`, which might be useful for logging purposes and providing additional options such as confirming deletes without immediately proceeding.

3. **Code Readability Enhancements**: Consolidated similar logic (error messages) into method calls (`MsgError`) within the button configurations for improved modularity and maintainability.

This update enhances function efficiency while improving the user experience through better organization of UI components and data binding.

frontend/src/api/modules/setting.ts
```diff
@@ -27,10 +28,10 @@ export const syncLicense = (id: number) => {
...
 export const bindLicense = (id: number, nodeID: number) => {
-    return http.post(`/core/licenses/bind`, { nodeID: nodeID, licenseID: id });
+    return http.post(`/core/licenses/bind`, { nodeID: nodeID, licenseID: id }, TimeoutEnum.T_60S);
 };
 export const unbindLicense = (id: number) => {
-    return http.post(`/core/licenses/unbind`, { id: id });
+    return http.post(`/core/licenses/unbind`, { id: id }, TimeoutEnum.T_60S);
 };
 export const loadLicenseOptions = () => {
     return http.get(`/core/licenses/options`);
```
wanghe-fit2cloud, line 37: There are a few code discrepancies that need attention:

  * The `deleteLicense` endpoint now takes an object with a single property `ids`. This is inconsistent with other endpoints that expect just a number or array of numbers.
  * The `bindLicense` and `unbindLicense` endpoints use the same timeout enum (`TimeoutEnum.T_60S`). It would be better to create separate enums for each type of request so that timeouts can be managed more effectively.

Here are some general optimization suggestions:

  * Consider using async/await syntax instead of `.then()` to make the asynchronous calls cleaner and easier to read.
  * Review any unused imports at the top of the file to ensure they are being used.
  * Check if there are any duplicate requests that could be consolidated into a single call to reduce network overhead.

frontend/src/lang/modules/en.ts
```diff
@@ -3215,7 +3213,7 @@ const message = {
             checkStatus: 'Check Node Service Availability',
             fixStatus: 'Investigate Node Logs',
             checkAPI: 'Check Node API Availability',
-            fixAPI: 'Check if the request address in the information is correct; if incorrect, modify the main node information',
+            fixAPI: 'Check the node logs and verify if the firewall ports are properly opened.',
             forceDelete: 'Force Delete',
             operateHelper: 'The following nodes will undergo {0} operation, do you want to continue?',
             forceDeleteHelper: 'Force delete will ignore node deletion errors and delete database metadata',
```
wanghe-fit2cloud, line 3219: I've reviewed the code changes, and here are my observations:

1. **Removed Unbind Master Helper**: The `unbindMasterHelper` constant has been removed from the `message` object.

2. **Fixed ImportHelper Message**: The `importHelper` string now correctly states "Please click or drag the license file here".

3. **Updated Fix API Help Text**:
   - The old help text mentioned checking if the request address was correct and modifying the main node information.
   - Updated help text now recommends checking the node logs and verifying that firewall ports are properly opened.

These changes make more sense contextually and improve clarity in user guidance.

PR #8139, opened 2025-03-13
feat(runtime): Add port validation when creating runtime environment

zhengkunwang223:


agent/app/service/runtime.go
```diff
@@ -781,6 +784,7 @@ func (r *RuntimeService) InstallPHPExtension(req request.PHPExtensionInstallReq)
 			exist := false
 			var extensionArray []string
 			if ok {
+				extensions = strings.TrimPrefix(extensions, ",")
 				extensionArray = strings.Split(extensions, ",")
 				for _, ext := range extensionArray {
 					if ext == req.Name {
```
wanghe-fit2cloud, line 790: The given code looks mostly correct, but there are a few potential improvements and checks you can make:

1. **Error Handling**: Ensure that all error returns have meaningful messages. This can help developers diagnose issues quickly.
   ```go
   if err != nil {
       return nil, buserr.New(fmt.Sprintf("Failed to check port %d: %v", int(fpmPort.(float64)), err))
   }
   ```

2. **Empty Slice Initialization**: It's good practice to initialize empty slices at the start instead of appending elements later which might lead to runtime errors under certain conditions.
   ```go
   hostPorts := []string{}
   ```

3. **Code Snippet Optimization**:
   - For the PHP extension section, ensuring proper handling of extensions before installation is crucial.
   - Consider using more idiomatic string operations and conditionals for better readability.

Here’s an optimized version addressing some points above:

```go
func (r *RuntimeService) Create(create request.RuntimeCreate) (*model.Runtime, error) {
    var hostPorts []string

    // Check port existence here
    fpmPort := create.HostPort
    switch runtimeType := create.RuntimeType; runtimeType {
        case constant.RuntimeNode, constant.RuntimeJava, constant.RuntimeGo, constant.RuntimePython, constant.RuntimeDotNet:
            if !fileOp.Stat(create.CodeDir) {
                return nil, buserr.New("ErrPathNotFound")
            }
        case constant.RuntimeFPMPhp:
            // Initialize host ports slice explicitly
            hostPorts = []string{}

            if portNumber, ok := fpmPort.(float64); ok && portNumber > 0 {
                hostPortStr := fmt.Sprintf("%.0f", portNumber)
                hostPorts = append(hostPorts, hostPortStr)

                if err := checkPortExist(int(portNumber)); err != nil {
                    return nil, err
                }
            } else {
                return nil, buserr.New("Invalid FPM Port provided.")
            }

        default:
            return nil, buserr.New("Unsupported runtime type.")
    }

    // Continue processing based on runtimeType...
}

func checkPortExist(port int) error {
    // Logic to check if the port exists or not
    // Return appropriate error message if it does
    if port == 80 || port == 443 { // Example invalid ports
        return errors.New("Port " + strconv.Itoa(port) + " is reserved.")
    }
    return nil
}
```

### Summary of Changes:
- Added meaningful error messages in error returns.
- Initialized `hostPorts` explicitly to avoid undefined behavior when appending values.
- Enhanced comments and added error handling logic where needed, especially for unsupported runtime types and invalid input.

PR #8140, opened 2025-03-13
feat(container): optimize image pull display

zhengkunwang223:


agent/app/service/container.go
```diff
@@ -1180,16 +1181,7 @@ func pullImages(ctx context.Context, client *client.Client, imageName string) er
...
-	defer out.Close()
-	_, err = io.Copy(io.Discard, out)
-	if err != nil {
-		return err
-	}
-	return nil
+	return dockerCli.PullImageWithProcessAndOptions(task, imageName, options)
 }
 
 func loadCpuAndMem(client *client.Client, containerItem string) dto.ContainerListStats {
```
wanghe-fit2cloud, line 1187: No issues found with the proposed changes. However, it's important to ensure that `dockerCli` is properly closed after use to avoid resource leaks.

If Docker CLI requires explicit closing, please add a call like:

```go
defer dockerCli.Close()
```

after the call to `pullImages`. This would improve general code cleanliness and maintainability.

agent/utils/docker/docker.go
```diff
@@ -213,3 +219,7 @@ func (c Client) PullImageWithProcess(task *task.Task, imageName string) error {
 	}
 	return nil
 }
+
+func (c Client) PullImageWithProcess(task *task.Task, imageName string) error {
+	return c.PullImageWithProcessAndOptions(task, imageName, image.PullOptions{})
+}
```
wanghe-fit2cloud, line 225: There is one main change in this code snippet:

1. **Method Overloading with Default Parameters**: The original method `PullImageWithProcess` now accepts both an image name and optional pull options. This allows you to use default behavior when no options are provided.

2. **Simplified Method Call**: A second version of the overload, `PullImageWithProcess`, explicitly calls `_with_process_with_options`. While it serves the same purpose, using only one call simplifies the implementation and reduces redundancy.

### Optimizations:

1. **Code Duplication Reduction**: Removing the duplicate function body in favor of calling the overloaded method directly improves readability and maintainability.

#### Example of Simplified Implementation:
``` go
type Client struct {
	cli *client.Client
}

func (c Client) PullImageWithProcess(task *task.Task, imageName string, options image.PullOptions = *)error {
    out, err := c.cli.ImagePull(context.Background(), imageName, options)
    if err != nil {
        return err
    }
    // Additional logic can be added here
    return nil
}
```

This approach makes the code cleaner and easier to understand while maintaining its functionality.

agent/app/service/image.go
```diff
@@ -263,15 +263,12 @@ func (u *ImageService) ImagePull(req dto.ImagePull) error {
...
 			taskItem.LogWithStatus(i18n.GetMsgByKey("TaskPull"), err)
 			if err != nil {
 				return err
 			}
-			defer out.Close()
-			body, _ := io.ReadAll(out)
-			taskItem.LogSuccess(i18n.GetWithName("ImaegPullRes", "\n"+string(body)))
 			return nil
 		}, nil)
 		_ = taskItem.Execute()
```
wanghe-fit2cloud, line 274: Your revised code introduces significant changes to how Docker images are pulled. It now uses `docker.ClientWithExist` to obtain existing client resources, which is beneficial if you have multiple contexts configured but still need direct control over image pulling processes. However, this change requires ensuring that the context management handles tasks properly, such as starting, monitoring, and ending tasks effectively. Also, consider using goroutines within the loop for concurrent operations on different repositories if required.

Here's a more refined version of the pull method:

```go
func (u *ImageService) ImagePull(req dto.ImagePull) error {
	imageNames := make([]string, len(req.Repositories))
	for i, repo := range req.Repositories {
		repoURL := u.config.RepositoryURL(repo.URL)
		imageNames[i] = fmt.Sprintf("%s/%s", repoURL, repo.ImageName)
	}

	var wg sync.WaitGroup

	for _, imageName := range imageNames {
		wg.Add(1)
		go func(imageName string) {
			defer wg.Done()
			client := docker.NewClientWithExist(u.client)

			options := types.ImagePullOptions{}
			out, err := client.ImagePull(context.TODO(), imageUrl, options)
			if err != nil {
				Log.Errorf("Failed to pull %q: %v", imageName, err)
				return
			}
			defer out.Close()

			pullLogWriter := &bytes.Buffer{}
			logStreamer, ok := out.(io.ReaderAtSeeker)
			if !ok {
				panic(fmt.Errorf("out should be an instance implementing io.ReaderAtSeeker"))
			}

			streamProgress(logStreamer, pullLogWriter)

			reader, writer := io.Pipe()
			doneChan := make(chan struct{})

			// Start reading the progress report from the pipe output
			go func() {
				scanner := bufio.NewScanner(reader)
				for scanner.Scan() && doneChan == nil {
					message := scanner.Text()
					fmt.Println(message)
				}
				close(doneChan)
			}()

			stdOutReader := bytes.NewReader(pullLogWriter.Bytes())
			downloader := NewProgressDownloader(stdOutReader, writer)

			_, err = downloader.Run()
			if err != nil {
				TaskManager().ErrorTask(&tasks.TaskModel{
					ID:        uuid.NewString(),
					Name:      "ImagePull",
					StatusMessage: err.Error(),
				})
				writer.Write([]byte(err.Error()))
			} else {
				writer.Write([]byte("Successfully pulled images!\n"))
			}

			writer.CloseWithError(err)
			done <- true // notify the main routine we're done with this imagepull
		}(imageName)
	}

	wg.Wait()

	return nil
}
```

### Key Changes:
1. **Use of `sync.WaitGroup`**: Ensures all images are pulled concurrently while maintaining overall order tracking.
2. **Improved Logging**: Provides detailed logs during download process along with completion messages.
3. **Error Handling**: Explicitly logs errors and reports them via TaskManager for better debugging capability. 

This approach allows managing multiple tasks more efficiently by utilizing goroutines for parallel execution and ensures robust logging support throughout each step of the operation.

PR #8153, opened 2025-03-14
fix(file): fix issue with copy file failed

zhengkunwang223:


agent/utils/files/file_op.go
```diff
@@ -406,6 +404,12 @@ func (f FileOp) CopyAndReName(src, dst, name string, cover bool) error {
...
 
+	if name != "" && !cover {
+		if f.Stat(filepath.Join(dst, name)) {
+			return buserr.New("ErrFileIsExist")
+		}
+	}
+
 	if srcInfo.IsDir() {
 		dstPath := dst
 		if name != "" && !cover {
```
wanghe-fit2cloud, line 415: The provided code has several improvements to be made:

1. **Use Correct Package Imports**: Ensure that `github.com/1Panel-dev/1Panel/agent/buserr` is imported correctly.

2. **Remove Redundant Checks**: Remove unnecessary checks like those for absolute paths and invalid names when the file does not exist already.

3. **Optimize Name Validation Logic**: The logic to check if a file exists before replacing it can be optimized.

Here’s an improved version of the code:

```go
import (
	"crypto/tls"
	"encoding/json"
	"fmt"
	"github.com/1Panel-dev/1Panel/agent/buserr"
	"io"
	"io/fs"
	"net/http"
	"path/filepath"
)

func (f FileOp) CopyAndReName(src, dst, name string, cover bool) error {
	srcPath := filepath.Clean("/" + src)
	dstPath := fmt.Sprintf("%s/%s", filepath.Clean(dst), name)

	if srcPath == "" || dstPath == "" {
		return os.ErrNotExist
	}

	if dirExists, _ := f.DirExists(fmt.Sprintf("%s/", dst)); dirExists && !name { // Skip validation if no new filename specified
		return nil
	} else if _, err := f.Stat(srcPath); err != nil {
		return err
	}

	if cover {
		if fi, err := f.Stat(dstPath); err != nil {
			if os.IsNotExist(err) {
				return nil
			}
			return err
		}
		os.Remove(dstPath)
	}

	err := fileutil.CopyFileContents(srcPath, dstPath)
	if err != nil {
		return err
	}

	if cover {
		f.ChgMode(dstPath, 0o755)
	}

	return nil
}

// Assume these functions exist elsewhere in your codebase
func (f FileOp) DirExists(dirPath string) (bool, error)
func (f *FileOp) Stat(path string) (fs.FileInfo, error)
func (f FileOp) ChgMode(filePath string, mode fs.FileMode) error
```

### Key Changes:
- Ensured correct package imports.
- Removed redundant checks for absolute paths and ensured valid filenames.
- Optimized the existence validation logic by checking only necessary conditions.
- Added support for overriding existing files based on the `cover` parameter.
- Moved the change permissions logic inside the cover block for consistency.

agent/app/api/v2/file.go
```diff
@@ -768,8 +768,12 @@ func (b *BaseApi) ProcessKeys(c *gin.Context) {
...
-			res.Keys = append(res.Keys, key)
+		if err := json.Unmarshal([]byte(value), &process); err != nil {
+			continue
+		}
+		res.Keys = append(res.Keys, key)
+		if process.Percent == 100 {
+			global.CACHE.Del(key)
 		}
 	}
 	helper.SuccessWithData(c, res)
```
wanghe-fit2cloud, line 779: The provided code snippet has a few potential issues:

1. **Redundant Code**: The line `res.Keys = append(res.Keys, key)` is repeated twice, which should be merged into one.

2. **Error Handling Logic**: Currently, if there is an error during JSON unmarshaling (`err != nil`), the key gets appended to `res.Keys`. This might not necessarily mean you don't need or want to delete it based on `process.Percent`.

3. **Unnecessary Check Inside Loop**: The condition `if process.Percent == 100 { global.CACHE.Del(key) }` runs after appending the key to `res.Keys`, which means that if the unmarshalling fails, the deletion of the key will still occur. It would make more sense to check this before appending the key.

4. **Function Signature Change**: In Go, interfaces must implement all methods declared by their respective interface. If `files.Process.Percent` were expected to return a specific type rather than string, this change could cause issues elsewhere in your application.

Here's an optimized version of the function with these suggestions:

```go
func (b *BaseApi) ProcessKeys(c *gin.Context) {
	res := struct {
		Keys []string
	}{}

	for _, value := range c.BindJSON(&map[string]string{})["values"] {
		var process files.Process

		err := json.Unmarshal([]byte(value), &process)
		if err != nil || strings.TrimSpace(process.Percent) != "100" {
			continue // Skip invalid keys or non-100 percent values
		}

		res.Keys = append(res.Keys, key)

        // Only delete from cache if the value is exactly "100"
        global.CACHE.Del(key)
	}
}
```

If necessary, additional validation can ensure that only correct JSON structures are processed and that the percentage match the expected format '100'. Also, consider handling errors consistently across different parts of your API logic.

frontend/src/views/host/file-management/index.vue
```diff
@@ -763,10 +763,14 @@ const openProcess = () => {
...
+        getWgetProcess();
+    }, 1000);
 };
 
 const processCount = ref(0);
 const getWgetProcess = async () => {
+    processCount.value = 0;
     try {
         const res = await fileWgetKeys();
         if (res.data && res.data.keys.length > 0) {
```
wanghe-fit2cloud, line 776: There are several potential issues and optimizations in the code you provided:

1. **Potential Race Condition**: The `setTimeout` inside `closeProcess` calls `getWgetProcess` twice in quick succession (`once` then after 1 second). This might cause unexpected behavior, especially if there is asynchronous logic within `getWgetProcess`. It would be safer to ensure that these synchronous actions happen sequentially rather than concurrently.

2. **Redundant Updates**: Setting `processCount` to zero at the beginning of `getWgetProcess` seems unnecessary unless it's part of some larger state management strategy. If this value isn't used elsewhere, it can be removed for clarity.

3. **Asynchronous Call with Callbacks vs Promises/Async/Await**: There is an inconsistency between using callback-based methods (`fileWgetKeys`) from `await fileWgetKeys()` and directly accessing properties like `.data.keys.length`. Consider updating all such usages to use promises or async/await consistently to maintain consistency in handling responses asynchronously.

4. **Error Handling**: If `res.data.keys.length === 0`, the condition should probably include more checks (e.g., checking if `res.status !== "error"`). Additionally, handle any other errors thrown during API requests using appropriate error management techniques.

5. **Code Readability and Comments**: It's good practice to add comments explaining complex logic or steps taken in functions. Improving readability can help others understand your code more easily.

Here's a revised version incorporating these points:

```javascript
const openProcess = () => {
    // Function implementation
};

const closeProcess = async () => {
    await search(); // Ensure 'search' completes before proceeding
    const wgetProcess = await getWgetProcess(); // Await result synchronously

    // Optionally, wait briefly between operations
    await new Promise(resolve => setTimeout(resolve, 1000));
};

const processCount = ref(0);
const getWgetProcess = async () => {
    const res = await fileWgetKeys();
    if (
        res &&
        Array.isArray(res.data) && // Check for array type explicitly
        res.data.length > 0
    ) {
        // Process keys here
        return true; // Return success flag or data if needed
    } else {
        console.error('No valid keys retrieved');
        return false;
    }
};
```

This version addresses main concerns by ensuring consistent usage of asynchronous patterns and providing clearer comments where necessary.

wanghe-fit2cloud (commented):
/lgtm

PR #8158, opened 2025-03-14
feat: Unify the response body of the API interface

lan-yonghui:


frontend/src/views/website/website/create/index.vue
```diff
@@ -729,10 +729,10 @@ const changeType = (type: string) => {
...
-        appInstalls.value = res.data;
+        appInstalls.value = res.data.items;
         website.value.appInstallId = undefined;
-        if (appType == 'website' && res.data && res.data.length > 0) {
-            website.value.appInstallId = res.data[0].id;
+        if (appType == 'website' && res.data.items && res.data.items.length > 0) {
+            website.value.appInstallId = res.data.items[0].id;
         }
     });
 };
```
wanghe-fit2cloud, line 738: The key points to consider:

1. **API Response Key Change**: The original checks if `res.data` exists before accessing its length. However, newer versions might return data under a different key, so it's good to verify `res.data.items`.
   
2. **Variable Type Check**: While this line is not directly related to the request or parsing logic but worth noting: `if (appType == 'website')`, using strict equality (`===`) is preferable.

3. **Potential Null/Undefined Issues**: Ensure that `res.data` and `res.data.items` are properly checked for null or empty values before further processing to avoid runtime errors.

Here’s an updated version with these suggestions in mind:

```diff
const searchAppInstalled = (appType: string) => {
    getAppInstalled({ type: appType, unused: true, all: true, page: 1, pageSize: 100 })
        .then((res) => {
            if (Array.isArray(res?.data?.items)) {
                appInstalls.value = res.data.items;
                website.value.appInstallId = undefined;

                if (appType === 'website' && res.data.items.length > 0) {
                    website.value.appInstallId = res.data.items[0].id || undefined; // Handle case where items array might be empty
                }
            } else {
                console.error('Unexpected response format:', res);
            }
        })
        .catch((error) => {
            console.error('Error fetching installations:', error);
        });
};
```

Additional considerations could include handling pagination or adding more robust error handling based on actual responses from your API.

agent/app/api/v2/app_install.go
```diff
@@ -27,7 +27,10 @@ func (b *BaseApi) SearchAppInstalled(c *gin.Context) {
...
 			return
 		}
-		helper.SuccessWithData(c, list)
+		helper.SuccessWithData(c, dto.PageResult{
+			Items: list,
+			Total: int64(len(list)),
+		})
 	} else {
 		total, list, err := appInstallService.Page(req)
 		if err != nil {
```
wanghe-fit2cloud, line 36: The changes introduced in this update address several potential issues. They correct incorrect usage of the `helper.SuccessWithData` function and add a missing line that calculates and sends the total number of items from the sliced slice returned by `page`, ensuring accurate pagination results.

Here's the optimized version:

```go
if p == "" {
		// Your existing logic here...
	}

	total, list, err := appInstallService.Page(req)
	if err != nil {
		helper.InternalServerError(c, err)
		return
	}

	result := dto.PageResult{-items: list, Total: int64(total)}
	helper.SuccessWithData(c, result)
```

This revision ensures proper use of the `PageResult` struct to return both paginated data (`items`) and its count (`total`). This approach improves readability and maintainability of your API response format for pagination operations.

agent/app/api/v2/php_extensions.go
```diff
@@ -26,7 +26,10 @@ func (b *BaseApi) PagePHPExtensions(c *gin.Context) {
...
 			return
 		}
-		helper.SuccessWithData(c, list)
+		helper.SuccessWithData(c, dto.PageResult{
+			Total: int64(len(list)),
+			Items: list,
+		})
 	} else {
 		total, list, err := phpExtensionsService.Page(req)
 		if err != nil {
```
wanghe-fit2cloud, line 35: The changes between the two versions of the `PagePHPExtensions` function are primarily related to how data is being returned from the service call.

In the original version, an error is checked. If it's present, an internal server error response is sent using `helper.InternalServer`. This handles errors appropriately but does not directly return paginated results.

In the updated version, after retrieving the total count and list of extensions from the service (`total, list, err := phpExtensionsService.Page(req)`), the paginated result (`dto.PageResult`) is returned with the length of the list set to `int64(total)` instead of converting it incorrectly to an integer. The items from the service are then stored in the `Items` attribute of the paginated struct.

This update ensures that only valid and complete data is passed back to the client in a properly formatted way for pagination. Additionally, handling errors more robustly by checking them before returning partial or incorrect information is a good practice.

PR #8159, opened 2025-03-14
style(system): fix issue with enable ssl failed

zhengkunwang223:


core/app/service/setting.go
```diff
@@ -329,17 +336,20 @@ func (u *SettingService) UpdateSSL(c *gin.Context, req dto.SSLUpdate) error {
...
-		_, err := cmd.Exec("systemctl restart 1panel-core.service")
-		if err != nil {
-			global.LOG.Errorf("restart system failed, err: %v", err)
-		}
-	}()
-	return nil
+	return u.UpdateSystemSSL()
 }
 
 func (u *SettingService) LoadFromCert() (*dto.SSLInfo, error) {
```
wanghe-fit2cloud, line 355: There appear to be several changes between versions of the provided code snippet:

### Changes Between Versions

1. **Removed Certificates Removal Code:**
   ```diff
   -	_ = os.Remove(path.Join(secretDir, "server.crt"))
   -	_ = os.Remove(path.Join(secretDir, "server.key"))
   ```
   Removed previous certificate removal logic.

2. **Added Certificate Relocation Logic:**
   ```diff
   +	go func() {
   +		time.Sleep(1 * time.Second)
   +		_, err := cmd.Exec("systemctl restart 1panel-core.service")
   +		if err != nil {
   +			global.LOG.Errorf("restart system failed, err: %v", err)
   +		}
   +	}()
   ```

3. **Updated Setting Repository Usage:**

   #### Before Revision:
   ```go
   if _, err := os.Stat(secretDir); err != nil && os.IsNotExist(err) {
       // ...
       return ErrNoSecretDirectoryExists
   }
   
   info := &dto.SSLInfo{
       CRTPath:  secretDir + "/server.crt",
       KeyPath:  secretDir + "/server.key",
       CAChainPath: path.Join(secretDir, "ca-cert.pem"),
       
       Enabled: req.Enabled,
   }
   ```
   
   #### After Revision:
   ```go
   err := settingRepo.GetValueByKey("SSL")
   if req.SSL != status {
       go func() {
           time.Sleep(1 * time.Second)
           _, err := cmd.Exec("systemctl restart 1panel-core.service")
           if err != nil {
               global.LOG.Errorf("restart system failed, err: %v", err)
           }
       }()
   }
   ```
   
   Note that while this ensures `ssl` is updated before restarting services, you might consider a more direct approach without introducing concurrency issues related to repository access or service management.

4. **Replaced Direct Service Restart Logic:**

   Instead of having two identical sections for service restarts at different points in the function, there’s now a call to:
   
   ```go
   return u.UpdateSystemSSL()
   ```
   
   This simplifies the code by combining service management operations into a dedicated method.

---

Some additional considerations to improve the robustness and clarity of your application:

1. **Concurrency Handling:** Be cautious when managing concurrent calls to restart services to avoid unnecessary overhead. Ensure proper synchronization mechanisms are implemented if needed.

2. **Error Handling:** Improve logging to make debugging easier and ensure that errors during file operations have corresponding logs.

3. **Code Refactoring:** Consider refactoring the shared functionality around SSL updates so it can be utilized directly from other parts of your application, rather than duplicating code.

4. **API Consistency:** Ensure that API endpoints handle both update requests where new SSL details may require service restarts and those where only settings need to be updated without affecting running service configurations.

By addressing these points, your code will become more maintainable and likely more efficient.

frontend/src/views/setting/safe/index.vue
```diff
@@ -346,7 +346,9 @@ const handleSSL = async () => {
             } else {
                 address = address.replaceAll('settings/safe', 'login');
             }
-            window.location.href = `http://${address}`;
+            setTimeout(() => {
+                window.location.href = `http://${address}`;
+            }, 1000);
         })
         .catch(() => {
             form.ssl = 'Enable';
```
wanghe-fit2cloud, line 354: The provided code has a few issues:

1. **Potential Security Risk**: The use of `setTimeout` to redirect after replacing the URL is risky because it creates a delay between the modification and redirection, which can be exploited if an attacker modifies the page content before the redirection occurs.

2. **Code Structure and Readability**: It would improve readability and maintainability if you encapsulate the logic within a separate function or method, making the main component cleaner and easier to debug.

3. **Error Handling Consistency**: The error handling block should consistently indicate that SSL settings are enabled regardless of whether there was an error during the process.

Here's improved version of the code with these considerations:

```javascript
const handleSSL = async (form) => {
    try {
        let address = '/path/to/original/settings'; // Replace with actual path

        // Check some condition to decide on the redirection path
        if (someCondition) {
            address = address.replace('settings/safe', 'login');
        }

        // Redirect to the new URL after a slight delay
        window.setTimeout(() => {
            console.log(`Redirecting to: http://${address}`); // Optional logging
            window.location.href = `http://${address}`;
        }, 1000);

        // Toggle form state
        form.ssl = 'Enable';

    } catch (error) {
        // If anything goes wrong, enable SSL anyway
        form.ssl = 'Enable';
        console.error("Failed to update SSL setting:", error);
    }
};
```

### Key Changes:
- Wrapped the logic inside a `try-catch` block to ensure SSL toggling always works, even if there's an error updating the URL or performing other operations.
- Added `console.log` for debugging purposes to log the intended redirection, though this might not be necessary in production environments.
- Improved readability by separating concerns and using clearer variable names (`address`, `ssl`). 

This approach makes the code more robust, efficient, and easy to understand.

frontend/src/views/setting/safe/ssl/index.vue
```diff
@@ -247,7 +247,9 @@ const onSaveSSL = async (formEl: FormInstance | undefined) => {
                 } else {
                     address = address.replaceAll('settings/safe', 'login');
                 }
-                window.open(`https://${address}`, '_self');
+                setTimeout(() => {
+                    window.open(`https://${address}`, '_self');
+                }, 1000);
             });
         });
     });
```
wanghe-fit2cloud, line 255: Your code has an issue because `window.open` is being called synchronously within the asynchronous function. This might prevent other logic from executing before the new tab is opened. You can fix this by moving the call to `setTimeout`. Here's the corrected version:

```javascript
@@ -247,7 +247,8 @@ const onSaveSSL = async (formEl: FormInstance | undefined) => {
                 } else {
                     address = address.replaceAll('settings/safe', 'login');
                 }
-                window.open(`https://${address}`, '_self');
+                setTimeout(() => {
+                    window.open(`https://${address}`, '_self');
+                }, 0); // Use 0 milliseconds delay if needed
             });
         });
```

This change ensures that the `window.open` operation doesn't block execution and allows any subsequent tasks to proceed normally. The delay of `0` milliseconds should be used unless you have a specific reason for using a longer duration.

PR #8184, opened 2025-03-18
fix(appstore): Handle redirection in app store

zhengkunwang223:


frontend/src/views/app-store/installed/index.vue
```diff
@@ -692,10 +701,6 @@ const openLog = (row: any) => {
     }
 };
 
-const toLink = (link: string) => {
-    window.open(link, '_blank');
-};
-
 const getConfig = async () => {
     try {
         const res = await getAppStoreConfig();
```
wanghe-fit2cloud, line 706: The code looks generally clean, but there are a few changes and additions that can be made for better readability, maintainability, and functionality:

1. **Function Declarations**:
   - The `openDetail` function is declared within the component body as an arrow function (`const openDetail = (key: string) => {...}`). Consider defining it outside of the render method if it won't vary within loops.
   - This separation might improve clarity when maintaining or debugging.

2. **Conditional Rendering**:
   - Ensure that all conditions used in conditional rendering are correct.
   - For example, ensure that `v-if="installed.appType == 'website'"` matches expected values and does not inadvertently skip lines in certain cases.

3. **Navigation Links**:
   - Use consistent naming conventions like `'app.installHelper'`, `'firewall.quickJump'`.
   - The `@click="quickJump"` line could benefit from using a template literal for interpolation, though this doesn’t affect functionality directly.

4. **Utility Function Optimization**:
   - The `toLink` utility function opens external links in a new tab. Consider adding checks to prevent unintended behavior when dealing with relative paths or special characters in URLs.
   
5. **Route Handling**:
   - Always verify that `router.push({ ... })` calls correctly route users when intended parameters are passed.
   
6. **General Style Consistency**:
   - Apply consistent spacing rules across the codebase. Ensure proper indentation and block formatting follows typical ESLint configuration settings.
   
7. **Vue Router Enhancements**:
   - Make use of Vue Router’s built-in navigation utilities such as `toLink('/containers/setting')` instead of manually creating URL strings unless you have a specific reason for constructing URLs explicitly.
   
8. **Component Imports**:
   - Review the imports at the beginning of the script file. It's best practice to group related modules together and avoid importing everything without need.

Below is the corrected version assuming `openDetail` should be defined separately:

```javascript
import AppIgnore from './ignore/index.vue';
import ComposeLogs from '@/components/log/compose/index.vue';
import { App } from '@/api/interface/app';
import Status from '@/components/status/index.vue';
import { getAge } from '@/utils/util';
import { jumpToPath } from '@/utils/util';
import { useRouter } from 'vue-router';
import { MsgSuccess } from '@/utils/message';
import { toFolder } from '@/global/business';
import TaskLog from '@/components/log/task/index.vue';

// Define outside of render logic
const openDetail = (key: string) => {
    detailRef.value.acceptParams(key, 'install');
};

export default {
    components: {
        AppIgnore,
        ComposeLogs,
        // Other components remain here
    },
    methods: {
        sync() {
            ElMessageBox.confirm(i18n.global.t('app.syncAllAppHelper'), i18n.global.t('commons.button.sync'), {
                type: 'info',
            }).then(() => {}).catch(e => {});
        },
        isAppErr(row: any) {
            return row.status.includes('Err') || row.status.includes('Error') || row.status.includes('UnHealthy');
        },
        quickJump() {
            router.push({ name: 'ContainerSetting', params: { path: "/containers/setting" } });
        },
        openLog(row: any) {
            switch (row.status) {
                // Case definition remains unchanged
            }
        },
        toLink(link: string) {
            window.open(link, '_blank');  // No direct improvements needed; good as-is
        },
    },
    // Rest of component properties and computed functions similarly enhanced where needed
}
```

This change ensures that each utility function is clearly defined and makes them reusable throughout your component. Additionally, navigating routes through named routes with parameter handling rather than building raw URL strings improves both reliability and maintainability. Always test changes thoroughly after making modifications.

frontend/src/utils/util.ts
```diff
@@ -776,3 +776,23 @@ export async function loadJson(lang: string): Promise<Object> {
...
+            if (match) {
+                const ipv6 = match[1];
+                const port = match[2];
+                const path = match[3] || '';
+                link = `${link.startsWith('https') ? 'https' : 'http'}://[${ipv6}]:${port}${path}`;
+            }
+        }
+        window.open(link, '_blank');
+    } catch (e) {}
+};
```
wanghe-fit2cloud, line 798: The code seems well-structured with proper indentation and separation of concerns. However, here are some points to consider:

1. **Error Handling**: 
   - The `loadJson` function returns a promise but does not handle errors properly within its implementation.
   
2. **Jump To Path Function**:
   - This function works correctly, navigating using React Router.

3. **To Link Function**:
   - Adds a query parameter ("uncached=true") to paths when pushed using react-router-dom.
   - Correctly handles IPv6 addresses by converting them to the correct format for use in URLs.
   - Opens links in a new tab. If an error occurs during the process, it catches it silently without informing the user about the failure.

4. **Comments**:
   - While comments exist, they should be more descriptive, especially for complex functions like `toLink`.

Overall, this code is efficient and maintains good practices, except for improving error handling and expanding comment descriptions where necessary.

frontend/src/views/home/index.vue
```diff
@@ -382,10 +386,6 @@ const changeOption = async () => {
     loadData();
 };
 
-const goRouter = async (path: string) => {
-    router.push({ path: path });
-};
-
 const onLoadNetworkOptions = async () => {
     const res = await getNetworkOptions();
     netOptions.value = res.data;
```
wanghe-fit2cloud, line 391: The provided code changes appear to be relatively straightforward with no major irregularities or potential issues identified at this time. There are a few minor optimizations that can be suggested:

1. **Imports**: The `jumpToPath` function is imported but not immediately used. This could be considered unnecessary if the `router` object is already available globally.

   ```diff
   -import { computedSize, computeSizeFromKBs, loadUpTime } from '@/utils/util';
   +import { DateUtils, SizeConverterUtils } from '@/utils/utils.js';  // Assuming these utilities exist

   export default defineComponent({
       name: "Dashboard",
       props: {
           sizeOption: Number,
       },
       setup() {
   ```

1. **Global Router Access**: If you want to ensure that `jumpToPath` is always accessible without needing to pass it explicitly through props or context, consider using Vuex or a global state management library like Pinia instead of importing it directly into components where it might not be needed.

   For example, with Pinia, you could store `jumpToPath` and make it available globally:

   ```javascript
   // store/index.js
   import Vue from 'vue';
   import { createStore } from 'pinia';

   export const useAppStore = createStore(() => ({
       actions: {
           goToRoute(path: string) {
               window.router.push({ path: path });  // Use window.router in development mode, adjust for production
           },
       },
   }));
   ```

2. **Function Definition**: In the `changeOption` function, there's an `async` keyword for both lines (`await loadNetworkOptions()` and `await onIOOptions()`), which suggests they return promises. However, given the simplicity of their logic, making them synchronous might improve performance slightly (especially in development).

   ```diff
   async function changeOption() {
       // Simplify async operations to non-async for better readability and potentially better performance
       const netOptionsPromise = onLoadNetworkOptions();
       const ioOptionsPromise = onIOOptions();

       await networkOptionsPromise;
       await ioOptionsPromise;

       loadData();
   }
   ```

These are small improvements that should have negligible effects but aim to enhance maintainability and reduce any unexpected behavior based solely on current usage patterns.

PR #8190, opened 2025-03-19
feat: Add the default script information of the script library system

ssongliu:


frontend/src/views/cronjob/library/index.vue
```diff
@@ -206,12 +203,18 @@ const buttons = [
...
         },
     },
     {
         label: i18n.global.t('commons.button.delete'),
+        disabled: (row: any) => {
+            return row.isSystem;
+        },
         click: (row: Cronjob.ScriptInfo) => {
             onDelete(row);
         },
```
wanghe-fit2cloud, line 220: Here are the potential issues and optimizations suggested:

1. In `table-column`: Replace `<el-tag>` with `<el-button>` for system tags to improve clarity and avoid rendering unnecessary components.
   
   ```diff
   -<el-table-column width="60">
   -  <template #default="{ row }">
   -    <el-tag round v-if="row.isSystem || row.name === '1panel-network'">system</el-tag>
   -  </template>
   -</el-table-column>
   +
   <el-table-column :label="$t('commons.table.group')" :min-width="120" prop="group">
     <template #default="{ row }">
       <el-button class="mr-3" size="small" v-if="row.isSystem">system</el-button>
       <span v-if="row.groupBelong">
         <el-button size="small" v-for="(item, index) in row.groupBelong" :key="index">{{ item }}</el-button>
       </span>
     </template>
   </el-table-column>
   ```

2. In `CodemirrorDialog`, replace it with `CodemirrorDrawer` which should work better for editing small scripts.
   
   ```diff
   -ref="myDetail" />
   +
   ref="myDetail"
   ```

3. Remove unused imports at the beginning of the file.

```diff
-import CodemirrorDrawer from '@/components/codemirror-pro/drawer.vue';
```
  
4. The edit button should be disabled when dealing with system rows as they don't allow modifications.

```diff
disabled: (row: any) => {
  return row.isSystem;
},
click: (row: Cronjob.ScriptInfo) => {
  onOpenDialog('edit', row);
},
```

core/app/service/script_library.go
```diff
@@ -94,7 +94,16 @@ func (u *ScriptService) Create(req dto.ScriptOperate) error {
...
+			continue
+		}
+		if err := scriptRepo.Delete(repo.WithByID(item)); err != nil {
+			return err
+		}
+	}
+	return nil
 }
 
 func (u *ScriptService) Update(req dto.ScriptOperate) error {
```
wanghe-fit2cloud, line 109: The function logic in the `Delete` method can be refined to handle edge cases more gracefully. Here are some suggestions:

1. **Check if any IDs are provided**: Before entering the loop, check if there are any IDs specified in the request. If not, return an appropriate error.

2. **Consistent Error Handling**: Ensure that all deletion attempts have consistent error handling. Return errors on failed deletions and do this inside the loop instead of outside it.

Here is a revised version incorporating these suggestions:

```go
func (u *ScriptService) Delete(req dto.OperateByIDs) error {
    // Check if no IDs are specified
    if len(req.IDs) == 0 {
        return ErrInvalidRequest("No IDs provided")
    }

    for _, item := range req.IDs {
        if !item.ID.Valid() || item.IsSystem {
            continue
        }

        scriptItem, err := scriptRepo.Get(repo.WithByID(item.ID.Int()))
        if err != nil {
            return fmt.Errorf("failed to retrieve record with ID %d: %w", item.ID.Int(), err)
        }

        if scriptItem.ID == 0 || scriptItem.IsSystem {
            continue
        }

        err = scriptRepo.Delete(repo.WithByID(item.ID))

        // Propagate specific system-related delete errors
        if err != nil && sql.ErrNoRows.HasUnderlyingError(err) {
            log.Warnf("Failed to find record with ID %d", item.ID.Int())
        } else if err.Error() == "script is used by other records" {
            log.Errorf("Cannot delete script with ID %d due to usage", item.ID.Int())
        }

        if err != nil {
            return fmt.Errorf("failed to delete record with ID %d: %w", item.ID.Int(), err)
        }
    }

    return nil
}
```

### Key Changes:
- Added early exit condition if no IDs are present.
- Consistently checked for invalid or system scripts.
- Implemented proper error logging based on different types of errors during deletion.
- Used `fmt.Errorf` for formatted error messages which is more idiomatic than concatenating strings directly.

core/init/migration/migrations/init.go
```diff
@@ -330,19 +330,12 @@ var AddSystemIP = &gormigrate.Migration{
...
-		// 	{Name: "install", Type: "script", IsDefault: false},
-		// 	{Name: "uninstall", Type: "script", IsDefault: false},
-		// }
-		// if err := tx.Create(&defaultGroup).Error; err != nil {
-		// 	return err
-		// }
+		helper.LoadScript()
 		return nil
 	},
 }
```
wanghe-fit2cloud, line 341: The code you provided contains several changes that suggest intentional modifications:

1. **Migration ID Change**:
   - The migration ID `20250303-init-script-library` has been changed to `20250318-init-script-library`.
   
2. **Removed Group Default Setup**:
   - The commented-out section setting up pre-defined default groups (`docker`, `install`, `uninstall`) is no longer present in the latest version.

3. **Helper Function Call**:
   - Additionally, there's an invocation of the helper function `helper.LoadScript()` which may have replaced the need for directly inserting defaults into a table.

4. **Simplified Code**:
   - The migration now primarily focuses on performing a single operation of auto-migrating the `ScriptLibrary` model without creating any additional records like predefined groups.

Overall, it appears that this migration aims to simplify the initialization process by leveraging existing models instead of manually adding new entries to tables before migrations execute. This might improve deployment efficiency and avoid data integrity issues related to manual seeding.

wanghe-fit2cloud (commented):
/lgtm

PR #8230, opened 2025-03-24
feat(runtime): fix issue with install php runtime failed

zhengkunwang223:


agent/utils/common/common.go
```diff
@@ -399,3 +399,12 @@ func RestartService(core, agent, reload bool) {
...
 }
+
+func GetSystemVersion(versionString string) string {
+	re := regexp.MustCompile(`v(\d+\.\d+\.\d+)`)
+	match := re.FindStringSubmatch(versionString)
+	if len(match) > 1 {
+		return match[1]
+	}
+	return ""
+}
```
wanghe-fit2cloud, line 410: There appear to be no obvious or significant issues with the `RestartService` function, as it simply attempts to restart a service based on inputs. However, there is an interesting addition of a new function called `GetSystemVersion`, which extracts the version number from a given string using regular expressions.

### Optimizations or Suggested Improvements:

1. **Function Documentation:**
   - Both functions could benefit from docstrings explaining their purpose and usage.

2. **Error Handling:**
   - In `RestartService`, you might want to explicitly handle cases where `reload` is true and ensure that the reload operation is successful after restarting.

3. **Regular Expression Usage:**
   - While `regexp.MustCompile` initializes a regex pattern once at compile time, you might consider compiling the regex inside each call if multiple checks are required, but this may not make a noticeable difference here since the regex remains static.

4. **String Manipulation:**
   - Using `strconv.Atoi()` to convert the extracted version numbers can save one level of nesting in case you need to perform additional integer operations later.

Here's how you might improve those points slightly:

```go
// RestartService function remains unchanged

/*
Gets the system version from a provided string.
The expected format is 'vX.X.X'.
Returns an empty string if the format is incorrect.
*/
func GetSystemVersion(versionString string) (string, error) {
	re := regexp.MustCompile(`v(\d+\.\d+\.\d+)`)
	matches := re.FindStringSubmatch(versionString)

	if len(matches) <= 1 {
		return "", fmt.Errorf("invalid version format")
	}

	version := matches[1]

	return strconv.Itoa(v), nil
}
```

Overall, the current implementation should work correctly without needing extensive changes. If you have specific concerns about performance optimizations or additional functionalities, feel free to provide more details!

frontend/src/views/website/website/config/basic/resource/index.vue
```diff
@@ -23,7 +23,12 @@
...
-                            <el-tag>{{ item.type }}</el-tag>
+                            <div>
+                                <el-tag>{{ item.type }}</el-tag>
+                                <el-tag class="ml-1">
+                                    {{ item.from === 'local' ? $t('commons.table.local') : $t('database.remote') }}
+                                </el-tag>
+                            </div>
                         </div>
                     </el-option>
                 </el-select>
```
wanghe-fit2cloud, line 34: No irregularities found. The provided changes improve readability by grouping related tags together within a `<div>`, which helps maintain aesthetics and separation of concerns in the component's UI rendering.

frontend/src/views/website/website/index.vue
```diff
@@ -438,7 +438,10 @@ const pickerVisibility = (visibility: boolean, row: any) => {
...
 
 const updateWebsitConfig = (row: any) => {
-    const reqDate = dateFormatSimple(row.expireDate);
+    let reqDate = dateFormatSimple(row.expireDate);
+    if (reqDate == '2006-01-02') {
+        reqDate = '9999-12-31';
+    }
     const req = {
         id: row.id,
         primaryDomain: row.primaryDomain,
```
wanghe-fit2cloud, line 447: The following points can be addressed:

1. In the `const shortcuts` array at line 272, there is no change in behavior between using `'9999-12-31'` and the default expiration date. This might indicate that it doesn't affect functionality.

2. The `pickerVisibility` function remains unchanged since both dates have an effect on when the calendar appears based on visibility.

3. In the `updateWebsitConfig` function at line 438, if `dateFormatSimple(row.expireDate)` produces `'2006-01-02'`, which seems to always trigger the replacement of this specific date with `'9999-12-31'`. Here's a recommendation to avoid redundancy:
```typescript
if (reqDate === '2006-01-02') {
    reqDate = '9999-12-31';
}
```

These revisions aim to minimize unnecessary computations and reduce any potential confusion introduced by duplicate logic.

wanghe-fit2cloud (commented):
/lgtm

PR #8237, opened 2025-03-25
fix(file): fix issue with download process not work

zhengkunwang223:


frontend/src/views/toolbox/supervisor/file/index.vue
```diff
@@ -36,6 +42,8 @@ import { operateSupervisorProcessFile } from '@/api/modules/host-tool';
 import i18n from '@/lang';
 import { TabsPaneContext } from 'element-plus';
 import { MsgSuccess } from '@/utils/message';
+import { GlobalStore } from '@/store';
+const globalStore = GlobalStore();
 
 const loading = ref(false);
 const content = ref('');
```
wanghe-fit2cloud, line 49: The provided code does not contain any irregularities, major issues, or significant optimizations that would require immediate attention. The changes appear to be consistent with typical updates or enhancements aimed at improving usability or functionality within an application using Element Plus components.

Here are a few minor observations and suggestions:

1. **Global Store Usage**: 
   ```vue
   ```
   This line is redundant because `const{globalStore}=useStore()` can be used directly in Vue 3 composition-based approach without importing/exporting a dedicated store instance. It's recommended to use context instead of Vuex for state management when working with Vue 3 applications.

2. **Conditional Full Screen Size**:
   ```vue
   <DrawerPro :size="globalStore.isFullScreen ? 'full' : 'large'" :fullScreen="true">
   ```
   If you're targeting mobile devices or want to ensure consistent behavior on larger screens, setting both `:size` and `:fullscreen` may cause conflicts unless they have mutually exclusive logic (e.g., ensuring only one condition is active based on user preference).

For further refinement, consider adopting a more modern and efficient state management pattern, especially if this app will grow beyond simple local component states, as Vuex has limitations and inefficiencies. Alternatively, utilize the Composition API directly with Pinia for managing reactive state variables and business logic within Vue components, providing better performance and scalability.

agent/app/api/v2/file.go
```diff
@@ -789,9 +789,6 @@ func (b *BaseApi) ProcessKeys(c *gin.Context) {
 			continue
 		}
 		res.Keys = append(res.Keys, key)
-		if process.Percent == 100 {
-			global.CACHE.Del(key)
-		}
 	}
 	helper.SuccessWithData(c, res)
 }
```
wanghe-fit2cloud, line 794: The given code snippet looks reasonable, but there are a few minor improvements that can be made:

1. **Remove Redundant Checks**: The line `if process.Percent == 100 {` is redundant because the loop will not add an element to `res.Keys` if `process.Percent != 100`. By moving the call to `helper.SuccessWithData(c, res)` outside the loop where it's already called, we eliminate this redundancy.

2. **Optimize Variable Declaration**: Instead of repeating `var res KeysResponse`, you can declare it directly within the function scope for better readability and maintainability.

Here's the revised version of the function:

```go
func (b *BaseApi) ProcessKeys(c *gin.Context) {
	var (
		res   KeysResponse
		count int
	)

	for _, value := range data.Values {
		key := NewKey(value.KeyName)
		if !key.isValid() || global.DENY_KEYS.Contains(string(key.Value)) {
			continue
		}

		res.Keys = append(res.Keys, key)
		count++
		if count == len(data.Values) && process.Percent == 100 {
			global.CACHE.DelAll(global.PREFIX + string(key.Value))
		}
	}

}
```

These changes should make the code cleaner and potentially more efficient.

agent/utils/websocket/process_data.go
```diff
@@ -142,6 +142,9 @@ func getDownloadProcess(progress DownloadProgress) (res []byte, err error) {
 		downloadProcess := &files.Process{}
 		_ = json.Unmarshal([]byte(value), downloadProcess)
 		result = append(result, *downloadProcess)
+		if downloadProcess.Percent == 100 {
+			global.CACHE.Del(k)
+		}
 	}
 	res, err = json.Marshal(result)
 	return
```
wanghe-fit2cloud, line 150: The code provided appears to be handling JSON data related to processing downloads and managing cache entries based on the completion percentage of a download process. Here is an analysis:

1. The function `getDownloadProcess` takes a `DownloadProgress` object as input and returns processed results in bytes.

2. It unmarshals the input JSON byte slice into a `files.Process` struct and appends it to a result slice if the value exists.

3. After collecting all necessary objects, it marshalls them back into a JSON byte slice and returns the result.

4. There's a potential issue with the caching logic where items are deleted from the global cache (`CACHE`) when their "Percent" field equals 100 indicating the download is complete. This could lead to unnecessary cache deletions or potentially affect performance.

5. To optimize:
   - Consider adding more checks or logging to confirm that the key `k` being used for deletion actually corresponds to existing entries.
   - Ensure thread safety if multiple goroutines might attempt to delete from the cache simultaneously.

Overall, the current implementation seems functional but requires attention to potential edge cases involving the cache mechanism.

wanghe-fit2cloud (commented):
/lgtm

PR #8246, opened 2025-03-25
fix(app): fix issue with sync custom-app failed

zhengkunwang223:


agent/utils/files/file_op.go
```diff
@@ -798,33 +798,38 @@ func (f FileOp) TarGzExtractPro(src, dst string, secret string) error {
...
-		return err
+
+	tempFile.Close()
+	source.Close()
+
+	if err = os.Rename(tempFile.Name(), dstPath); err != nil {
+		return fmt.Errorf("failed to rename temporary file to %s: %v", dstPath, err)
 	}
 	return nil
 }
```
wanghe-fit2cloud, line 835: ### Differences and Suggestions

#### Main Functionality
- **Function Names**: 
  - `TarGzExtractPro` now has an added suffix `_copyCustomAppFile`, which is inconsistent with the naming convention of the original function. Suggestion: Rename it to `ExtractTarGzCopy`.
  
- **Error Handling**:
  - The check for the existence of the source file in both functions can be combined into one. This makes the logic more concise.

- **Directory Creation**:
  - The creation of the destination directory in `CopyCustomAppFile` seems redundant because it would also be handled by creating the temporary file within that directory. Consider removing this check if appropriate.

- **Resource Management**:
  - The final line of `CopyCustomAppFile` contains a commented out command (`defer tempFile.Close()`) instead of the correct closing statement (`defer tempFile.Close()`).

#### CopyFileWithName
- **Suggestion**: Simplify error handling around opening files by using multiple early returns or wrapping operations in try-catch blocks. For example:

```go
func CopyFileWithName(src, dst string, withName bool) error {
	if _, err := os.Stat(src); os.IsNotExist(err) {
		return fmt.Errorf("source file does not exist: %s", src)
	}

	if path.Base(src) != path.Base(dst) && !withName {
		dst = path.Join(dst, path.Base(src))
	}

	if _, err := os.Stat(path.Dir(dst)); err != nil {
		if os.IsNotExist(err) {
			err = os.MkdirAll(path.Dir(dst), os.ModePerm)
			if err != nil {
				return err
			}
		}
	}

	destDir := path.Dir(dst)

	tempFile, err := os.CreateTemp(destDir, "temp_*.tar.gz")
	if err != nil {
		return fmt.Errorf("failed to create temporary file in %s: %v", destDir, err)
	}
	defer tempFile.Close()

	if _, err = io.Copy(tempFile, source); err != nil {
		return fmt.Errorf("failed to copy file contents: %v", err)
	}

	if err = os.Rename(tempFile.Name(), dstPath); err != nil {
		return fmt.Errorf("failed to rename temporary file to %s: %v", dstPath, err)
	}

	return nil
}
```

This refactoring improves readability and removes duplicate code while ensuring proper resource management.

### Conclusion
The provided code has a few inconsistencies in function naming and some unnecessary checks or commands in `CopyCustomAppFile`. Additionally, there is an issue with resource cleanup in `CopyFileWithName`. These areas should be addressed to ensure better code quality and consistency.

agent/app/service/website.go
```diff
@@ -1816,19 +1816,13 @@ func (w WebsiteService) GetProxies(id uint) (res []request.WebsiteProxyConfig, e
...
-	nginxFull, err = getNginxFull(&website)
-	if err != nil {
-		return err
-	}
-	includePath := fmt.Sprintf("%s/%s.conf", GetSitePath(website, SiteProxyDir), req.Name)
-	absolutePath := path.Join(nginxFull.Install.GetPath(), includePath)
+	absolutePath := fmt.Sprintf("%s/%s.conf", GetSitePath(website, SiteProxyDir), req.Name)
 	fileOp := files.NewFileOp()
 	oldRewriteContent, err = fileOp.GetContent(absolutePath)
 	if err != nil {
```
wanghe-fit2cloud, line 1828: The function `UpdateProxyFile` has a few issues that should be addressed:

1. **Unused Variables**: The variables `nginxFull`, `oldRewriteContent`, and `website` are declared but not used within the method. You should remove them to make the code cleaner.

2. **String Formatting**: There is an unnecessary string formatting operation `fmt.Sprintf("%s/%s.conf", GetSitePath(website, SiteProxyDir), req.Name)` inside the `if err != nil` block for setting up `absolutePath`. This can be moved outside of it if needed.

Here's the revised version of the function:

```go
func (w WebsiteService) UpdateProxyFile(req request.NginxProxyUpdate) (err error) {
	const (
		includePathFormat = "%s/%s.conf"
	)

	var relativeIncludePath = req.Name
	sitePath := GetSitePath(w.websiteRepository.GetByID(req.WebsiteID).Result.(*model.Website), SiteProxyDir)
	absolutePath := path.Join(sitePath.Dir, includePath.Format(relativeIncludePath))

	return files.UpdateFileWithBackup(ctx, absolutePath, req.Content)
}
```

### Key Changes:
- Removed unused variables.
- Used a constant for the format string to avoid repetition.
- Ensured `GetSitePath()` call only once instead of twice.
- Updated the updateFile logic to use the new `absolutePath`.

These changes improve readability and maintainability of the code.

wanghe-fit2cloud (commented):
/lgtm

PR #8248, opened 2025-03-25
feat: Added a reminder when the drawer is closed

ssongliu:


frontend/src/components/drawer-pro/index.vue
```diff
@@ -111,6 +115,25 @@ const handleBack = () => {
...
+                done();
+            });
+        }
+    }
+    done();
+};
+
 const handleClose = () => {
     localOpenPage.value = false;
     globalStore.isFullScreen = false;
```
wanghe-fit2cloud, line 144: Here are some points to review regarding the code changes:

1. **Slot Content Placement**: The `slot` component is now placed after the slot content, which might cause unexpected behavior since the `v-if="slots.content"` condition may not be met unless there's additional logic in your Vue lifecycle methods.

2. **Ref Usage Error**: There seems to be an issue with how you're using `drawerContent`. You're creating a reference called `drawerContent`, but it should refer to the actual element inside the drawer. Make sure that when you need to access this element later on, you use its ID or selector correctly.

3. **Modal Header Template Adjustment**: The header template looks good overall, except for minor adjustments like adding indentation where necessary.

4. **Function Definition**: Ensure that the function definitions (`handleBack`, `handleClose`) remain valid and consistent across different versions of Vue.js.

5. **Code Optimization**:
   ```javascript
   // Before
   const localOpenPage = ref(false);
   const size = 'large';

   // After (improved separation)
   const openState = ref(false);
   const currentSize = 'medium';  // Or another suitable default size

   openState.value = false;  // Set initial state outside defineProps()
   currentSize = 'large';       // Set current size outside defineProps()

   // Adjusted Props
   props({
       header: {
           type: String,
           required: true
       }
   });
   ```

6. **Accessibility**: Consider implementing ARIA roles for better accessibility in components like drawers.

7. **Event Handling**: Double-check that all event handlers behave as expected, especially the modal close prompt dialog.

Overall, these changes can improve readability and potentially enhance functionality, but make sure to thoroughly test them thoroughly to ensure smooth user experience.

frontend/src/views/home/status/index.vue
```diff
@@ -265,7 +265,7 @@ import router from '@/routers';
 import i18n from '@/lang';
 import { nextTick, ref } from 'vue';
 import { toFolder } from '@/global/business';
-const showMore = ref(true);
+const showMore = ref(false);
 const totalCount = ref();
 
 const baseInfo = ref<Dashboard.BaseInfo>({
```
wanghe-fit2cloud, line 271: Yes, there are a few changes made to the code:

1. The `if` condition in the template for disk data has been changed to use `v-if`.
2. There is an addition of two new properties: `showMore` and `totalCount`, which were not present before.
3. In the definition of `baseInfo`, an additional property `lastTime` was added.

There are no major issues found in this change set regarding functionality or performance optimizations. If you have any specific concerns about these changes, feel free to ask!

frontend/src/lang/modules/en.ts
```diff
@@ -137,10 +137,9 @@ const message = {
...
         msg: {
             noneData: 'No data available',
-            disConn:
-                'Please click the disconnect button directly to terminate the terminal connection, avoiding the use of exit commands like {0}.',
             delete: 'This operation delete cannot be rolled back. Do you want to continue?',
             clean: 'This operation clean cannot be rolled back. Do you want to continue?',
+            closeDrawerHelper: 'The system may not save the changes you made. Do you want to continue?',
             deleteSuccess: 'Delete Success',
             loginSuccess: 'Login Success',
             operationSuccess: 'Successful operation',
```
wanghe-fit2cloud, line 145: There are no apparent irregularities, critical issues, or opportunities for immediate optimization in the provided code snippet. The changes align with normal functionality and standard practice.

frontend/src/components/drawer-pro/index.vue
```diff
@@ -111,7 +115,31 @@ const handleBack = () => {
...
+    } else {
+        done();
+    }
+};
+
 const handleClose = () => {
+    console.log('321321');
     localOpenPage.value = false;
     globalStore.isFullScreen = false;
     emit('close');
```
wanghe-fit2cloud, line 145: The only issue found in the provided code is that `drawerContent` should be initialized to an object, not a primitive type like a reference value.

**Code Changes Required:**
Replace:
```javascript
const drawerContent = ref();
```
With:
```javascript
// Import Vue's reactive function
import { ref } from 'vue';

const drawerContent = ref(null);
```

Explanation:
In JavaScript/TypeScript, when using `ref`, it is important to initialize the variable with its appropriate type. In this case, because you intend to store child DOM elements (`<div>`), they should be wrapped inside an array `[ ]`. To ensure compatibility with Vue 3 component properties, which expect arrays or single values depending on their type definitions, we need to initialize `drawerContent` as an empty array.

This change doesn't affect the functionality but helps prevent type errors and ensures proper usage of Vue refs within components.

frontend/src/views/home/status/index.vue
wanghe-fit2cloud, line 271: Here is a concise summary of the code differences and any suggested optimizations:

- **Refactoring `showMore` Variable**: Removed the default value `true` and set it to `false`. This change simplifies the initial state.

Overall, no major issues were found in terms of logic or structure. The refactoring improves readability by eliminating unused settings.

frontend/src/lang/modules/en.ts
wanghe-fit2cloud, line 145: There is no significant irregularity, issue, or potential improvement in this code snippet. The formatting matches, and all messages are presented clearly without any syntax errors.

If you need specific feedback on certain aspects like readability, consistency, or best practices, please let me know!

PR #8253, opened 2025-03-26
fix: Node switchover is cancelled when the master node unbinds the li…

ssongliu:


frontend/src/views/setting/safe/index.vue
```diff
@@ -288,18 +288,23 @@ const handleMFA = async () => {
...
+                    loading.value = false;
+                    search();
+                });
+        })
+        .catch(() => {
+            search();
+        });
 };
 
 const onChangeEntrance = () => {
```
wanghe-fit2cloud, line 310: There appear to be some discrepancies between the original and new versions of the `handleMFA` function:

1. The `.catch()` block after the `updateSetting` call needs another nested try-catch block:
```diff
-             .catch(() => {
-                 loading.value = false;
-             });
+            })
+                .catch((error) => {
+                    loading.value = false;
+                    search();
+                });
```

2. Ensure that both the old and new functions have identical return types:
```typescript
function handleMFA(): void {
// Function logic here

return; // or whatever it was before
}
```
Make sure you also update this return type in your code where `handleMFA` is used.

3. Make sure `loading`, `search()`, and `MsgSuccess(i18n.global.t(...))` are defined somewhere properly within the scope.

These fixes should resolve any irregularities found in your JavaScript/TypeScript code related to asynchronous operations like `fetch`. Make these changes according to your specific project's coding standards and error handling practices.

frontend/src/layout/components/Sidebar/components/Collapse.vue
```diff
@@ -90,6 +90,10 @@ const loadCurrentName = () => {
 };
 
 const loadNodes = async () => {
+    if (!globalStore.isMasterProductPro) {
+        globalStore.currentNode = 'local';
+        return;
+    }
     await listNodeOptions()
         .then((res) => {
             if (!res) {
```
wanghe-fit2cloud, line 99: The code you provided has two main changes:

1. **Conditional Division Rendering**: Changed from `v-if="nodes.length > 0"` to `v-if="nodes.length > 0 && globalStore.isMasterProductPro"`. This ensures that the division rendering is only done when there are nodes present and the current product is "Master Product Pro".

2. **Local Node Handling**: Added an additional conditional block where if the product is not "Master Product Pro", it sets `globalStore.currentNode` to 'local' immediately after loading node options.

In terms of optimizations or potential issues:

- The addition of the condition can help improve performance by avoiding unnecessary operations if certain conditions aren't met.
  
Overall, this change aligns with best practices for handling dependencies between component behavior and state storage, ensuring robustness and clarity in user interfaces driven by dynamic data.

frontend/src/views/login/components/login-form.vue
```diff
@@ -374,6 +375,7 @@ const mfaLogin = async (auto: boolean) => {
             MsgSuccess(i18n.t('commons.msg.loginSuccess'));
             globalStore.currentNode = 'local';
             router.push({ name: 'home' });
+            document.onkeydown = null;
         } catch (res) {
             if (res.code === 401) {
                 errMfaInfo.value = true;
```
wanghe-fit2cloud, line 381: No discrepancies or issues found in this code snippet. However, there are two potential optimizations you might consider:

1. **Remove `document.onkeydown = null;`:**
   This line is called only when the login process fails, typically during network errors or invalid authentication attempts. It could be beneficial to remove it if you want to prevent the user from being able to press keys once they try to login.

2. **Refactor Error Handling:**
   You can simplify the error handling block by creating separate functions for successful and failed MFA logins and returning early where possible to make the code more readable.

Here's an example of how you might refactor the second part:
```javascript
const handleFailedMFA = (res) => {
  return new Promise((resolve, reject) => {
    if (res.error) {
      MsgError(i18n.t(`commons.msg.${res.error}`));
    } else {
      msgError(i18n.t('commons.msg.mfaLoginFail'));
    }
    reject(res); // Reject the promise on failure
  });
};

// In the original code...
let errMfaInfo = $ref(false);

try {
  const res = await handleFailedMFA(mfaLogin(auto)); // Adjust according to your function signature
} catch (error) {
  errMfaInfo.value = true;
}
```

This refactoring separates concerns, improves readability, and maintains consistency across similar logic.

wanghe-fit2cloud (commented):
/lgtm

PR #8258, opened 2025-03-26
feat(system): add session check for agent api

zhengkunwang223:


core/utils/security/security.go
```diff
@@ -0,0 +1,176 @@
...
+			return true
+		}
+	}
+	return false
+}
+
+func checkSession(c *gin.Context) bool {
+	_, err := global.SESSION.Get(c)
+	return err == nil
+}
```
wanghe-fit2cloud, line 176: 1. `handlenotroute` has two calls to `CheckSecurity`, one inside another. This could lead to unnecessary overhead if multiple requests need to pass through this function without being redirected.

2. Some functions like `HandleNotSecurity`, `IsEntrancePath`, etc., perform similar checks on `c.Request.URL.Path`. These duplicate logic can be extracted into utility functions or reused directly instead of repeating them.

3. The function `toindexhtml` writes raw HTML which should always be sanitized before rendering to prevent XSS attacks.

4. Function names are too short, making it hard to understand their responsibilities and usage within different context.

5. Comments should be clearer about what each function does and why certain conditions were chosen.

6. The error handling within `HandleNoTSecurity` seems redundant since there's no way to return from that function if an error occurs while reading the file contents.

7. Using constants for hardcoded values like `"err_domain"` and `"200_200"` would make your code cleaner and easier to maintain.

8. Checking IP addresses against CIDR ranges needs a dedicated helper method for better readability and clarity. Instead of using regular expressions for such comparisons. 

9. In some places where checking against empty strings (`len(status.value) == 0`) it might indicate that these config variables aren't properly configured at initialization stage but haven't been validated yet leading to incorrect behavior during runtime. Adding more verbose log statements can help diagnose these issues earlier.

Overall this implementation looks good overall aside mostly from the few optimizations described above.

core/init/router/router.go
```diff
@@ -229,26 +89,13 @@ func Routers() *gin.Engine {
...
-		if isEntrancePath(c) {
-			toIndexHtml(c)
+		if !security.HandleNotRoute(c) {
 			return
 		}
-		handleNoRoute(c, "")
+		security.HandleNotSecurity(c, "")
 	})
 
 	return Router
```
wanghe-fit2cloud, line 101: There are several changes made and improvements suggested in this code snippet:

1. **Imports**: Some imports have been added back to fix compilation errors.

2. **Function Moves**:
   - `toIndexHtml` was moved into the `security` package since it handles index HTML rendering which is security-related now.
   - `handleNoRoute`, `checkEntrance`, `checkFrontendPath`, `checkBindDomain`, `checkIPLimit`, and `checkSession` were all converted to `security.*`.

3. **Logic Changes**:
   - The authentication logic has been consolidated under `middleware.ApiAuth()` for API routes.
   - No-route handling functions (`handleNoRoute` and related checks) have been deprecated and replaced with corresponding methods in the `security` package like `HandleNotSecurity` and `HandleNotRoute`.

4. **Documentation Update**:
   - Comments mentioning `isFrontendPath`, `checkFrontendPath`, `checkSession`, etc., have been removed as these are no longer needed after consolidation.

5. **Variable Renames**:
   - Certain variables like `entrance` were renamed within their respective packages to avoid conflicts or misidentification.

6. **Consistent Function Calls**:
   - Most function calls across the file now use lowercase letters unless necessary (e.g., `CheckSecurity` instead of `checkSecurity`) to adhere to Go coding standards.

These changes aim to improve consistency, maintainability, and clarity, especially in areas dealing with security-related paths and functionality.

core/middleware/api_auth.go
```diff
@@ -0,0 +1,117 @@
...
+		}
+	}
+	return false
+}
+
+func GenerateMD5(param string) string {
+	hash := md5.New()
+	hash.Write([]byte(param))
+	return hex.EncodeToString(hash.Sum(nil))
+}
```
wanghe-fit2cloud, line 117: The provided code has several issues and areas for improvement:

### Issues:
1. **Magic Numbers**: The use of constants like `tolerance` and `60` across multiple places could lead to magic numbers. Consider defining these as named variables within functions where they're used.
  
2. **Error Handling in Logs**: The logging calls can be optimized by using formatted logs instead of printing raw errors directly to logs.

3. **Repetition in Code**: Functions like `isValid1PanelTimestamp`, `isValid1PanelToken`, and `GenerateMD5` have similar logic repeated in different methods. This redundancy can be streamlined.

4. **Security Concerns**:
   - The MD5 hashing function should be replaced with stronger cryptographic algorithms that resist collision attacks, such as SHA-256.
   
5. **Documentation**: Comments are present but not detailed enough to explain how each part of the code works.
 
6. **Performance**: Ensure that parsing IP addresses and validating CIDRs efficiently.

### Optimization Suggestions:
1. **Use Named Variables for Tolerances**: For example, instead of writing `int64(60)` repeatedly, define them at the top level or in local contexts.

Example refactored version:
```go
const (
    defaultTolerance = 60 // seconds tolerance allowed
)

// Define other constants similarly

func ValidateAPITokenAndPermissionMiddleware() gin.HandlerFunc {
    return func(c *gin.Context) {
        if strings.HasPrefix(c.Request.URL.Path, "/api/v2/core/auth") {
            c.Next()
            return
        }

        panelToken := c.GetHeader("1Panel-Token")
        apiKeyTimestamp := c.GetHeader("1Panel-Timestamp")

        // Check API interface status and timestamps etc...
    }
}

func IsValidAPIKeyTimestamp(timestampStr string, validTimeInSeconds int, currentTimeOffsetSeconds int) bool {
    currentUnixTime := getCurrentUnixTime(currentTimeOffsetSeconds)

    parsedTimestamp, err := strconv.ParseInt(timestampStr, 10, 64)
    if err != nil {
        log.Printf("Failed to parse timestamp: %v\n", err)
        return false
    }

    thresholdTime := currentUnixTime + time.Duration(validTimeInSeconds)*time.Second

    if parsedTimestamp > thresholdTime {
        log.Println("Invalid API key timestamp.")
        return false
    }

    return true
}

func GetCurrentUnixTime(offsetSeconds int) int64 {
    current := time.Now().Unix()

    // Apply offset if needed
    current += int64(offsetSeconds)
    return current
}

func GenerateSecureHash(message string) string {
    hash := sha256.New()
    hash.Write([]byte(message))
    return hex.EncodeToString(hash.Sum(nil))
}
```

This refactoring reduces repetition, improves logging quality by formatting messages properly, and enhances security by utilizing a stronger encryption method (`sha256`) and ensuring proper error handling and parameter validation.

wanghe-fit2cloud (commented):
/lgtm

PR #8269, opened 2025-03-28
feat(website): Restrict unhealthy application selection during websit…

zhengkunwang223:


frontend/src/views/website/website/create/index.vue
```diff
@@ -107,13 +107,19 @@
...
+                                :disabled="appInstall.status !== 'Running'"
+                            >
+                                <div class="flex justify-between items-center w-full">
+                                    <span>{{ appInstall.name }}</span>
+                                    <span><Status :key="appInstall.status" :status="appInstall.status"></Status></span>
+                                </div>
+                            </el-option>
                         </el-select>
                     </el-form-item>
                     <div v-if="website.appType == 'new'">
```
wanghe-fit2cloud, line 125: In the provided code change, it has been updated to make the `el-select` wider (300px instead of 200px) and ensure that only installed apps with a status of "Running" appear in the dropdown. Additionally, each option now includes a div containing both the app name and its corresponding status, formatted using the `<Status>` component.

**OptimizationSuggestions:**
- Use Vue's computed property for dynamic data binding if `appInstalls` list frequently changes.
- Ensure that there is an appropriate error state handling in case no running installations exist.
- Consider caching the result of checking the application status rather than calling a new API on every render for better performance.

frontend/src/views/login/components/login-form.vue
```diff
@@ -346,12 +347,15 @@ const login = (formEl: FormInstance | undefined) => {
...
                 if (res.message === 'ErrAuth') {
                     globalStore.ignoreCaptcha = false;
                     errCaptcha.value = false;
                     errAuthInfo.value = true;
+                    return;
                 }
+                MsgError(res.message);
             }
             loginVerify();
         } finally {
```
wanghe-fit2cloud, line 361: The code provided has several minor changes and suggestions to improve it:

1. **Import Update**: The `MsgSuccess` import statement was updated from `'@/utils/message'` to `'@/utils/message'`. This is unnecessary since the same module is being imported again.

2. **Console Log Removal**: The comment about logging an error during login (`console.log('loiin syc');`) should be removed as it serves no purpose.

3. **Return Statements**: 
   - In both cases where the response message is "ErlAuth" or another authentication error, return statements are added after handling these scenarios. This ensures that the function does not proceed past the error checks.
   
Here's the optimized version of the code:
```javascript
const login = async (formEl: FormInstance | undefined) => {
    if (!formEl)
        return;

    await formEl.validate((valid) => {
        if (valid) {
            try {
                await loginApi({
                    username: loginForm.username,
                    password: encryptedPassword(loginForm.password),
                    // Add other necessary parameters here if required
                });

                tabsStore.removeAllTabs();

                document.onkeydown = null;
            } catch (res) {
                if (res.message === 'ErrCaptcha') {
                    loginForm.captcha = '';
                    errCaptcha.value = true;
                    errAuthInfo.value = false;
                    return;
                }
                if (res.message === 'ErrAuth' || res.message.includes("error occurred")) {
                    globalStore.ignoreCaptcha = false;
                    errCaptcha.value = false;
                    errAuthInfo.value = true;
                    return;
                }
                MsgError(res.message);
            }

            loginVerify();
        }
    });
};
```

### Key Changes and Suggestions:
- **Removed Import Duplicates**: Eliminated redundant imports for readability.
- **Console Logging Removal**: Removed unneeded console log statement.
- **Added Return Statements**: Ensured functions terminate early when appropriate authentication errors occur.
- Improved variable names and formatting for better readability.

core/app/api/v2/auth.go
```diff
@@ -39,7 +39,7 @@ func (b *BaseApi) Login(c *gin.Context) {
 
 	user, msgKey, err := authService.Login(c, req, string(entrance))
 	go saveLoginLogs(c, err)
-	if msgKey == "ErrAuth" {
+	if msgKey == "ErrAuth" || msgKey == "ErrEntrance" {
 		helper.BadAuth(c, msgKey, err)
 		return
 	}
```
wanghe-fit2cloud, line 45: There seems to be an issue with the `SaveLoginLogs` function call. The method expects a single argument, but you are passing two arguments (`c` and `err`). You should modify it to correctly pass just the one required parameter.

Additionally, there's no need to use `go` keyword here; calling this function directly will suffice. Here is the corrected line:

``` go
saveLoginLogs(c, err)
```

Here's the updated Go function:

``` go
func (b *BaseApi) Login(c *gin.Context) {
        
    user, msgKey, err := authService.Login(c, req, string(entrance))

    if msgKey == "ErrAuth" || msgKey == "ErrEntrance" {
        helper.BadAuth(c, msgKey, err)
        return
    }

    // Optionally, handle other scenarios based on msgKey

}
```

This ensures that the function `saveLoginLogs` receives only the necessary parameter.

wanghe-fit2cloud (commented):
/lgtm

PR #8288, opened 2025-04-01
fix: Fix the exception of refreshing the token of Ali Cloud Disk

ssongliu:


agent/app/service/logs.go
```diff
@@ -64,26 +61,3 @@ func (u *LogService) ListSystemLogFile() ([]string, error) {
...
-		if err := handleGunzip(fileGzPath); err != nil {
-			return "", fmt.Errorf("handle ungzip file %s failed, err: %v", fileGzPath, err)
-		}
-	}
-	content, err := os.ReadFile(filePath)
-	if err != nil {
-		return "", err
-	}
-	return string(content), nil
-}
```
wanghe-fit2cloud, line 89: Code differences: The method `LoadSystemLog` checks for existence of both `.log` and `.gz` files with the current date suffix but only returns the content of the `.log` file if it exists after uncompression. This might not handle cases where users log in with a previous date which does not have a corresponding `.gz` file.

Optimization suggestion: Instead of directly returning the contents of the `.log` file, consider handling both `.log` and `.gz` files within the same method to ensure consistency, especially for logging into any day prior to today's logging period.

Additional notes: Ensure that the `global.Dir.LogDir` is correctly set up and accessible. Also, make sure there are no typos or unnecessary lines left from earlier version changes.

core/app/service/logs.go
```diff
@@ -44,39 +36,6 @@ func (u *LogService) CreateLoginLog(operation model.LoginLog) error {
...
-	sort.Slice(files, func(i, j int) bool {
-		return files[i] > files[j]
-	})
-
-	return files, nil
-}
-
 func (u *LogService) PageLoginLog(ctx *gin.Context, req dto.SearchLgLogWithPage) (int64, interface{}, error) {
 	options := []global.DBOption{
 		repo.WithOrderBy("created_at desc"),
```
wanghe-fit2cloud, line 41: The provided Go code snippet contains several discrepancies, potential issues, and optimizations suggestions:

1. **Imports**:
   - The `os` package is commented out but used in another line (`path.Join(global.CONF.Base.InstallDir, "1panel/log")`). This might not be intended.
   - Consider importing only necessary packages to avoid cluttering the imports section.

2. **Method Signature**:
   - In method signatures like `CreateLoginLog`, redundant parameter names such as `ctx` from `*gin.Context` should be omitted for clarity. For example, just use `context`.

3. **Function Logic**:
   - Method `ListSystemLogFile()` uses `filepath.Walk` with a filter that includes all `.log` files except `1Panel.log`. However, this logic seems more complex than needed. Instead of listing all log files every time, consider loading them on demand and caching results or filtering when accessing individual entries.

   ```go
   // Example optimized version using cache
   var systemLogsCache map[string]*FileInfo

   func ListSystemLogFile() ([]string, error) {
       if systemLogsCache != nil {
           return cachedKeys(systemLogsCache), nil
       }

       files := make(map[string]*UserInfo)
       // Fetch raw data into files slice with key format: YYYY-MM-DD
       loadLogsIntoFiles(files)

       sortedItems := sortByOldest(files)
       logEntries := convertToEntry(sortedItems)

       syslogs := []string{}
       for _, entry := range logEntries {
           syslogs = append(syslogs, entry.Time.Format("2006-01-02"))
       }

       systemLogsCache = newSyslogCache(entries)
       return getCachedKeys(systemLogsCache), nil
   }
   
   // Load raw logs into files map
   func loadLogsIntoFiles(target map[string]*UserInfo) {}
   
   // Sort files by oldest date
   func sortByOldest(data map[string])*[]struct {Key string; Value UserInfo} {}
   
   // Convert keys and values to structured Entry slices
   func convertToEntry(slice []*struct {Key string; Value UserInfo}) []
   ```

4. **Documentation and Comments**:
   - While existing comments are present, ensure they describe the logic well enough without redundancy. Avoid using overly verbose comments unless necessary.

5. **Error Handling**:
   - Check errors at appropriate points to handle them effectively instead of silently ignoring them (e.g., returning all possible non-nil errors during initialization).

By addressing these aspects, the code will enhance its readability, performance, and maintainability while being free from common pitfalls associated with repetitive task execution and inefficient data handling.

agent/app/service/ssh.go
```diff
@@ -288,25 +287,26 @@ func (u *SSHService) LoadLog(ctx *gin.Context, req dto.SearchSSHLog) (*dto.SSHLo
...
+		if _, err := os.Stat(itemFileName); err != nil && os.IsNotExist(err) {
+			if err := handleGunzip(itemPath); err == nil {
+				fileList = append(fileList, sshFileItem{Name: itemFileName, Year: info.ModTime().Year()})
 			}
 		}
-		return nil
-	}); err != nil {
-		return nil, err
 	}
 	fileList = sortFileList(fileList)
```
wanghe-fit2cloud, line 312: Changes:

1. The use of `filepath.Walk` has been replaced with `os.ReadDir`, which simplifies directory traversal without handling errors directly.

2. Removed unused imports: `path/filepath`.

3. Simplified the condition for processing files from directories that do not start with "secure" or "auth".

4. Updated the error handling to return early when an error occurs, improving performance.

5. Added type assertions `_`, since `Info()` returns `os.FileInfo` but is only used on a boolean expression.

6. Simplified the logic for reading non-gzipped files and unconditionally appending them to the list.

7. Moved some initialization and variable declarations closer to where they are used, reducing scope and improving readability.

8. Used `continue` for skipping unnecessary operations rather than returning inside the loop.

wanghe-fit2cloud (commented):
/lgtm

PR #8330, opened 2025-04-07
fix: Fix the problem that the theme color display is abnormal after t…

ssongliu:
…he upgrade

### DIFF
--- a/frontend/src/views/login/components/login-form.vue
+++ b/frontend/src/views/login/components/login-form.vue
@@ -411,8 +411,7 @@ const getSetting = async () => {
         i18n.warnHtmlMessage = false;
         globalStore.setOpenMenuTabs(res.data.menuTabs === 'Enable');
         globalStore.updateLanguage(res.data.language);
-        let theme = globalStore.themeConfig.theme === res.data.theme ? res.data.theme : globalStore.themeConfig.theme;
-        globalStore.setThemeConfig({ ...themeConfig.value, theme: theme, panelName: res.data.panelName });
+        globalStore.setThemeConfig({ ...themeConfig.value, theme: res.data.theme, panelName: res.data.panelName });
     } catch (error) {}
 };
 
--- a/frontend/src/views/setting/panel/index.vue
+++ b/frontend/src/views/setting/panel/index.vue
@@ -327,6 +327,7 @@ const search = async () => {
     form.apiKeyValidityTime = res.data.apiKeyValidityTime;
     form.hideMenu = res.data.hideMenu;
     form.systemIP = res.data.systemIP;
+    form.theme = res.data.theme;
 
     if (isMasterProductPro.value) {
         const xpackRes = await getXpackSetting();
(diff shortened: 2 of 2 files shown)


frontend/src/views/login/components/login-form.vue
```diff
@@ -411,8 +411,7 @@ const getSetting = async () => {
         i18n.warnHtmlMessage = false;
         globalStore.setOpenMenuTabs(res.data.menuTabs === 'Enable');
         globalStore.updateLanguage(res.data.language);
-        let theme = globalStore.themeConfig.theme === res.data.theme ? res.data.theme : globalStore.themeConfig.theme;
-        globalStore.setThemeConfig({ ...themeConfig.value, theme: theme, panelName: res.data.panelName });
+        globalStore.setThemeConfig({ ...themeConfig.value, theme: res.data.theme, panelName: res.data.panelName });
     } catch (error) {}
 };
```
wanghe-fit2cloud, line 417: The provided `getSetting` function is largely consistent with other code patterns, but there's one small optimization suggestion:

```diff
-        globalStore.setThemeConfig({ ...themeConfig.value, theme: theme, panelName: res.data.panelName });
+        globalStore.setThemeConfig({
```

Replace the last line of your change block with the corrected version to ensure all necessary fields are included in the object you're passing to `globalStore.setThemeConfig()`. This makes sure that the updated configuration includes both the new theme and the new panel name, rather than just merging them without explicitly specifying the field names.

frontend/src/views/setting/panel/index.vue
```diff
@@ -340,7 +341,7 @@ const search = async () => {
             form.proxyDocker = xpackRes.data.proxyDocker;
         }
     } else {
-        form.theme = globalStore.themeConfig.theme || res.data.theme || 'light';
+        globalStore.themeConfig.theme = form.theme;
     }
 };
```
wanghe-fit2cloud, line 347: The code snippet you've provided contains a few potential issues:

1. In version 330, `globalStore.themeConfig.theme` is set to the same value from res.data.theme using an assignment (`= `). This means that `res.data.theme` cannot be used again later in this scope without losing its original value.

To fix this issue, update `globalStore.themeConfig.theme = form.theme;`.

2. The variable `xpackRes` might not be defined before attempting to use it inside the if-statement.

Additionally, there's no need to reassign `globalStore.themeConfig.theme = form.theme;` since you can directly access `form.theme` without affecting other parts of the program. A more readable approach would be to use ternary operator or conditional statement instead.

Optimization suggestion:
Use template literals for string concatenation, which makes the code more compact and easier to read.

After making these changes, the corrected code should look like this:

```javascript
const search = async () => {
    // ... previous lines ...

    globalStore.themeConfig.theme = form.theme;

    if (isMasterProductPro.value) {
        const xpackRes = await getXpackSetting();
        // ... rest of the code ...
    } else {
        // Reuse the value of form.theme stored previously 
        // No need to assign it here again.
    }

    // Further processing using form.theme...
};
```

With these updates, your code will behave correctly and efficiently.

wanghe-fit2cloud (commented):
/lgtm

PR #8332, opened 2025-04-07
fix: Set bridge as the default network for container creation.

ssongliu:


frontend/src/views/container/container/index.vue
```diff
@@ -546,6 +546,7 @@ const onOpenDialog = async (
     rowData: Partial<Container.ContainerHelper> = {
         cmd: [],
         cmdStr: '',
+        network: 'bridge',
         publishAllPorts: false,
         exposedPorts: [],
         cpuShares: 1024,
```
wanghe-fit2cloud, line 552: The provided code is mostly clean and well-formatted. However, there are two minor areas that could be improved:

1. **Network Configuration**: The default `network` value set to `'bridge'` is generally considered sufficient in most contexts. However, if you have specific requirements or prefer a different network setup (e.g., Docker Swarm), you might want to introduce an option for users to select the desired network type.

2. **Initialization of Variables**: While it's not strictly necessary to explicitly initialize these variables with values (`cmd`, `cmdStr`, etc.), this can help future developers understand the expected state when using the function without checking individual assignments later in the code. If no initial values are intended, you could omit them entirely.

Here's an example of how you might modify the code slightly for clarity:

```typescript
const onOpenDialog = async (
  rowData: Partial<Container.ContainerHelper> | undefined = undefined,
): Promise<void> => {
  let rowData_: Container.ContainerHelper;
  
  // Initialize rowData_ if rowData was not provided
  if (!rowData) {
    rowData_ = { cmd: [], cmdStr: '', network: 'bridge', publishAllPorts: false, exposedPorts: [], cpuShares: 1024 };
  } else {
    rowData_ = rowData;
  }

  // Your other logic here...
};
```

These changes ensure that `rowData` defaults to initializing with known properties, enhancing readability and maintainability.

wanghe-fit2cloud (commented):
/lgtm

PR #8351, opened 2025-04-08
 fix: Fix the problem of failed MySQL permission modification

ssongliu:


backend/utils/mysql/client/local.go
```diff
@@ -188,18 +188,19 @@ func (r *Local) ChangeAccess(info AccessChangeInfo) error {
...
+		Permission:  info.OldPermission,
+		ForceDelete: true,
+		Timeout:     300}); err != nil {
+		return err
+	}
+	if info.Username == "root" {
+		return nil
 	}
 	if err := r.CreateUser(CreateInfo{
 		Name:       info.Name,
```
wanghe-fit2cloud, line 206: The code provided contains a few issues and potential improvements:

1. **Redundancy in `Delete` Method Call**: The first branch of the conditional statement deletes the user with old permissions, while the second branch does not. This can lead to unintended behavior if there are no existing users with that permission.

2. **Handling Root User Deletion**: The root user is handled differently based on whether the condition is met within the loop or after it. However, this logic seems unnecessary given how the root user deletion should only occur when all checks pass.

3. **Optimization Suggestion**:
   - Remove the redundant `info.Name = "*"`. It doesn't affect function behavior but could be removed for clarity.

Here's the cleaned-up version:

```go
func (r *Local) ChangeAccess(info AccessChangeInfo) error {
	info.Password = r.Password

	switch {
	case info.Permission != "":
		if err := r.Delete(DeleteInfo{
			Version:     info.Version,
			Username:    info.Username,
			Permission:  info.OldPermission,
			ForceDelete: true,
			Timeout:     300}); err != nil {
			return err
		}

		if info.Username == "root" {
			return nil
		}
	default:
		if err := r.createUser(CreateInfo{
			Name:       info.Name,
            // Rest of the createUser parameters
		}); err != nil {
			return err
		}
		return r.changeUserAccess(UserModifyOptions{ // Assuming changeUserAccess exists
			OldPermissions: []string{info.OldPermission},
			NewPermissions: []string{info.Permission},
			Usernames:      []string{info.Username},
			Version:        info.Version,
		})
	}
}
```

### Key Improvements Made:
- Simplified the conditional structure using a switch, making the code more concise.
- Removed the `info.Name = "*"` assignment since it wasn't doing anything meaningful.
- Added a default case to handle cases where `info.Permission` isn’t an empty string, ensuring proper handling of both new and permission changes.

backend/utils/mysql/client/remote.go
```diff
@@ -199,18 +199,19 @@ func (r *Remote) ChangeAccess(info AccessChangeInfo) error {
...
+		Permission:  info.OldPermission,
+		ForceDelete: true,
+		Timeout:     300}); err != nil {
+		return err
+	}
+	if info.Username == "root" {
+		return nil
 	}
 	if err := r.CreateUser(CreateInfo{
 		Name:       info.Name,
```
wanghe-fit2cloud, line 217: There is no significant irregularity, issue, or optimization suggestion in the provided code snippet. The changes you made to modify permission handling do not seem to introduce any new problems or improve efficiency beyond what was already present in the original version.

backend/app/service/cronjob_helper.go
```diff
@@ -350,12 +350,18 @@ func (u *CronjobService) removeExpiredBackup(cronjob model.Cronjob, accountMap m
...
 			}
 			_ = snapshotRepo.Delete(commonRepo.WithByName(strings.TrimSuffix(records[i].FileName, ".tar.gz")))
 		} else {
 			for _, account := range accounts {
+				if _, ok := accountMap[account]; !ok {
+					continue
+				}
 				if len(account) != 0 {
 					_, _ = accountMap[account].client.Delete(path.Join(accountMap[account].backupPath, records[i].FileDir, records[i].FileName))
 				}
```
wanghe-fit2cloud, line 367: The changes made do not have significant irregularities or issues.

However, there is room for improvement:

1. **Comments**: The comment about skipping accounts that cannot be found in the map can be moved outside of the loop if you want to reduce redundant checks within each iteration cycle.
   
   ```go
   for _, account := range accounts {
       if _, ok := accountMap[account]; !ok {
           continue
       }
       // Rest of the deletion logic
   }
   ```

2. **Code Readability**:
   - The line `if _, ok := accountMap[account]; !ok {` could use `continue;` on its own per iteration instead of repeating the expression multiple times.
   - Consider extracting the client creation logic into a helper function to maintain cleaner code structure.

Overall, the patch looks well-maintained and optimized given the current constraints.

wanghe-fit2cloud (commented):
/lgtm

PR #8353, opened 2025-04-08
fix: Fix the problem that jumping to a directory that does not exist …

ssongliu:


backend/app/api/v1/file.go
```diff
@@ -410,11 +410,20 @@ func (b *BaseApi) CheckFile(c *gin.Context) {
...
+			return
+		}
+		helper.SuccessWithData(c, true)
 		return
 	}
-	helper.SuccessWithData(c, true)
+	helper.SuccessWithData(c, false)
 }
 
 // @Tags File
```
wanghe-fit2cloud, line 429: The code appears to be designed for checking whether a file exists at the specified path or creating it if necessary. However, there are a few areas that could be improved:

1. **Simplify Error Handling**: The error handling logic is somewhat repetitive. Instead of repeating `helper.SuccessWithData(c, false)` multiple times inside different branches of the conditional statements, consider consolidating those into a single call after detecting an error.

   ```go
   if err := helper.CheckBindAndValidate(&req, c); err != nil {
       return
   }

   ...
```

2. **Use Consistent Function Calls**: Ensure consistent function calls across the codebase. For example, use the same method name (`Stat`), parameter list, and context parameters for operations like checking existence and directory creation.

   ```go
   fileOp := files.NewFileOp()
   
   ...
```

3. **Return Value Check**: After attempting to perform operations, check the result before returning success or failure directly. This helps avoid unexpected behavior due to asynchronous operations or incomplete setup.

```go
if !fileOp.Stat(req.Path) {
    // Handle specific cases where file does not exist
} else if err := fileOp.CreateDir(req.Path, 0644); err != nil {
    helper.SuccessWithData(c, false)
    // Additional logs or retry can be added here
}
```

4. **Logging for Debugging**: Consider adding logging within error paths to help debug issues more easily. You can use Go's standard library log package or third-party libraries like zap.

```go
import "log"

func (b *BaseApi) CheckFile(c *gin.Context) {
    // Existing code
}

...

log.Fatalf("Failed to create directory %s: %v", req.Path, err)

...
```

5. **Comments for Complex Logic**: Add comments around complex sections of code that might not be immediately clear to others reading the code.

By making these improvements, the code will be clearer, maintainable, and less prone to errors.

frontend/src/views/setting/snapshot/import/index.vue
```diff
@@ -109,7 +110,12 @@ const checkDisable = (val: string) => {
...
-    router.push({ path: '/hosts/files', query: { path: backupPath.value } });
+    const res = await CheckFile(backupPath.value, true);
+    if (res.data) {
+        router.push({ path: '/hosts/files', query: { path: backupPath.value } });
+    } else {
+        MsgError(i18n.global.t('file.noSuchFile'));
+    }
 };
 
 const submitImport = async (formEl: FormInstance | undefined) => {
```
wanghe-fit2cloud, line 121: 1. The function `toFolder` now includes an asynchronous check before navigating to the files page. It calls the `CheckFile` API to verify if the file exists at the specified path. If the file does not exist, it displays a message error.

   ```javascript
       const res = await CheckFile(backupPath.value, true);
       if (res.data) {
           router.push({ path: '/hosts/files', query: { path: backupPath.value } });
       } else {
           MsgError(i18n.global.t('file.noSuchFile'));
       }
   ```
   
2. In the `submitImport` function, you should handle errors properly when using promises returned by the Axios request (`snapshotImport`). Use `.catch()` to catch the rejected promise and display appropriate messages if something goes wrong during the upload process. Additionally, ensure that all variables used within the `.then()` block are correctly scoped.

   ```javascript
           snapshotImport(params).then(() => {
               // Handle success, e.g., hide dialog, clear fields, notify user
               drawerVisible.value = false;
               formRef.value?.resetFields();
           }).catch(error => {
               console.error("Upload failed:", error); // Log error for debugging purposes
               MsgError(i18n.global.t('upload.fail')); // Display error message
           });
   ```

These changes improve robustness of your application, ensuring better user experience in certain situations where there may be issues with the data being processed.

frontend/src/api/modules/files.ts
```diff
@@ -49,8 +49,8 @@ export const SaveFileContent = (params: File.FileEdit) => {
     return http.post<File.File>('files/save', params);
 };
 
-export const CheckFile = (path: string) => {
-    return http.post<boolean>('files/check', { path: path });
+export const CheckFile = (path: string, withInit: boolean) => {
+    return http.post<boolean>('files/check', { path: path, withInit: withInit });
 };
 
 export const BatchCheckFiles = (paths: string[]) => {
```
wanghe-fit2cloud, line 56: The provided code contains a minor issue in the `CheckFile` function:

1. **Parameter Change**: The `CheckFile` function now takes an additional parameter `withInit`. This appears to be intended as part of the request data being sent to the server, but there are no changes needed in the logic.

2. **Return Type**: The response from both functions is expected to be either a boolean indicating success or failure (`boolean`). No other types should typically be returned unless they have specific meanings based on context.

Overall, the main difference is simply adding another attribute to one of the function parameters and not altering the return type or logic significantly.

wanghe-fit2cloud (commented):
/lgtm

PR #8355, opened 2025-04-09
feat: update allowPort check

zhengkunwang223:


frontend/src/views/ai/mcp/server/operate/index.vue
```diff
@@ -186,16 +194,12 @@ const acceptParams = async (params: AI.McpServer) => {
...
-        } catch (error) {}
+        if (mcpDomainRes.data && mcpDomainRes.data.connUrl != '') {
+            const parts = mcpDomainRes.data.connUrl.split(/(https?:\/\/)/).filter(Boolean);
+            mcpServer.value.protocol = parts[0];
+            mcpServer.value.url = parts[1];
+            mcpServer.value.baseUrl = mcpDomainRes.data.connUrl;
+        }
     }
     open.value = true;
 };
```
wanghe-fit2cloud, line 205: There are no obvious issues with the provided code fragment, but there's a minor suggestion for improvement:

### Optimizations/Suggestions

- **Avoid Unnecessary Try/Catch**: The `try/catch` block around API calls is not strictly necessary here, especially given that we only care about handling potential errors when setting up the website URL (if available). We can simplify this by removing the `catch` block.

Here's the revised version of the relevant part:

```javascript
const acceptParams = async (params: AI.McpServer) => {
    hasWebsite.value = false;
    mode.value = params.id ? 'edit' : 'create';
    
    try {
        const mcpDomainRes = await getMcpDomain();
        
        if (mcpDomainRes.data.connUrl != '') {
            hasWebsite.value = true;
        }
    } finally { // Ensure that the state is correctly updated regardless of success or failure
        if (hasWebsite.value) {
            let parts = mcpDomainRes.data.connUrl.split(/(https?:\/\/)/).filter(Boolean);
            mcpServer.value.protocol = parts[0];
            mcpServer.value.url = parts[1];
            mcpServer.value.baseUrl = mcpDomainRes.data.connUrl;
        }
    }

    ...
}      
```

If you want to explicitly handle errors without affecting the logic flow elsewhere, consider using logging instead or wrapping it in a separate error handler function that checks if the operation should be retried. However, for simpler setup workflows, the current approach should suffice.

wanghe-fit2cloud (commented):
/lgtm

PR #8356, opened 2025-04-09
fix: fix issue with mcp bind website

zhengkunwang223:


backend/app/service/mcp_server.go
```diff
@@ -502,13 +507,13 @@ func updateMcpServer(mcpServer *model.McpServer) error {
...
+	_ = mcpServerRepo.Save(mcpServer)
 	composePath := path.Join(constant.McpDir, mcpServer.Name, "docker-compose.yml")
 	_, _ = compose.Down(composePath)
 	if _, err := compose.Up(composePath); err != nil {
 		mcpServer.Status = constant.RuntimeError
 		mcpServer.Message = err.Error()
 	}
-
 	return mcpServerRepo.Save(mcpServer)
 }
```
wanghe-fit2cloud, line 519: The provided code has several issues that can be addressed:

1. **Race Condition**: The `go` functions in the code snippet will not run concurrently due to the lack of a proper goroutine structure. This means that if multiple servers are being updated simultaneously, some updates could be missed.

2. **Concurrency Issues**: The loop that iterates over servers is executed sequentially within each goroutine, which might lead to inefficiencies or unexpected behavior when dealing with many servers.

3. **Missing Return Value in Anonymous Function**: The anonymous function used in the race condition section does not return an error value, which prevents catching errors from `updateMcpServer`.

To address these issues:

1. **Modify Loop Execution**:
   - Use a channel or waitgroup to ensure concurrent execution.

2. **Return Error Values**:
   - Modify the anonymous function to capture and return errors properly.

Here's an optimized version of the affected part of the code:

```go
var wg sync.WaitGroup

for _, server := range servers {
    wg.Add(1)
    go func(s *model.McpServer) {
        defer wg.Done()

        if server.BaseURL != baseUrl {
            server.BaseURL = baseUrl
            server.HostIP = "127.0.0.1"
            if err := updateMcpServer(s); err != nil {
                log.Printf("Failed to update MCP Server %s: %v", s.Name, err)
            } else {
                mcpServerRepo.Save(s) // Ensure this call is successful before updating docker-compose.
            }
        }

        composePath := path.Join(constant.McpDir, s.Name, "docker-compose.yml")
        _, _ = compose.Down(composePath)
        if _, err := compose.Up(composePath); err != nil {
            s.Status = constant.RuntimeError
            s.Message = err.Error()
            mcpServerRepo.Save(s)
        }
    }(server)
}

wg.Wait() // Wait for all goroutines to complete.
```

This version ensures that each server update operation runs concurrently using channels. It also includes logging for failed operations to help debug potential issues later on. Make sure to initialize the waitgroup correctly at the start of the function and ensure `mcpServerRepo.Save` is always called successfully after every relevant action in the goroutine.

backend/app/repo/group.go
```diff
@@ -56,6 +57,12 @@ func (u *GroupRepo) WithByHostDefault() DBOption {
...
 
+func (u *GroupRepo) WithByWebsiteDefault() DBOption {
+	return func(g *gorm.DB) *gorm.DB {
+		return g.Where("is_default = ? AND type = ?", 1, "website")
+	}
+}
+
 func (u *GroupRepo) Delete(opts ...DBOption) error {
 	db := global.DB
 	for _, opt := range opts {
```
wanghe-fit2cloud, line 68: There is one change made in this patch. The function `WithByWebsiteDefault` has been added to the `IGroupRepo` interface and also implemented in the `GroupRepo` struct. This likely adds functionality for filtering groups that are both default and have a specific group type ("website") when performing database queries.

Optimization suggestions: No apparent performance improvements can be directly suggested from this single-line addition, but having more methods on an interface could potentially make the implementation slightly more modular or easier to maintain depending on future use cases.

wanghe-fit2cloud (commented):
/lgtm

PR #8378, opened 2025-04-11
fix: Fix the mysql root account permission switchover exception

ssongliu:


backend/utils/mysql/client/local.go
```diff
@@ -188,19 +188,18 @@ func (r *Local) ChangeAccess(info AccessChangeInfo) error {
...
+			ForceDelete: true,
+			Timeout:     300}); err != nil {
+			return err
+		}
+		if info.Username == "root" {
+			return nil
+		}
 	}
 	if err := r.CreateUser(CreateInfo{
 		Name:       info.Name,
```
wanghe-fit2cloud, line 205: The given code snippet contains two main logic branches:

1. **Regular Branch:**
   - This branch checks if `info.Permission` is equal to `info.OldPermission`.
   - If they are equal, it returns `nil`, effectively skipping the deletion step.
   
2. **Special Case for Root User:**
   - If the current user (`info.Username`) is "root", it also skips the deletion process.

### Potential Issues and Improvements:

1. **Code Reorganization:**
   - It might be beneficial to extract the common logic of checking if `info.Permission == info.OldPermission` into a separate function call to improve readability and maintainability.

2. **Error Handling:**
   - The function does not handle errors returned by `CreateUser`. Although it catches errors from `DeleteInfo`, handling other functions should also be considered within this scope.

3. **Documentation/Comments:**
   - Adding comments explaining each section and key decisions could help clarify the purpose of different parts of the code, especially for others who may read this later.

4. **Variable Naming:**
   - Consistent naming conventions could make the code more readable, such as using camelCase for variable names rather than underscores.

Overall, the overall structure looks clean and well-organized, but these minor improvements can enhance its clarity and robustness.

backend/utils/mysql/client/remote.go
```diff
@@ -199,19 +199,18 @@ func (r *Remote) ChangeAccess(info AccessChangeInfo) error {
...
+			ForceDelete: true,
+			Timeout:     300}); err != nil {
+			return err
+		}
+		if info.Username == "root" {
+			return nil
+		}
 	}
 	if err := r.CreateUser(CreateInfo{
 		Name:       info.Name,
```
wanghe-fit2cloud, line 216: The provided code has several optimizations and corrections:

1. The `if` statement that checks if `info.Permission == info.OldPermission` is unnecessary if it returns before calling `r.Delete`. You can move the call to `r.CreateUser` outside this block.

2. There should be no need to create an empty user object (`CreateInfo{}`) when creating the root user since the `Name` field is already set. This redundancy can be removed.

3. Ensure that there are proper error handling around critical operations like `r.CreateUser()` in case of any failure during creation.

This refactored version of the function ensures cleaner logic by reducing nested structures where possible:

```
func (r *Remote) ChangeAccess(info AccessChangeInfo) error {
	if info.Username == "root" && info.Permission != info.OldPermission {
		return nil
	}

	if err := r.Delete(DeleteInfo{
		Version:     info.Version,
		Username:    info.Username,
		Permission:  info.OldPermission,
		ForceDelete: true,
		Timeout:     300}); err != nil {
		return err
	}

	return r.CreateUser(CreateInfo{Name: info.Name})
}
```

frontend/src/views/database/mysql/password/index.vue
```diff
@@ -167,6 +171,10 @@ const submitChangeInfo = async (formEl: FormInstance | undefined) => {
             }
             return;
         }
+        if (changeForm.privilege === oldPrivilege.value && changeForm.privilegeIPs === oldPrivilegeIPs.value) {
+            changeVisible.value = false;
+            return;
+        }
         if (changeForm.privilege !== 'ip') {
             param.value = changeForm.privilege;
         } else {
```
wanghe-fit2cloud, line 180: There are no major irregularities or obvious issues in this code snippet, but there are some areas that can be optimized or improved:

1. **Consolidate Code**: The two `else` blocks under the conditional checks can be combined to reduce redundancy.

2. **Empty Check**: It's good practice to add an empty string check before comparing strings to avoid unexpected behavior when one of them might be null or undefined.

3. **Simplify Boolean Logic**: You could simplify the boolean logic by checking for equality between arrays explicitly instead of comparing their reference values.

4. **Use Object Destructuring**: If these references will always contain properties with the same names, you could destructure them directly to make the code slightly more concise.

Here is a revised version incorporating these suggestions:

```javascript
if (formEl && formEl.validate(async valid => {
  if (!valid) {            
      return;
  }      
  await validateEmail(changeForm.from);

  if (changeForm.privilege === oldPrivilege.value && equalArrays(changeForm.privilegeIPs, oldPrivilegeIPs.value)) {
      changeVisible.value = false;
      return;
  }

  if (changeForm.privilege !== 'ip') {
      param.value = changeForm.privilege;
  } else {
      // continue processing
  }
})):
```

**Function `equalArrays`:**
If you don't already have such a function, here it is as a helper to compare array elements:

```javascript
function equalArrays(a, b) {
  if (a.length !== b.length)
    return false;

  for (let i = 0; i < a.length; i++) 
      if (a[i] !== b[i])    
          return false;

  return true;
}
```

These improvements should enhance readability, maintainability, and potentially performance marginally.

wanghe-fit2cloud (commented):
/lgtm

PR #8379, opened 2025-04-11
fix(website): fix issue with create website with  local php-fpm failed

zhengkunwang223:


frontend/src/views/website/website/index.vue
```diff
@@ -51,15 +51,14 @@
...
+                <fu-table-column-select
+                    :columns="columns"
+                    trigger="hover"
+                    :title="$t('commons.table.selectColumn')"
+                    popper-class="popper-class"
+                    :only-icon="true"
+                />
             </template>
             <template v-if="!openNginxConfig" #main>
                 <ComplexTable
```
wanghe-fit2cloud, line 64: There are minor changes to your code. Here is an updated list of observations:

1. **Removed Duplicate `div` Element**: The line `<div class="!ml-2.5">...</div>` seems redundant due to the preceding elements.

2. **Consolidated Table Search Component**: You consolidated the table search component by removing the duplicate instantiation with different names (`searchName`).

3. **Renamed Refresh Button to Update**: Changed the name of the refresh button from `refreshButton` to `update`.

Here's a concise version of the updated code block:

```diff
@@ -51,18 +51,17 @@
                     </div>
                 </el-select>
                 <TableSearch @search="search()" v-model:searchName="req.name" />
+                <TableRefresh @update="search()" />
+                <fu-table-column-select
+                    :columns="columns"
+                    trigger="hover"
+                    :title="$t('commons.table.selectColumn')"
+                    popper-class="popper-class"
+                    :only-icon="true"
+                />
             </template>
             <template v-if="!openNginxConfig" #main>
                 <ComplexTable
```

This should eliminate unnecessary redundancy and improve clarity in the component structure.

frontend/src/utils/runtime.ts
```diff
@@ -3,16 +3,22 @@ import { Runtime } from '@/api/interface/runtime';
...
-                row.status === 'Building'
+                row.status === 'Building' ||
+                row.resource == 'local'
             );
         case 'restart':
-            return row.status === 'Recreating' || row.status === 'Building';
+            return row.status === 'Recreating' || row.status === 'Building' || row.resource == 'local';
         case 'edit':
             return row.status === 'Recreating' || row.status === 'Building';
         case 'extension':
```
wanghe-fit2cloud, line 24: There is an issue with this code in terms of logic. Specifically, for the "stop" and "restart" cases, the condition includes rows where `row.resource` equals `'local'`, but it should ideally only include such conditions when `type` equals `"stop"` or `"restart"`. This discrepancy could cause unexpected behavior.

Additionally, for the case where `type` does not match one of the expected values (e.g., `null`), the function will still return `false` because all other checks will return `true`.

Here's a more precise implementation:

```javascript
import { Runtime } from '@/api/interface/runtime';

export function disabledButton(row: Runtime.Runtime, type: string): boolean {
    switch (type) {
        case 'stop':
            return (
                row.status === 'Recreating' ||
                row.status === 'Stopped' ||
                row.status === 'Building'
            );
        case 'start':
            return (
                row.status !== 'Started' &&
                row.status !== 'FailedRestarting'
            ) && ( // Ensure we don't start if already started/restarting
                    row.status !== 'Recreating' ||
                    row.status !== 'Stopped' ||
                    row.status !== 'Building'
                )
            ) ? false : true; // True if starting is allowed by resource status

        case 'restart': // Handle restart slightly differently to allow local resources
            return row.status === 'ReCreating' || row.status === 'Building' || row.resource === 'local';

        case 'edit':
            return row.status === 'Recreating' || row.status === 'Building';
        case 'extension':
            return row.status === 'Initializing'; // Assuming Initializing means creating or starting
        default:
            console.warn(`Unsupported operation type ${type}. Returning true.`);
            return true;
    }
}
```

Changes Made:
- Corrected conditions for "stop" and "restart".
- Added additional conditions for handling non-supported types.
- Ensured that the logic for stopping and restarting accounts for whether the server can run on a local device based on its resource status (`rows.resource = 'local'`).

frontend/src/views/website/website/create/index.vue
```diff
@@ -808,9 +808,13 @@ const changeRuntime = (runID: number) => {
...
+            if (runtimeResource.value == 'local') {
+                website.value.port = 9000;
+            } else {
+                runtimePorts.value = item.port.split(',').map((port: string) => parseInt(port.trim(), 10));
+                if (runtimePorts.value.length > 0) {
+                    website.value.port = runtimePorts.value[0];
+                }
             }
         }
     });
```
wanghe-fit2cloud, line 820: There is an issue with the line `if (runtimeResource.value == 'local')` where it should be `===`. The use of `==` can lead to unexpected behavior and errors, especially when comparing numbers versus strings. Additionally, the logic for setting the WebSocket port could be simplified by handling the case when both 'docker' and 'remote' resources exist but no explicit port is set in the configuration.

Here's a corrected version:

```javascript
const changeRuntime = (runID: number) => {
    runtimes.value.forEach((item) => {
        if (item.id === runID) {
            runtimeResource.value = item.resource;
            runtimePorts.value = item.port ? item.port.split(',').map((port: string) => parseInt(port.trim(), 10)) : [];
            website.value.port = runtime ports.length > 0 ? runtimePorts[0] : null;

            // Optimization suggestion: Handle cases when multiple resource types are present
            if ((item.resource === 'docker' && !website.value.port) || 
                (item.resource === 'remote' && !website.value.port)) {
                console.warn(`No specified port found for resource type '${item.resource}'. Using default port.`);
                website.value.port = 9000;
            }

            break; // Once a matching run ID is found, exit the loop
        }
    });
};
```

In this version:
- The comparison operator used for checking `runtimeResource.value` has been changed to `===`.
- A check for whether `Website.value.port` is set within the ternary operation ensures that we only update the port if there is one available, avoiding unnecessary checks.
- An additional conditional block has been added to log warnings and potentially set a default port (`9000`) if none is explicitly configured for certain types of resources. This simplifies the code by reducing redundant conditions.

wanghe-fit2cloud (commented):
/lgtm

PR #8452, opened 2025-04-22
feat: File management optimization

lan-yonghui:
Refs: #3545 #3387

### DIFF
--- a/agent/app/service/file.go
+++ b/agent/app/service/file.go
@@ -1,13 +1,18 @@
 package service
 
 import (
+	"bufio"
 	"context"
 	"fmt"
+	"github.com/1Panel-dev/1Panel/agent/app/dto"
 	"io"
 	"io/fs"
 	"os"
+	"os/user"
 	"path"
 	"path/filepath"
+	"sort"
+	"strconv"
 	"strings"
 	"time"
 	"unicode/utf8"
@@ -55,6 +60,8 @@ type IFileService interface {
 
 	GetPathByType(pathType string) string
 	BatchCheckFiles(req request.FilePathsCheck) []response.ExistFileInfo
+	GetHostMount() []dto.DiskInfo
+	GetUsersAndGroups() (*response.UserGroupResponse, error)
 }
 
 var filteredPaths = []string{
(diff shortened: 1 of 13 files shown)


agent/app/service/file.go
```diff
@@ -544,3 +551,101 @@ func (f *FileService) BatchCheckFiles(req request.FilePathsCheck) []response.Exi
...
+			Username: username,
+			Group:    groupName,
+		})
+		groupSet[groupName] = struct{}{}
+	}
+	if err := scanner.Err(); err != nil {
+		return nil, nil, fmt.Errorf("failed to scan /etc/passwd: %w", err)
+	}
+	return users, groupSet, nil
+}
```
wanghe-fit2cloud, line 651: There are several potential issues and optimizations in the provided code:

### Potential Issues

1. **Security Vulnerabilities**: Accessing `/etc/group` and `/etc/passwd` files directly can pose security risks. Consider using system calls to read these files safely.

2. **Performance**: Reading large lists of users and groups can be slow. Consider caching results if they change infrequently.

3. **Error Handling**: More detailed error messages should be logged or returned to handle unexpected errors gracefully.

4. **Resource Management**: Ensure that resources like file handles are closed properly after use.

5. **Comments and Documentation**: There is little to no documentation explaining the purpose of some functions and fields. It's important for readability and maintenance.

### Optimization Suggestions

1. **Use System Calls**: Instead of using `bufio.Scanner`, consider using Go’s standard library facilities (`os.Read`) for better performance and control over reads.

2. **Caching**: Cache the result of getting valid users and groups to avoid unnecessary repeated scanning on each call.

3. **Logging**: Implement logging around critical sections of the code to help with debugging and monitoring.

Here's an updated version addressing some of these points:

```go
package service

import (
	"context"
	"fmt"
	"golang.org/x/sys/unix"
	"io"
	"log"
	"os"
	"path/filepath"
)

// ... (rest of the code remains mostly unchanged, just comments added)
```

### Logging Example

To add basic logging, you could modify the function where the disk information is loaded:

```go
func getValidGroups() (map[string]bool, error) {
	log.Printf("Loading group info from /etc/group...")
	groupFile, err := os.Open("/etc/group")
	if err != nil {
		log.Fatalf("Failed to open /etc/group: %v", err)
	}
	defer groupFile.Close()

	groupMap := make(map[string]bool)
	scanner := bufio.NewScanner(groupFile)
	for scanner.Scan() {
		...
	}
	if err := scanner.Err(); err != nil {
		log.Fatalf("Failed to scan /etc/group: %v", err)
	}
	return groupMap, nil
}

// This pattern applies similarly to other functions using / etc/passwd
```

This will allow you to track which methods are being called and what might have gone wrong, helping during troubleshooting and maintenance.

frontend/src/views/host/file-management/terminal/index.vue
```diff
@@ -0,0 +1,47 @@
...
+
+function handleClose() {
+    onClose();
+    terminalVisible.value = false;
+}
+
+defineExpose({
+    acceptParams,
+});
+</script>
```
wanghe-fit2cloud, line 47: This template and script appear to create a Vue component for displaying a terminal within a DrawerPro dialog. Here are some considerations:

1. **Data Handling**: The `acceptParams` function sets the `terminalVisible` state to true, indicating that the terminal should be opened. This is logical.

2. **Initialization Functionality**: The `initTerm` function initializes the terminal with specified arguments:
   - It awaits a tick to ensure the DOM is fully updated before triggering operations on it.
   - It calls `acceptParams` on the terminal instance with specific configuration parameters, including an endpoint, command, and an empty error message (`error: ''`) which might need further clarification if this was intended as part of its functionality.

3. **Event Cleanup**: There's no explicit cleanup logic in place when closing the drawer/dialog unless specifically managed elsewhere (e.g., using lifecycle hooks like `beforeUnmount`). Ensure proper teardown mechanisms exist.

4. **Template Optimization**:
   - Using `size="large"` directly within `<template>` doesn't seem necessary; specifying default values in parent components' configurations typically suffices.
   
Here’s a slightly optimized version of your code snippet:

```typescript
<template>
    <DrawerPro v-model="terminalVisible" :header="$t('menu.terminal')"> <!-- Use default size -->
        <template #content>
            <Terminal style="height: 80vh" ref="terminalRef"> </Terminal>
        </template>
    </DrawerPro>
</template>

<script lang="ts" setup>
import { ref, nextTick } from 'vue';
import Terminal from '@/components/terminal/index.vue';

// Initialize props and methods as per original
const terminalVisible = ref(false);
const terminalRef = ref<typeof Terminal | null>(null);

interface DialogProps {
    cwd: string;
    command: string;
}

const acceptParams = async (params: DialogProps): Promise<void> => {
    terminalVisible.value = true;
    await initTerm(params.cwd);
};

const initTerm = async (cwd: string) => {
    await nextTick();
    const params = {
        endpoint: '/api/v2/hosts/terminal',
        args: encodeURIComponent(`clear && cd ${cwd}`),
        error: '',
        initCmd: '',
    };
    try {
        terminalRef.value!.acceptParams(params); // Assuming acceptParams can handle promises or callbacks nicely
    } catch(error) {
        console.error("Failed to initialize terminal:", error);
        // Optionally inform user of failure through UI update or alert box.
    }
};

// Define expose options if needed
defineExpose({        
    acceptParams,      
});
```

**Recommendations**:
- Check the documentation of `acceptParams` for best practices regarding handling errors and ensuring it returns expected data types.
- Consider adding logging or user feedback upon successful terminal initialization.
- Review event listeners or other parts of your application where the dialog/closed events occur, especially since `Dialog.close()` does not immediately close the dialog but rather emits a close event.

If you encounter issues such as the terminal not initializing correctly or unexpected behavior after clicking "OK", debugging could involve setting breakpoints around `initTerm`, checking network requests related to the terminal interface, or inspecting element states visually via browser developer tools.

frontend/src/components/system-upgrade/index.vue
```diff
@@ -14,31 +14,22 @@
...
+                        <el-link class="ml-2" :underline="false" type="primary" @click="onLoadUpgradeInfo">
+                            {{ $t('commons.button.update') }}
+                        </el-link>
+                    </el-badge>
+                    <el-tag v-if="version === 'Waiting'" round class="ml-2.5">
+                        {{ $t('setting.upgrading') }}
+                    </el-tag>
+                </div>
             </div>
         </div>
```
wanghe-fit2cloud, line 35: The code appears to be functioning correctly for its intended purpose. However, there are a few minor suggestions:

1. The `el-divider` component's vertical position can be adjusted by explicitly setting the margin-top property instead of using `-mt-0.5`. This ensures better consistency with other elements.

2. In the `if (condition)` statements, you might find it useful to add comments explaining what each condition does or linking to external resources if needed.

Here's the revised section with these adjustments:

```html
<div class="flex flex-wrap items-center" v-if="props.footer">
    <el-link type="primary" :underline="false" @click="toForum">
        {{ $t('setting.forum') }}
    </el-link>
    <!-- Add more links here -->
</div>

<!-- Divide the content into two separate divs for clarity -->

<div class="flex justify-between mt-2">  <!-- New line added to set margin top -->
    
    <div class="flex flex-wrap mr-auto gap-x-4">
        
        <!-- Existing code remains unchanged except adding "mr-auto" to center align the content on the right side -->


        <div class="inline-block">  
            <el-link :underline="false" type="primary" @click="toLxware"> 
                {{ !$isMasterProductPro ? $t("license.community") : $t("license.pro") }}  
            </el-link>  
            <span class="version ml-2">  {{ version }}</span>  
            
            <template v-if="!$isMasterProductPro && globalStore.hasNewVersion"> <!-- Change condition from == 'Waiting' to !== 'Idle' --> 

                <el-badge is-dot><i class="ri-notification-line"></i></el-badge>   <!-- Replace '-mt-0.5' with 'mb-px' -->

                <el-link type= primary :underline=false class="hover:text-blue-50">Update Available!</el-link> 

            </template>


            <el-tag
              v-if="$isMasterProductPro || ($isFreeTrialEnabled && trialExpired)"
              size="small"   
              round
              class="text-black"     
          
            >             
              Trial Ended!
            </el-tag>
      </div>



    </div>  

    

</div>```
Note that in this revision, I added conditions to hide badges when there's no new version available while maintaining clear separation between main and auxiliary buttons, which enhances readability and maintainability of the HTML structure. Also corrected some typos such as `$it`, which should ideally refer to translation functions provided via Vue-i18n like so:
```javascript
$t('key');
```

These changes make the template cleaner and potentially slightly easier to read compared to the original one.

wanghe-fit2cloud (commented):
/lgtm

PR #8485, opened 2025-04-27
feat(ssl): add vercel dns provider

zhengkunwang223:
Refs https://github.com/1Panel-dev/1Panel/issues/2025

### DIFF
--- a/agent/utils/ssl/acme.go
+++ b/agent/utils/ssl/acme.go
@@ -243,9 +243,8 @@ func createHTTPClientWithProxy(proxyURL, username, password string) *http.Client
 			TLSHandshakeTimeout:   60 * time.Second,
 			ResponseHeaderTimeout: 60 * time.Second,
 			TLSClientConfig: &tls.Config{
-				InsecureSkipVerify: true,
-				//ServerName: os.Getenv("LEGO_CA_SERVER_NAME"),
-				//RootCAs:    initCertPool(),
+				ServerName: os.Getenv("LEGO_CA_SERVER_NAME"),
+				RootCAs:    initCertPool(),
 			},
 		},
 	}
--- a/agent/utils/ssl/client.go
+++ b/agent/utils/ssl/client.go
@@ -4,10 +4,19 @@ import (
 	"crypto"
 	"encoding/json"
 	"github.com/1Panel-dev/1Panel/agent/app/dto"
+	"github.com/go-acme/lego/v4/providers/dns/alidns"
 	"github.com/go-acme/lego/v4/providers/dns/clouddns"
+	"github.com/go-acme/lego/v4/providers/dns/cloudflare"
+	"github.com/go-acme/lego/v4/providers/dns/dnspod"
 	"github.com/go-acme/lego/v4/providers/dns/freemyip"
+	"github.com/go-acme/lego/v4/providers/dns/godaddy"
 	"github.com/go-acme/lego/v4/providers/dns/huaweicloud"
+	"github.com/go-acme/lego/v4/providers/dns/namecheap"
+	"github.com/go-acme/lego/v4/providers/dns/namedotcom"
+	"github.com/go-acme/lego/v4/providers/dns/namesilo"
 	"github.com/go-acme/lego/v4/providers/dns/rainyun"
+	"github.com/go-acme/lego/v4/providers/dns/tencentcloud"
+	"github.com/go-acme/lego/v4/providers/dns/vercel"
 	"github.com/go-acme/lego/v4/providers/dns/volcengine"
 	"os"
 	"strings"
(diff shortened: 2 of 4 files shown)


agent/utils/ssl/client.go
```diff
@@ -234,7 +234,21 @@ func (c *AcmeClient) UseDns(dnsType DnsType, params string, websiteSSL model.Web
...
+		return nil, err
+	}
+	return p, nil
+}
+
+func (c *AcmeClient) UseDns(dnsType DnsType, params string, websiteSSL model.WebsiteSSL) error {
+	p, err := getDNSProviderConfig(dnsType, params)
 	if err != nil {
 		return err
 	}
```
wanghe-fit2cloud, line 254: 1. The code has two instances where `json.Unmarshall` is called to parse JSON parameters for configuring DNS providers. However, it returns early if `err` occurs on both calls, but the variable `p` holding the provider configuration is never returned or used. This can lead to confusion and possibly errors if other parts of the system depend on this value.

2. A new function `getDNSProviderConfig` is introduced at the beginning of the file with similar functionality as one of the previous uses. However, there's no clear separation between these functions, which could make maintenance more difficult in the future.

3. Instead of using separate constants for each supported cloud service (e.g., Alibaba Cloud, Huawei Cloud), consider introducing an enumeration type that maps strings to DNS types (`DnsType`). This would improve the readability and maintainability of the codebase.

4. It looks like some unused imports should be removed from the end of the file, as they don't appear to be utilized within the functions defined earlier. These could potentially cause compile errors or reduce performance by not needed modules being loaded.

5. When creating DNS configurations via methods specific to certain services (e.g., alidns.NewDefaultConfig()), ensure all necessary fields are configured unless specifically allowed to use default values.

6. Ensure proper error checks after calling `UseDns` across different scenarios and return appropriate feedback if issues arise.

7. The comments describing the various DNS parameter settings and what each field does would benefit from additional details about their expected usage and potential edge cases.

Here is improved version of relevant portions:

```go
// Improved getDNSProviderConfig function definition
func getDNSServiceProvider(dnsType DnsType, params string) (challenge.Provider, error)
{
    ...

    switch dnsType {
        // ... same logic goes here ...
        
        case Vercel:
            vConfig.AuthToken = param.Token
            config.PropagationTimeout = propagationTimeout
            config.PollingInterval = pollingInterval
            
            provider, err := vercel.NewDNSProviderConfig(vConfig)
            
            ...
}

// Usage example: 
provider, _ := getDNSServiceProvider(dnsType, dnsParamJSONString)

// Use acquired provider for ACME client operation:
acmeClient.UseDNS(provider)
```

In summary, this code needs refinement and documentation improvements especially regarding its intended usage and error handling mechanisms. Additionally, modularizing certain functionality into smaller reusable components might enhance readability and support long-term maintainability.

agent/utils/ssl/acme.go
```diff
@@ -243,9 +243,8 @@ func createHTTPClientWithProxy(proxyURL, username, password string) *http.Client
...
 			ResponseHeaderTimeout: 60 * time.Second,
 			TLSClientConfig: &tls.Config{
-				InsecureSkipVerify: true,
-				//ServerName: os.Getenv("LEGO_CA_SERVER_NAME"),
-				//RootCAs:    initCertPool(),
+				ServerName: os.Getenv("LEGO_CA_SERVER_NAME"),
+				RootCAs:    initCertPool(),
 			},
 		},
 	}
```
wanghe-fit2cloud, line 250: The code appears to be correct and optimized for a basic `HttpClient` setup with TLS settings. The proxy handling is done correctly within the existing logic, but here are some minor improvements and notes:

1. **InsecureSkipVerify**: While setting `InsecureSkipVerify` can save time when dealing with self-signed certificates or development environments, it's generally not recommended for production use due to security concerns. If you need to handle untrusted CA certificates, consider using a custom root CA pool.

2. **TLSClientConfig.Secure**: If there are specific reasons to disable certificate verification (e.g., testing against local services), ensure that this is documented clearly because it increases the risk of man-in-the-middle attacks.

3. **Initialization of RootCA Pool**: Ensure that the `initCertPool()` function retrieves the appropriate CA certificates if needed. This could involve loading from files specified via environment variables or hardcoding trusted CAs directly if they don't change frequently.

4. **Code Clarity**: Adding comments might help explain the purpose of each part, especially considering its complexity (especially the inclusion of `LEGO_CA_SERVER_NAME` and `os.Getenv`). It would also make the code easier to understand for those who may review or maintain it in the future.

Overall, the current implementation should work as intended for a secure HTTP client with optional proxy support.

frontend/src/global/mimetype.ts
```diff
@@ -188,6 +188,10 @@ export const DNSTypes = [
         label: 'Cloudflare',
         value: 'CloudFlare',
     },
+    {
+        label: 'Vercel',
+        value: 'Vercel',
+    },
     {
         label: 'CloudDNS',
         value: 'CloudDns',
```
wanghe-fit2cloud, line 197: The provided code seems to be missing semicolons at the end of each object in the `DNSTypes` array. This is important in JavaScript for proper syntax and performance reasons.

Here’s the corrected version with added semicolons:

```javascript
export const DNSTypes = [
    {
        label: 'Cloudflare',
        value: 'CloudFlare',
    },
    {
        label: 'Vercel',
        value: 'Vercel',
    },
    {
        label: 'CloudDNS',
        value: 'CloudDns',
    }
];
```

Additionally, this list does not appear to include any specific requirements related to DNS types or configurations. If you have further questions about optimizing DNS settings or implementing these providers into a system (e.g., managing domains), please let me know!

PR #8488, opened 2025-04-27
feat: Hide the favorites button for installed items.

zhengkunwang223:


frontend/src/views/app-store/installed/index.vue
```diff
@@ -184,7 +184,7 @@
                                                                     </el-button>
                                                                 </el-tooltip>
                                                             </span>
-                                                            <span class="ml-1">
+                                                            <span class="ml-1" v-if="mode === 'installed'">
                                                                 <el-tooltip
                                                                     effect="dark"
                                                                     :content="$t('website.cancelFavorite')"
```
wanghe-fit2cloud, line 190: The code snippet you provided appears to be for an AngularJS component with Bootstrap Vue and i18next libraries. There are no specific inconsistencies, but there is an opportunity for some improvements:

1. **Conditionally Rendering `el-tooltip`:** It seems like the `el-tooltip` should only render when `mode` is `'installed'`. The current code will always render it regardless of `mode`.

2. **Potential Misplacement:** Adding a conditional check using `v-if` on `.ml-1` could lead to unexpected positioning if applied inconsistently elsewhere in the template.

3. **Suggested Enhancement:** You can simplify the logic slightly by removing the redundant `.ml-1` span inside another conditional block where you already have access to control spacing dynamically via CSS or additional classes within your parent element's styling scope.

Here’s a refined version based on these considerations:

```diff
@@ -184,7 +184,7 @@
                                                                 </el-button>
                                                             </el-tooltip>
                                                         </span>
-                                                        <span class="ml-1" v-if="mode === 'installed'">
+                                                        <!-- Assuming mode determines whether this tooltip should show -->
+                                                        <el-tooltip 
+                                                            v-if="mode === 'installed'"  
+                                                            effect="dark" 
+                                                            :content="$t('website.cancelFavorite')" 
                                                                placement="top-end"
                                                          >
                                                              <i class="fas fa-heart mr-1"></i> {{ $t('website.remove') }}
                                                          </el-tooltip>
                                                      </template>

```

This refactors the condition so that the tooltip is rendered only when necessary, adhering more closely to best practices for conditional rendering.

wanghe-fit2cloud (commented):
/lgtm

PR #8490, opened 2025-04-27
feat: Remove drawer exit prompt.

zhengkunwang223:


frontend/src/components/drawer-pro/index.vue
```diff
@@ -115,25 +114,6 @@ const handleBack = () => {
...
-            done();
-        });
-    } else {
-        done();
-    }
-};
-
 const handleClose = () => {
     localOpenPage.value = false;
     globalStore.isFullScreen = false;
```
wanghe-fit2cloud, line 119: The main difference noted in the code is an attempt to enhance user interaction when closing a document editing interface by providing confirmation prompts for non-terminal, form-based drawers. 

**Potential Improvements:**
1. **Use of Confirm Box:** Replace direct calls to `handleClose` directly with confirm boxes where necessary. This could make the UX smoother by giving users immediate feedback on their actions without having to explicitly click another button.
   
Here is a simple example of how you might refactor it:

```javascript
const handleBeforeClose = async (done: () => void) => {
    const isTerminal = drawerContent.value?.classList.contains('terminal');
    if (isTerminal || !(drawerContent.value instanceof HTMLElement)) {
        done();
        return;
    }

    try {
        const response = await ElMessageBox.confirm(
            i18n.global.t('commons.msg.closeDrawerHelper'),
            i18n.global.t('commons.button.close'),
            {
                type: 'warning',
            }
        );

        // User confirmed closure, proceed to close the dialog box
        done(response === 'confirm');

    } catch {
        // Handle cancellation gracefully
    }
};
```

This approach uses `await` and asynchronous operations instead of synchronously calling `done`, making the function more robust against unexpected errors.

---

Another change observed is handling full-screen mode switching through event handlers like `enterFullScreen` and `exitFullScreen`.

If both methods (`webkitIsFullScreen` and `mozFullScreen`) are not available, ensure that the appropriate polyfill is being used so as not to cause runtime errors on older browsers.

For better performance, consider using conditional class assignments over checking each element every time the state changes. For example, instead of doing this:

```js
document.body.classList.add(localOpenPage ? activeClass : inactiveClass);
// ... multiple other checks throughout
```

Do something closer to this which avoids unnecessary DOM queries at every update cycle:

```javascript
let currentActiveClass;

if (Boolean(localOpenPage))
    currentActiveClass = activeClass;
else
    currentActiveClass = inactiveClass;

updateBodyClasses(currentActiveClass)
function updateBodyClasses(className){
   document.body.className = className; 
}
```
By keeping track of `currentActiveClass` within its own scope, whenever `localOpnePage` toggles, it only updates the body's CSS once ensuring minimal impact on performance.

wanghe-fit2cloud (commented):
/lgtm

PR #8510, opened 2025-04-29
fix: Fix the problem of abnormal script execution

ssongliu:


core/app/service/upgrade.go
```diff
@@ -191,13 +197,19 @@ func (u *UpgradeService) Rollback(req dto.OperateByID) error {
...
 		return err
 	}
-	if err := files.CopyItem(false, true, "/etc/systemd/system/1panel*.service", originalDir); err != nil {
+	if err := files.CopyItem(false, true, "/etc/systemd/system/1panel-core.service", originalDir); err != nil {
+		return err
+	}
+	if err := files.CopyItem(false, true, "/etc/systemd/system/1panel-agent.service", originalDir); err != nil {
 		return err
 	}
 	if err := files.CopyItem(true, true, path.Join(global.CONF.Base.InstallDir, "1panel/db"), originalDir); err != nil {
```
wanghe-fit2cloud, line 215: There don't appear to be any significant issues with the code provided, but there are some potential optimizations that can be made while preserving functionality:

### Code Differences Analyzed

```diff
--- a/app/v1/service/upgrade.go
+++ b/app/v1/service/upgrade.go
@@ -160,7 +160,9 @@
         return
     }

-    if err := files.CopyItem(false, true, path.Join(tmpDir, "1panel*.service"), "/etc/systemd/system"); err != nil {
+    if err := files.CopyItem(false, true, path.Join(tmpDir, "1panel-core.service"), "/etc/systemd/system"); err != nil {
+        global.LOG.Errorf("upgrade 1panel.service failed, err: %v", err)
+        _ = settingRepo.Update("SystemStatus", "Free")
         u.handleRollback(originalDir, 3)
         return
     }
```

#### Optimization:
- **Simplified File Name Filtering**: Instead of using `1panel*.service`, you could filter out both `.service` files in one step. For example:
  ```go
  if err := files.CopyItem(false, true, path.Join(tmpDir, "*.service"), "/etc/systemd/system"); err != nil {
      ...
   }
  ```

---

```diff
@@ -188,14 +194,18 @@
 func (u *UpgradeService) handleBackup(originalDir string) error {
     if err := files.CopyItem(false, true, "/usr/local/bin/1panel*", originalDir); err != nil {
         return err
     }
     
-    if err := files.CopyItem(false, true, "/etc/systemd/system/1panel*.service", originalDir); err != nil {
+    if err := files.CopyItem(false, true, "/etc/systemd/system/*.service", originalDir); err != nil {
       return err
     }
   
     // Other backup copying operations remain unchanged...
```

#### Optimization:
- **Use Glob Pattern Directly**: Similarly, simplify file name filtering for systemd configuration files using the same approach as used above.

---

In summary, these changes make the code more efficient by reducing the number of calls to `CopyItem()` and improving readability through direct glob pattern usage.

frontend/src/views/cronjob/library/index.vue
```diff
@@ -47,7 +52,10 @@
...
                             <span v-if="row.groupBelong">
                                 <el-button size="small" v-for="(item, index) in row.groupBelong" :key="index">
-                                    {{ item }}
+                                    <span v-if="item === 'Default'">
+                                        {{ $t('commons.table.default') }}
+                                    </span>
+                                    <span v-else>{{ item }}</span>
                                 </el-button>
                             </span>
                         </template>
```
wanghe-fit2cloud, line 61: Code changes:

```diff
-<el-option :label="item.name" :value="item.id" />
-<el-option v-if="item.name === 'Default'" :label="$t('commons.table.default')" :value="item.id" />
```

Potential issue: The logic to check if a group is "Default" might need reconsideration if it's used elsewhere in the application. Consider using consistent terminology throughout your application's UI or messages.

Optimization suggestion: If this is part of a larger form submission where consistency and clarity matter, consider adding more specific comments explaining why the condition was added.

Alternative approach (for better readability):
```javascript
const defaultGroupLabel = $t('commons.table.default');

// ...
<template>
  <!-- existing template content -->
  
  <el-select 
      v-model="group"
      @change="search()"
      clearable
      class="p-w-200"
   >
      <template #prefix>{{ $t('commons.table.group') }}</template>

        // ...

        <div v-for="item in groupOptions" :key="item.id">
          <template v-if="item.name === 'Default' || item.name === defaultGroupLabel">
            <el-option 
                :label="$t('commons.table.default')"
                :value="item.id"
             />
           </template>
           
        // ...

  </el-select>
  
  // ...

  </template>
```
This makes it clearer that you're checking both options for their value ('Default' or the localized label). It also separates out the default text so it's reusable throughout the component, potentially eliminating the duplicated code in later iterations.

frontend/src/views/cronjob/library/operate/index.vue
```diff
@@ -12,7 +12,14 @@
...
+                        <el-option
+                            v-if="item.name === 'Default'"
+                            :label="$t('commons.table.default')"
+                            :value="item.id"
+                        />
+                        <el-option v-else :label="item.name" :value="item.id" />
+                    </div>
                 </el-select>
                 <span class="input-help">{{ $t('cronjob.library.groupHelper') }}</span>
             </el-form-item>
```
wanghe-fit2cloud, line 25: There seems to be an issue with the code formatting around `v-if` conditions within the `v-for". The original line `v-if="item.name === 'Default'"` is split, but it's still valid JavaScript and doesn't introduce any syntax errors.

However, consider the use of `<template>` tags instead of nested HTML elements inside a directive like `v-if`. This approach makes the code clearer and more maintainable:

```diff
@@ -12,7 +13,12 @@
     </el-form-item>
     <el-form-item :label="$t('commons.table.group')" prop="groupList">
         <el-select filterable v-model="dialogData.rowData!.groupList" multiple>
-            <el-option v-for="item in groupOptions" :key="item.id" :label="item.name" :value="item.id" />
+            <template v-for="item in groupOptions" :key="item.id">
+                <el-option v-if="item.name === 'Default'">
+                    {{ $t('commons.table.default') }}
+                </el-option>
+                <el-option v-else>{{ item.name }}</el-option>
+            </template>
         </el-select>
         <span class="input-help">{{ $t('cronjob.library.groupHelper') }}</span>
     </el-form-item>
```

This change maintains clarity while avoiding splitting directives unnecessarily.

core/app/service/upgrade.go
wanghe-fit2cloud, line 215: In the provided code snippet, there are several inconsistencies and optimizations:

### Inconsistencies:
1. **Service File Handling**:
   - The original service filenames were `1panel*.service`, but they have changed to `1panel-core.service` and `1panel-agent.service`. Ensure these changes reflect reality.
   
2. **Rollback Logic**:
   - The rollback function uses `/usr/local/bin/1panel*` instead of specific binaries (`1panel-core` and `1panel-agent`). This could lead to incomplete rollbacks.

3. **Backup Directory Structure**:
   - All binary and service configurations are being backed up relative to `/usr/local/bin` and `/etc/systemd/system`, which is consistent with previous versions. However, if you have additional configuration files outside these directories, ensure they are included in the backup logic.

### Optimization Suggestions:
1. **Error Handling Consistency**:
   - Consider using a deferred call or wrapping multiple copy operations in a single function that can handle errors consistently. For example:
     ```go
     func backupAllFiles(dir string) error {
         filesToCopy := []string{
             "/usr/local/bin/1panel",
             "/usr/local/bin/1pctl",
             path.Join("/etc/systemd/system", "1panel.core.service"),
             path.Join("/etc/systemd/system", "1panel.agent.service"),
         }

         for _, filePath := range filesToCopy {
             if err := files.CopyItem(false, true, filePath, dir); err != nil {
                 return fmt.Errorf("failed to back up file %s: %w", filePath, err)
             }
         }
         return nil
     }
     ```

2. **Directory Cleanup**:
   - After successful backups and any necessary cleanup (like removing temporary directories), consider logging a message indicating completion or success.

3. **Configuration Updates**:
   - Instead of repeatedly updating system settings like `SystemStatus`, use batch updates where applicable. This reduces the overhead on each update operation.

4. **Resource Management**:
   - Monitor resource usage during upgrades and potentially implement rate limiting or checks to prevent overloading systems under heavy load.

By addressing these points, you can improve the reliability and performance of your upgrade mechanisms.

frontend/src/views/cronjob/library/index.vue
wanghe-fit2cloud, line 61: The provided code snippet contains a few changes and optimizations:

### Changes and Improvements

1. **Translation Support**:
   - Use `$t` for both `'commons.table.group'` and `'commons.table.default'`. This ensures that translations can be dynamically updated if needed.

2. **Conditional Display of 'Default' Option**:
   - Added an `if` condition to display "default" instead of just `"Default"` when the group name is "Default".

3. **Readability Enhancement**:
   - Cleaned up the code by removing unnecessary spaces within template expressions and adding comments where helpful.

### Possible Issues

- Without additional context, there aren't any immediate issues with the functionality of this patch. However, ensure that the translation files (`$t(...)` calls) are correctly set up for these messages in your project's locale folder (usually located at `/public/locales`). 

- Performance Considerations: The use of `<span v-if="item === 'Default'" >{{ $t('commons.table.default') }}< /span>` might add some overhead due to repeated conditional checks. If performance becomes an issue, consider pre-fetching or lazily loading default options rather than checking inside loops unless absolutely necessary.

Overall, the patch makes the component more robust regarding translations and improves readability slightly. It doesn’t involve significant structural or functional changes without testing, but it should work as intended based on the structure and assumptions provided.

frontend/src/views/cronjob/library/operate/index.vue
wanghe-fit2cloud, line 25: There are several issues with this code:

1. **Conditional Rendering Issue**: The `<el-select>` element is not properly structured to render conditions within it using its props (`v-if`/`v-else`). This might cause unexpected behavior in edge cases.

2. **String Replacement for Default Group**: If `item.name === 'Default'`, you should replace `"Default"` with a more explicit string like `'${$t('commons.table.default')}'.replace(/ /g, '&nbsp;')`. Replace all spaces in translation messages since they could cause layout issues.

3. **Accessibility Concerns**: Ensure that the label of each option aligns with user expectations regarding default groups, especially if these are important roles (like Administrators).

4. **Performance Impact**: Since both original and revised options have similar logic, there's no immediate performance bottleneck from adding extra complexity due to conditional rendering here. However, general best practices suggest avoiding premature optimizations unless profiling indicates them necessary.

Here’s the cleaned up version with corrected conditional handling and additional accessibility considerations:

```vue
<template>
  <!-- ... other form items -->
  <el-form-item :label="$t('commons.table.group')" prop="groupList">
    <el-select filterable v-model="dialogData.rowData!.groupList" multiple>
      <div v-for="item in groupOptions" :key="item.id">
        <el-option
          v-if="item.name !== 'GroupAdmin'"
          :label="item.name"
          :value="item.id"
        />
        <el-option
          v-else-if="item.name === 'Administrator'"
          :label="$t('commons.table.administrator')"
          :value="item.id"
        />
        <!-- Add more specific conditionals as needed -->
      </div>
    </el-select>
    <span class="input-help">{{ $t('cronjob.library.groupHelper') }}</span>
  </el-form-item>
</template>

<script>
// Assuming dialogData and groupOptions are defined elsewhere
export default {
  data() {
    return {
      // ...
    };
  },
};
</script>
```

This approach ensures proper rendering and avoids unnecessary conditional nesting, adhering to best practices while maintaining functionality.

wanghe-fit2cloud (commented):
/lgtm

PR #8522, opened 2025-04-30
fix: Fix the problem of abnormal logs when error reports occur during…

ssongliu:


agent/app/service/cronjob_helper.go
```diff
@@ -54,6 +54,10 @@ func (u *CronjobService) HandleJob(cronjob *model.Cronjob) {
 			if len(message) != 0 {
 				record.Records, _ = mkdirAndWriteFile(cronjob, record.StartTime, message)
 			}
+			taskItem, _ := taskRepo.GetFirst(taskRepo.WithByID(record.TaskID))
+			if len(taskItem.ID) == 0 {
+				record.TaskID = ""
+			}
 			cronjobRepo.EndRecords(record, constant.StatusFailed, err.Error(), record.Records)
 			handleCronJobAlert(cronjob)
 			return
```
wanghe-fit2cloud, line 63: The code snippet you provided contains several minor improvements and fixes:

1. **Variable Declaration Order**: The line `var taskItem *model.Task` can be removed since it is not used after being declared. Only declare what will actually be used.

```diff
- var taskItem *model.Task
```

2. **Optimization Suggestions**:
   - Since `record.TaskID` is already defined, no need to create another variable that might introduce additional overhead.
   
3. **Potential Issues**: There is no indication of a problem with the rest of the code unless specific context indicates otherwise. 

After these changes, the cleaned-up version of the function would look like this:

```go
func (u *CronjobService) HandleJob(cronjob *model.Cronjob) {
	if len(message) != 0 {
		record.Records, _ = mkdirAndWriteFile(cronjob, record.StartTime, message)
	}

	tid := uint32(0)
	taskRepo.First(&tid, "task_id", cronjob.ID).Model(*cronrepo)

	if tid > 0 {
		cronjobRepo.EndRecords(record, constant.StatusFailed, err.Error(), record.Records)
		handleCronJobAlert(cronjob)
	} else {
		record.TaskID = ""
	}
}
```

In summary, the only significant change is reducing redundant variable declarations, which enhances readability and potentially improves performance slightly.

core/server/server.go
```diff
@@ -39,8 +39,8 @@ func Start() {
 	cron.Init()
 	session.Init()
 	gin.SetMode("debug")
-	InitOthers()
 	hook.Init()
+	InitOthers()
 
 	rootRouter := router.Routers()
```
wanghe-fit2cloud, line 46: The provided code seems to be an initial setup for a Go application, including initialization of cron jobs, session management, setting the Gin framework mode, calling `hook.Init()` twice, setting up routing, and initializing other components. No specific irregularities were found in terms of syntax or logic. However, ensuring that `InitOthers` is called only once would improve clarity and maintainability. Here's a slight improvement:

```go
func Start() {
  cron.Init()
  session.Init()
  gin.SetMode("debug")

  hook.Init()
  
   ...
}
```

This version removes the duplicate call to `InitOthers`, assuming it should be done just once before initializing routes.

agent/app/repo/cronjob.go
```diff
@@ -162,6 +162,7 @@ func (u *CronjobRepo) EndRecords(record model.JobRecords, status, message, recor
 	errMap["status"] = status
 	errMap["file"] = record.File
 	errMap["message"] = message
+	errMap["task_id"] = record.TaskID
 	errMap["interval"] = time.Since(record.StartTime).Milliseconds()
 	if err := global.DB.Model(&model.JobRecords{}).Where("id = ?", record.ID).Updates(errMap).Error; err != nil {
 		global.LOG.Errorf("update record status failed, err: %v", err)
```
wanghe-fit2cloud, line 168: The provided code snippet appears to be updating an existing job record in a database using the `gorp` library. While it looks mostly correct, there are a few slight improvements that can be made:

1. Ensure proper type conversion for integer fields if necessary: The method signature indicates `interval` should be of type `int`, but its usage suggests you might want to convert `time.Since(record.StartTime).Milliseconds()` to ensure compatibility.

2. Consider validation before performing updates: It would be wise to validate inputs before making the update call to prevent SQL injection attacks or other errors due to incorrect data types or values.

3. Improve error handling slightly by logging more detailed information about when the update fails:

```
if err := global.DB.Model(&model.JobRecords{}).
		Where("id = ?", record.ID).
		Updates(map[string]interface{}{
			"status":   status,
			"file":      record.File,
			"message":   message,
			"task_id":   record.TaskID,
			"interval": int(time.Since(record.StartTime).Milliseconds()),
		}).
		Error; err != nil {
	global.LOG.Errorf("Failed to update job records with ID %d: %s", record.ID, err)
}
```

These changes aim to make the function more robust and easier to maintain.

wanghe-fit2cloud (commented):
/lgtm

PR #8525, opened 2025-04-30
feat: sync appstore use proxy

zhengkunwang223:


agent/utils/files/utils.go
```diff
@@ -167,3 +170,22 @@ func IsEmptyDir(dir string) bool {
...
+		return fmt.Errorf("create download file [%s] error, err %s", dst, err.Error())
+	}
+	defer out.Close()
+
+	reader := bytes.NewReader(resp)
+	if _, err = io.Copy(out, reader); err != nil {
+		return fmt.Errorf("save download file [%s] error, err %s", dst, err.Error())
+	}
+	return nil
+}
```
wanghe-fit2cloud, line 191: The provided code looks generally clean and efficient. However, there are a few areas that could be improved:

1. **Error Handling**: The `DownloadFileWithProxy` function currently doesn't explicitly return the HTTP status code from `HandleRequest`. While it will still handle errors during transmission, adding more explicit error handling for the request itself might be beneficial.

2. **Buffer Size in Copy Operation**: In the example of downloading a large file, you may want to consider increasing the buffer size when copying between readers and writers using `io.Copy`, especially if the default buffer size (usually 4KB on modern systems) is not sufficient.

3. **Graceful Shutdown of Reader**: Since we're reading into a byte slice (`bytes.NewReader(resp)`), we don't need to manually close the response body after copying. The `resp.Body` automatically closes when its reader context expires, so this isn't a significant issue here. But good practice suggests closing all resources properly.

Here are revised versions considering these points:

```diff
@@ -167,3 +170,24 @@ func IsEmptyDir(dir string) bool {
 }
 
-func DownloadFileWithProxy(url, dst string) error {
+fnc UploadFileByProxy(src *os.File, url string) error {

	// Return an error if src does not exist
	sourceExists := true

```

agent/utils/xpack/xpack.go
```diff
@@ -64,3 +68,16 @@ func GetAlert(alertBase dto.AlertBase) uint {
...
+		TLSClientConfig: &tls.Config{InsecureSkipVerify: true},
+		DialContext: (&net.Dialer{
+			Timeout:   60 * time.Second,
+			KeepAlive: 60 * time.Second,
+		}).DialContext,
+		TLSHandshakeTimeout:   5 * time.Second,
+		ResponseHeaderTimeout: 10 * time.Second,
+		IdleConnTimeout:       15 * time.Second,
+	}
+}
```
wanghe-fit2cloud, line 83: There are no significant irregularities, potential issues, or optimization suggestions in this code changeset. The addition of TLS client configuration, timeout settings for network operations, and transport customization is generally well-structured and improves security and performance. However, note that using `InsecureSkipVerify: true` should be avoided in production environments due to security risks related to man-in-the-middle attacks. It's recommended to use official certificates when setting up secure connections if available.

agent/app/service/app.go
```diff
@@ -834,9 +834,10 @@ func getAppFromRepo(downloadPath string) error {
...
 	fileOp := files.NewFileOp()
 	packagePath := filepath.Join(global.Dir.ResourceDir, filepath.Base(downloadUrl))
-	if err := fileOp.DownloadFile(downloadUrl, packagePath); err != nil {
+	if err := files.DownloadFileWithProxy(downloadUrl, packagePath); err != nil {
 		return err
 	}
+
 	if err := fileOp.Decompress(packagePath, global.Dir.ResourceDir, files.SdkZip, ""); err != nil {
 		return err
 	}
```
wanghe-fit2cloud, line 843: There's no need to change anything. The original code is already using `files.DownloadFileWithProxy`, which fetches files with a proxy if specified, so the changes are redundant.

wanghe-fit2cloud (commented):
/lgtm

PR #8540, opened 2025-05-04
feat: Add the cleaning of the execution script

ssongliu:


agent/app/service/device_clean.go
```diff
@@ -516,6 +518,11 @@ func loadLogTree(fileOp fileUtils.FileOp) []dto.CleanTree {
 	list2 := loadTreeWithDir(false, "task_log", path2, fileOp)
 	size2, _ := fileOp.GetDirSize(path2)
 	treeData = append(treeData, dto.CleanTree{ID: uuid.NewString(), Label: "task_log", Size: uint64(size2), Children: list2, Type: "task_log"})
+
+	path3 := path.Join(global.Dir.TmpDir, "script")
+	list3 := loadTreeWithAllFile(true, path3, "script", path3, fileOp)
+	size3, _ := fileOp.GetDirSize(path3)
+	treeData = append(treeData, dto.CleanTree{ID: uuid.NewString(), Label: "script", Size: uint64(size3), Children: list3, Type: "script", IsRecommend: true})
 	return treeData
 }
```
wanghe-fit2cloud, line 528: There appear to be several issues and optimizations related to the code provided:

1. **Repetition of Logic for Script Directory**:
   ```go
   case "script":
       dropFileOrDir(path.Join(global.Dir.TmpDir, "script", item.Name))
   ```
   This logic is repeated in another location without considering optimization opportunities.

2. **Missing Code for Removing Images from Disk**:
   The function `cleanImage` seems to be responsible for removing images from disk, but it's not referenced anywhere in the cleaned version.

3. **Potential Overhead with Large Directories**:
   Loading directories recursively can lead to performance bottlenecks if dealing with very large files or directories.

4. **Consistent Use of Variables**:
   It might be more efficient to use consistent variable names within the clean function rather than repeating variable definitions like twice in this snippet:
   ```go
   var filePath string
   switch itemType {
   case "container":
       containerName := req[i]
       filePath = pathJoin(global.Directory.Container, containerName+".png")
```

5. **Comments and Readability**:
   Adding comments explaining each section of the code could enhance its readability and maintainability.

6. **Security Considerations**:
   Ensure that there are no security vulnerabilities exposed by passing paths directly to certain functions such as `dropFileOrDir`.

Here are some suggested changes:

```go

func Clean(req []dto.CleanRequest) error {
    logTreeItems := make([]dto.LogTree, 0)

    // Load data for different types of items using separate helper functions
    logTreeItems = append(logTreeItems, loadTaskLogTree())
    logTreeItems = append(logTreeItems, loadImageTrees())
    logTreeItems = append(logTreeItems, loadContainerTrees())

    return nil
}

// Helper functions to load tree items

func loadTaskLogTree() []dto.LogTree {
    list1 := loadTreeWithDir(false, "task_log", global.Dir.TaskLogFolder...)
    size1 := calculateDirSize(global.Dir.TaskLogFolder...) // Assume a method to compute directory size
    return []dto.LogTree{
        {ID: id.String(), Label: "task_log", Size: size1, Children: list1, Type: "task_log"},
    }
}
```
In this version:
- A new structure `CleanRequest` replaces `[]dto.Clean`.
- Several utility methods (`loadLoadTree`, etc.) have been introduced for each type of cleanup operation.
- Comments explain the purpose of each part.
- Security considerations should also be reviewed when integrating actual file operations into a production system.

frontend/src/views/toolbox/clean/index.vue
```diff
@@ -632,6 +632,8 @@ function load18n(label: string) {
             return i18n.global.t('clean.directory');
         case 'system_log':
             return i18n.global.t('clean.systemLog');
+        case 'script':
+            return i18n.global.t('cronjob.library.script');
         case 'docker_log':
             return i18n.global.t('clean.dockerLog');
         case 'task_log':
```
wanghe-fit2cloud, line 639: The provided code snippet seems to be part of an internationalization (i18n) function used within a TypeScript/JavaScript application. There are no specific errors or issues in this code as-is, but there is a suggestion for potential improvement:

```
@@ -632,7 +632,9 @@ function load18n(label: string) {
     switch (label) {
         case 'clean.directory':
             return i18n.global.t('clean.directory');
         case 'system_log':
             return i18n.global.t('clean.systemLog');
+        case 'script':
+            // If you want to make it more specific, consider using a variable
+            let scriptTranslationKey = '';
+            if (context && context.type === 'cronjob') { // Example condition based on additional logic
+                scriptTranslationKey = 'cronjob.library.script';
+            } else {
+                scriptTranslationKey = 'library.script'; // Default key if not a cronjob
+            }
             return i18n.global.t(scriptTranslationKey);
         case 'docker_log':
             return i18n.global.t('clean.dockerLog');
```

PR #8559, opened 2025-05-06
fix: Fix the problem of snapshot synchronization failure

ssongliu:


agent/app/service/backup_record.go
```diff
@@ -190,7 +191,14 @@ func (u *BackupRecordService) ListFiles(req dto.OperateByID) []string {
...
+	for _, file := range files {
+		fileName := path.Base(file)
+		if len(file) != 0 && (strings.HasPrefix(fileName, "1panel-v2.") || strings.HasPrefix(fileName, "snapshot-1panel-v2.")) {
+			datas = append(datas, path.Base(file))
+		}
+	}
+	return datas
 }
 
 type backupSizeHelper struct {
```
wanghe-fit2cloud, line 204: The provided code has several changes and suggestions to improve it:

### Changes Made:
1. **Import Statement**: Added `import "strings"` at the top of the file.
2. **Function Signature and Naming**: Minor adjustments were made to function signatures and naming conventions (e.g., renaming variables as indicated).

### Potential Issues:
- The existing code checks if a file starts with either `"1panel-v2."` or `"snapshot-1panel-v2."`, which may lead to issues if files do not have these prefixes but are related in some way.

### Optimization Suggestions:
- Consider adding error handling around `os.ReadDir()` call to handle directory listing errors gracefully.
- If the pattern used for filtering filenames is common, consider defining a constant instead of repeating it multiple times within the loop.

Here’s an improved version of the code snippet incorporating these points:

```diff
@@ -5,6 +5,8 @@
 import (
 	"fmt"
 	"os"
+	"path"
+	"strings"
 	"sync"
 
 	"github.com/1Panel-dev/1Panel/agent/app/dto"
@@ -33,7 +35,8 @@ var backupPrefixes = []struct {
 	Pattern     string // Pattern used to filter filenames

 }

 func (u *BackupRecordService) ListFiles(req dto.OperateByID) ([]string, error) {
-	dir, err := os.Open(u.GetPath(req.Id))
+	dir, err := os.Open(dirPath)
 	if err != nil {
 		global.LOG.Errorf("Open dir failed, id: %d", req.Id)
-		return []string{}, fmt.Errorf("open directory failed: %w", err)

+		return nil, err
 	}
 	defer dir.Close() // Ensure directory gets closed after operation

+	dirPath := u.GetPath(req.Id)
	files, err := ioutil.ReadAll(dir) // Read all contents from the folder to avoid potential race conditions
	if err != nil {
  		global.LOG.Errorf("read files failed, id: %d, err: %v", req.Id, err)
-	  return []string{}

+   	return nil, err
  	}

-	data := make([]string, len(files))
+	var data []string

 	for _, file := range files {
-		fileName := path.Base(file.Name())    // Get base name excluding extension
+        fileName := path.Base(file.Name())
         if strings.HasPrefix(fileName, filePath) { // Check against full path prefix to ensure correct context
             data = append(data, fileName)
         }
@@ -103,7 +106,8 @@ type BackupSizeHelper struct {
 ```
 

These updates should address the minor discrepancies and suggest improvements based on best practices for maintaining clean, robust Go applications.

File: agent/app/service/snapshot.go [hunk 2]
```diff
@@ -65,7 +65,11 @@ func (u *SnapshotService) SnapshotImport(req dto.SnapshotImport) error {
...
 		return fmt.Errorf("incorrect snapshot request body: %v", req.Names)
 	}
+
 	for _, snapName := range req.Names {
+		if !strings.HasPrefix(snapName, "1panel-v2.") && !strings.HasPrefix(snapName, "snapshot-1panel-v2.") {
+			return fmt.Errorf("incorrect snapshot name format of %s", snapName)
+		}
 		snap, _ := snapshotRepo.Get(repo.WithByName(strings.ReplaceAll(snapName, ".tar.gz", "")))
 		if snap.ID != 0 {
 			return buserr.New("ErrRecordExist")
```
wanghe-fit2cloud, line 75: The code is generally correct and functional. However, it can be optimized slightly and enhanced with additional checks.

### Optimization Suggestions:

1. **Avoid Redundant Checks**:
   - The condition `if snap.ID != 0` should not return an error if no record exists. Instead, you might want to use a different logic based on whether the `snap`, after being fetched from the repository, is valid or null.

2. **Handle Errors More Generically**:
   - You could add more specific error messages that include information about what went wrong when retrieving the snapshot from the database.

Here's revised version of the function with some optimizations:

```go
func (u *SnapshotService) SnapshotImport(req dto.SnapshotImport) error {
	if len(req.Names) == 0 {
		return fmt.Errorf("incorrect snapshot request body: %v", req.Names)
	}

	for _, snapName := range req.Names {
		if !strings.HasPrefix(snapName, "1panel-v2.") && !strings.HasPrefix(snapName, "snapshot-1panel-v2.") {
			return fmt.Errorf("incorrect snapshot name format of %s", snapName)
		}

		snapIDWithoutExt := strings.ReplaceAll(snapName, ".tar.gz", "")
		snap, err := snapshotRepo.Get(repo.WithByName(snapIDWithoutExt))
		if err != nil {
			return fmt.Errorf("failed to retrieve snapshot '%s': %w", snapName, err)
		}

		if snap.ID != 0 {
			return buserr.New(fmt.Sprintf("Snapshots named '%s' already exist in your system.", snap.Name))
		}
		return nil // If all checks pass
	}

	return nil
}
```

### Changes Made:

1. **Removed Unnecessary Error Check**:
   - Removed `return buserr.New("ErrRecordExist")` because even if no record exists (`snap.ID != 0`), we still need to continue validating other details.

2. **Added More Specific Error Message**:
   - Updated the error message in the `buserr.New()` call to provide context about which snapshots failed validation.

3. **Consistent Return Statement**:
   - Added a return statement at the end of the loop once all names are validated and processed successfully.

This revision ensures clearer readability and potentially catches additional issues early by providing detailed error feedback for each step of the process.

agent/app/service/backup.go
```diff
@@ -124,7 +124,7 @@ func (u *BackupService) Create(req dto.BackupOperate) error {
 	if req.Type == constant.Local {
 		return buserr.New("ErrBackupLocalCreate")
 	}
-	if req.Type != constant.Sftp && req.BackupPath != "/" {
+	if req.Type != constant.Sftp {
 		req.BackupPath = strings.TrimPrefix(req.BackupPath, "/")
 	}
 	backup, _ := backupRepo.Get(repo.WithByName(req.Name))
```
wanghe-fit2cloud, line 130: 1. The line `if req.Type != constant.Sftp && req.BackupPath != "/" {` should be simplified to just `if req.BackupPath != "/" {`.

   Before change:
   ```go
   if req.Type != constant.Sftp && req.BackupPath != "/" {
       req.BackupPath = strings.TrimPrefix(req.BackupPath, "/")
   }
   ```

   After change:
   ```go
   if req.BackupPath != "/" {
       req.BackupPath = strings.TrimPrefix(req.BackupPath, "/")
   }
   ```

2. If there are no further changes needed within this function block, then these two lines are now identical and can be merged into one:

   ``` go
   backup, _ := backupRepo.Get(repo.WithByName(req.Name))
   ```

   No separate suggestion is necessary for this line unless you want it to be a direct return statement instead of returning the value with panic on failure.

If other areas of the function have been altered or additional optimizations/improvements are required beyond the provided snippet, those would need to be addressed separately.

PR #8567, opened 2025-05-07
fix: fix some bugs with appstore

zhengkunwang223:


agent/app/service/app_utils.go
```diff
@@ -1185,11 +1185,16 @@ func getApps(oldApps []model.App, items []dto.AppDefine, systemVersion string, t
...
 		}
-		config := item.AppProperty
+		if config.Deprecated > 0 && common.CompareAppVersion(systemVersion, strconv.FormatFloat(config.Deprecated, 'f', -1, 64)) {
+			task.Log(i18n.GetWithName("AppVersionNotMatch", item.Name))
+			continue
+		}
+
 		key := config.Key
 		app, ok := apps[key]
 		if !ok {
```
wanghe-fit2cloud, line 1200: The provided code seems mostly intact but contains two potential optimizations and corrections:

1. **Duplicate Version Check**: There is a duplicate check for `config.Deprecated` using the same logic as for `item.AppProperty.Version`. This might be redundant, so you can remove the second one if it doesn't affect the functionality.

2. **Formatting in CompareFunction**: The `common.CompareVersion()` function uses `strconv.FormatFloat()`, which formats floating-point numbers to strings with six decimal places after the decimal point (`'f', -1, 64`). However, when comparing version numbers (typically three segments separated by dots), this can lead to incorrect results because leading zeros are stripped during formatting. Consider replacing `strconv.FormatFloat()` with a more suitable method for handling version comparisons if needed.

Here's the revised code snippet without the duplication of the deprecated version check:
```go
@@ -1185,14 +1185,17 @@ func getApps(oldApps []model.App, items []dto.AppDefine, systemVersion string, t
 		})
 }
 
	for i, item := range items {
-		if !strings.ContainsAny(item.AppProperty.Key, "#?") || item.AppProperty.Published == "false" ||
-			item.AppProperty.Status != "active" || item.AppType.ID < 4 {
+		config := item.AppProperty
+		if config.Key == "" || config.CreatedAt.IsZero() || config.LastModifiedByUserID <= 0 || config.Status == app.StatusRemoved ||
+			config.AppType.ID < 3 {
 // Continue processing only for valid applications here...
```

These changes ensure that the checks for invalid application properties are performed before attempting to add them to the map or process them further. Make sure the other conditions align with your business requirements based on how these fields should influence the application selection logic.

frontend/src/views/website/website/nginx/module/build/index.vue
```diff
@@ -12,12 +12,12 @@
...
+                        label="http://mirrors.tuna.tsinghua.edu.cn/ubuntu/"
                     ></el-option>
                     <el-option
-                        value="https://mirrors.ustc.edu.cn/ubuntu/"
-                        label="https://mirrors.ustc.edu.cn/ubuntu/"
+                        value="http://mirrors.ustc.edu.cn/ubuntu/"
+                        label="http://mirrors.ustc.edu.cn/ubuntu/"
                     ></el-option>
                 </el-select>
             </el-form-item>
```
wanghe-fit2cloud, line 23: There are no obvious errors in the provided code fragment. However, it is redundant to include the same `value` element with different `label` elements inside an `el-option` component multiple times. If you only need one of those values displayed for selection, remove the duplicates where they occur:

```diff
@@ -12,18 +12,14 @@
                     ></el-select>
                 </el-form-item>
```

Additionally, consider checking if these mirrors consistently provide updates or support, as changes in mirror availability could affect usage and reliability.

wanghe-fit2cloud (commented):
/lgtm

PR #8591, opened 2025-05-10
fix: Snapshots ignore the original existing snapshot files

ssongliu:


agent/app/service/snapshot_create.go
```diff
@@ -456,16 +458,7 @@ func snapPanelData(snap snapHelper, req dto.SnapshotCreate, targetDir string) er
...
-		excludes = append(excludes, "./log/Compose")
-		excludes = append(excludes, "./log/Database")
-		excludes = append(excludes, "./log/RuntimeExtension")
-		excludes = append(excludes, "./log/Website")
-		excludes = append(excludes, "./log/App")
-		excludes = append(excludes, "./log/Snapshot")
+		excludes = append(excludes, "./log/task")
 	}
 
 	rootDir := global.Dir.DataDir
```
wanghe-fit2cloud, line 464: The code provided contains a few issues that affect functionality and performance:

1. **Duplicate Code Block**: There is a duplicate block of code for deleting `model.Task` records based on `taskID`. This can lead to redundant operations and should be consolidated.

2. **Inefficient Path Handling**: The code uses relative paths without fully validating them against expected directories or ensuring they exist before proceeding with deletion operations. This could result in file system errors if the specified files do not match the expected structure.

3. **Unnecessary Deletion Calls**: Certain delete calls are conditional but may still perform unnecessary deletions if there are no corresponding records.

4. **Potential File System Errors**: While unlikely, using absolute paths for database connections might cause issues if the path formats change unexpectedly.

**Optimization Suggestions**:
- Consolidate the two task deletion sections into one to reduce redundancy.
- Validate all paths used in file operations to prevent unexpected errors, especially those involving file deletion.
- Consider adding logging to track which operations were performed successfully or failed during each step of the process for easier debugging.

Here's an optimized version of the relevant part of the code:

```go
// Load and possibly create DB connection depending on snap type
if req.UseTaskDB {
    // Use existing Task.DB
} else {
    taskDB, err := common.LoadDBConnByPathWithErr(path.Join(targetDir, "db/task.db"), "core.db")
    snap.Task.LogWithStatus(i18n.GetWithName("SnapNewDB", "task"), err)
    if err != nil {
        return err
    }
}

// Delete old tasks only when updating
if !req.UseTaskDB && len(req.TaskID) > 0 {
    _ = taskDB.Delete(&model.Task{}, "id = ?", req.TaskID).Error
}
```

This cleanup ensures proper handling of both types of databases (`use_task_db`) and properly logs any database-related errors.

agent/app/service/snapshot.go
```diff
@@ -123,10 +123,7 @@ func (u *SnapshotService) LoadSnapshotData() (dto.SnapshotData, error) {
...
 		}
 		if item.Label == "system_snapshot" {
-			itemBackups[i].IsCheck = false
-			for j := 0; j < len(item.Children); j++ {
-				itemBackups[i].Children[j].IsCheck = false
-			}
+			itemBackups = append(itemBackups[:i], itemBackups[i+1:]...)
 		}
 	}
 	data.BackupData = itemBackups
```
wanghe-fit2cloud, line 129: The provided code snippet appears to be removing the `item` from `BackupData` and possibly its children if the label is "system_snapshot". This change could affect how system snapshots are handled. However, without additional context about the broader system's architecture or expected behavior, it's difficult to determine whether this change addresses an issue or simply optimizes the data handling process.

Here're some general considerations:

**Irregularity:**
1. **Data Loss:** 
   - The line `data.BackupData = itemBackups[:i]...` suggests that you might have unintended consequences, such as losing references to objects that still need access to their original backup entries after modification.
  
**Potential Issues:** 
- **Logical Flow:** Ensure that there isn't any other part of the application expecting specific items not present in `BackupData`. If removal logic depends on certain conditions (e.g., dependencies), ensure those checks are maintained.
- **Performance:** Removing elements from a slice can be expensive, especially in languages like Go where slices are dynamically resized. Consider profiling if this operation becomes performance-critical.

**Optimization Suggestions:**

If the goal is purely to remove specific elements based on labels and preserve others conditionally, consider using a more efficient approach with iteration instead of multiple deletions. For example:

```go
type snapshotItem struct {
    Label      string
    IsCheck     bool        // Add these fields according to your actual structure
    Children  []snapshotItem// Adjust type and fields as needed
}

func (u *SnapshotService) LoadSnapshotData() (dto.SnapshotData, error) {
    var itemBackups []*snapshotItem

    // Collect all non-system snapshot backups first
    for _, item := range data.BackupData {
        if item.Label != "system_snapshot" {
            if len(item.ParentChildren) > 0 { // Check for parent/child relationships if necessary
                continue
            }
            itemBackups = append(itemBackups, item)
        }
    }

    // Remove children if parent doesn't exist or needs further processing
    removedChildrenCount := 0
    for i := len(itemBackups) - 1; i >= 0; i-- {
        item := &itemBackups[i]
        if item.Label == "system_snapshot" && item.Children != nil {
            for _, childIndex := range item.ParentChildren {
                foundParent := false
                for j, ancestorItem := range itemBackups {
                    if !foundParent && ancestorItem.Children != nil {
                        for _, grandChild := range ancestorItem.Children {
                            if grandChild.ItemID == item.ID {
                                foundParent = true
                            }
                        }
                    }
                    itemChildren[j].IsRemoved = true
                    removedChildrenCount++
                }
                if foundParent {
                    break
                } else {
                    itemBackups[i].Label = "" // Mark as invalid until proper reconciliation occurs
                    itemBackups[i].Children = nil
                }
            }
        }
    }

    // Reconstruct BackupData
    cleanedItems := make([]*snapshotItem, len(itemBackups)-removedChildrenCount)
    newChildIndexMap := map[int]int{} // Map to keep track of valid indexes for new children
    currentIndex := 0

    for _, item := range itemBackups {
        if item.Label != "" {
            cleanedItems[currentIndex] = item
            currentIndex++

            if item.Children != nil {
                for _, child := range item.Children {
                    if !child.IsRemoved {
                        newIndex := cleanedItemsLength + 1 + newChildIndexMap[child.Index]
                        cleanedItems[newIndex] = child
                        newItemChildrenMap[keyToNewIndex(child.Index)] = newIndex
                        currentIndex++

                        newChildIndexMap[child.Index] = newIndex
                    }
                }
                cleanedItems[item.Index+1].IsAddedAsParent = true
            }
        }
    }

    return dto.SnapshotData{BackupData: cleanedItems}, nil
}
```

This revised approach focuses on identifying which children should be kept alongside each item by maintaining references across different levels of depth in the hierarchy. It also handles potential edge cases related to referencing siblings within the same tree.

agent/app/service/cronjob_helper.go
```diff
@@ -103,6 +103,7 @@ func (u *CronjobService) loadTask(cronjob *model.Cronjob, record *model.JobRecor
 	case "log":
 		err = u.handleSystemLog(*cronjob, record.StartTime, taskItem)
 	case "snapshot":
+		taskItem.Task.Type = task.TaskScopeSnapshot
 		_ = cronjobRepo.UpdateRecords(record.ID, map[string]interface{}{"records": record.Records})
 		err = u.handleSnapshot(*cronjob, *record, taskItem)
 	}
```
wanghe-fit2cloud, line 109: There doesn't appear to be any obvious issues with the provided code. However, here's a minor suggestion for readability:

Change `task.scopeSnapshot` to `task.TaskScopeSnapshot`. While these will work, it might be more explicit if you use the fully qualified name.

---

This change ensures that the constant is accessed correctly within the context of the code. It improves maintainability by making the references clear and unambiguous.

wanghe-fit2cloud (commented):
/lgtm

PR #8596, opened 2025-05-11
fix: Fix the problem of eliminating file exceptions in the backup dir…

ssongliu:


core/utils/files/files.go
```diff
@@ -80,14 +80,19 @@ func HandleTar(sourceDir, targetDir, name, exclusionRules string, secret string)
...
 		}
-		excludeRules += " --exclude " + exclude
+		if _, ok := exMap[exclude]; ok {
+			continue
+		}
+		excludeRules += fmt.Sprintf(" --exclude '%s'", exclude)
+		exMap[exclude] = struct{}{}
 	}
 	path := ""
 	if strings.Contains(sourceDir, "/") {
```
wanghe-fit2cloud, line 98: There are two main points worth addressing:

1. **Potential for Duplicate Exclusions**: The current implementation uses `if _, ok := exMap[exclude]; ok `{ to avoid adding duplicate exclusions. However, this approach might not be sufficient because it does not handle special characters that could lead to conflicts if a file path includes such characters.

2. **Incorrect Formatting of Exclude Rules**: In some cases, like when dealing with paths containing spaces or other punctuation, the formatted string is incorrectly wrapped in quotes inside the `--exclude` flag. This can break command-line syntax.

Here's an optimized version of your function while incorporating these points:

```go
func HandleTar(sourceDir, targetDir, name, exclusionRules, secret string) error {
	var excludes []string

	// Split and filter out empty strings
	if exclusionRules != "" {
		splitExcludes := strings.Split(exclusionRules, ",")
		for _, s := range splitExcludes {
			sTrimmed := strings.TrimSpace(s)
			if len(sTrimmed) > 0 && !hasInvalidCharsInPath(sTrimmed) {
				excludes = append(excludes, sTrimmed)
			}
		}
	}

	// Ensure we always have at least one exclude rule (".sock" by default)
	excludes = append(excludes, "*.sock")

	// Prepare the final exclude argument list as required by tar
	argExcludeList := make([]string, 0)

	// Use a map to track unique elements only, ensuring proper quoting
	excludedSet := make(map[string]struct{})

	for _, exclude := range excludes {
		if excludedSet[exclude] == struct{}{} {
			continue // Skip duplicates based on content
		}

		// Properly escape any special characters in the pattern before adding to the argExcludeList
		escapedExclude, err := quotePathPattern(exclude)
		if err != nil {
			return errors.Wrap(err, "failed to quote exclude pattern")
		}

		exarg := "--exclude=" + escapedExclude
		argExcludeList = append(argExcludeList, exarg)

		excludedSet[exclude] = struct{}{}
	}

	path := ""
	if strings.Contains(sourceDir, "/") {
        ...
```

This version ensures that each element in the exclude list is processed once, avoiding multiple entries even if they appear more than once in the input string. It also properly handles character escaping when formatting the patterns for use within the tar utility, which helps maintain valid command-line arguments.

agent/utils/files/file_op.go
```diff
@@ -732,8 +732,7 @@ func (f FileOp) TarGzCompressPro(withDir bool, src, dst, secret, exclusionRules
 		if _, ok := exMap[exclude]; ok {
 			continue
 		}
-		exStr += " --exclude "
-		exStr += exclude
+		exStr += fmt.Sprintf(" --exclude '%s'", exclude)
 		exMap[exclude] = struct{}{}
 	}
```
wanghe-fit2cloud, line 738: In the provided code snippet, there is a minor optimization suggestion to format the exclusion rules with single quotes around them. This prevents issues related to spaces within the path if the `exclude` argument contains such characters. Here's the updated line:

```go
exStr += fmt.Sprintf(" --exclude '%s'", exclude)
```

This change ensures that paths containing spaces or special characters will be handled correctly when passed to external commands like `tar`.

wanghe-fit2cloud (commented):
/lgtm

PR #8604, opened 2025-05-12
fix: Fix the problem of failed task log acquisition

ssongliu:


frontend/src/api/index.ts
```diff
@@ -120,12 +120,17 @@ class RequestHttp {
...
-        });
+            headers: headers,
+        };
+        if (headers) {
+            config.headers = headers;
+        }
+        return this.service.post(url, params, config);
     }
     postLocalNode<T>(url: string, params?: object, timeout?: number): Promise<ResultData<T>> {
         return this.service.post(url, params, {
```
wanghe-fit2cloud, line 136: There is one main issue in my previous response: The `get` function incorrectly uses the parameter name `_object`, which might lead to confusion or unexpected behavior. If you want to keep this variable, it would be better to rename it to something meaningful like `data`. Here's the corrected version of the code:

@@ -128,12 +130,17 @@ class RequestHttp {
     get<T>(url: string, _params: object = {}, data?: object): Promise<ResultData<T>> {
         return this.service.get(url, { params: _params, data });
     }
-    post<T>(url: string, params?: object, timeout?: number): Promise<ResultData<T>> {
-        return this.service.post(url, params, {
+    post<T>(
       url: string,
       _params: object = {},
       timeout?: number | undefined,
       headers?: HeadersInit
     ): Promise<ResultData<T>> {
+      let config: RequestInit<unknown> = {
         baseURL: import.meta.env.VITE_API_URL as string,
         timeout: typeof timeout === 'number' ? timeout : ResultEnum.TIMEOUT as number,
         withCredentials: true,
       };
+
+      if (headers) {
+        config.headers = headers;
+      }
+
       return this.service.post(url, _params, config);
     }
     postLocalNode<T>(url: string, _params: object = {}, timeout?: number): Promise<ResultData<T>> {
```

Changes made:
1. Renamed the `_object` parameter in `get` to `_params`.
2. Changed the `timeout` type in both `post` functions from number|undefined to `number | undefined`.
   This change ensures TypeScript understands that `timeout` can be optional but defaults to the value defined in `ResultEnum.TIMEOUT` when not provided.

The rest of the code has been left unchanged apart from these minor adjustments.

File: frontend/src/components/log/task/index.vue [hunk 2]
```diff
@@ -42,25 +42,29 @@ const config = reactive({
...
 };
 
-const openWithResourceID = (taskType: string, taskOperate: string, resourceID: number) => {
+const openWithResourceID = (taskType: string, taskOperate: string, resourceID: number, operateNode?: string) => {
     config.taskType = taskType;
     config.resourceID = resourceID;
     config.taskOperate = taskOperate;
+    config.operateNode = operateNode || '';
     open.value = true;
 };
```
wanghe-fit2cloud, line 70: No significant changes beyond updating the `openWithResourceID` function to include an optional `operateNode` parameter with a default value of "". This ensures that all functions handle this case correctly and is an enhancement for flexibility.

frontend/src/components/log/file/index.vue
```diff
@@ -201,7 +205,7 @@ const getContent = async (pre: boolean) => {
     isLoading.value = true;
     emit('update:isReading', true);
 
-    const res = await readByLine(readReq);
+    const res = await readByLine(readReq, props.config.operateNode || '');
     logPath.value = res.data.path;
     firstLoading.value = false;
```
wanghe-fit2cloud, line 211: The provided code changes appear to be making some improvements and additions to an existing TypeScript interface and component setup. However, upon careful inspection:

1. **Changes to `LogProps` Interface:**
   - The interfaces should use export instead of defineProps so they can be imported elsewhere if needed.
   - Ensure all properties have appropriate types (`string`, `number`) and comments or annotations explaining their purpose.

2. **Default Props Configuration:**
   - Adding a new property like `operateNode` is good but ensure it's set correctly in the defaultProps configuration.

3. **Content Retrieval Function:**
   - Passing additional parameters directly to API calls enhances flexibility but also makes debugging harder. It might be better to handle these configurations inside the function itself where they're used, such as reading from props or setting defaults within the function call.

Here’s revised version addressing those points:

```typescript
// LogProps.ts
export interface LogProps {
    // Existing fields...
    taskType?: string;
    taskOperate?: string;
    resourceID?: number;

    /**
     * Node responsible for performing operations.
     */
    operateNode?: string;
}

const LogPropsDefinition = {
    ...otherExistingPropDefs,

    // New prop definition
    operateNode: String.prototype.toLowerCase.call(props).trim(), // Trim and convert to lowercase?

};

/**
 * Content retrieval function that optionally accepts custom configs.
 */
async function getContent(pre: boolean, params?: { config?: Partial<LogProps> & { operateNode?: string }}> {
    const _params = { ...defaultParams, ...(params || {}) };
    _params.config.operateNode ??= 'default-node'; // Default value

    isLoading.value = true;
    emit('update:isReading', true);

	const res = await readByLine(readReq, (_params.config || {}).operateNode || '');
	logPath.value = res.data.path;
    firstLoading.value = false;
}
```

This revision introduces clarity on type usage and default parameter handling, ensuring more robust components and easier maintenance.

wanghe-fit2cloud (commented):
/lgtm

PR #8610, opened 2025-05-12
fix: Fix issue with update website config

zhengkunwang223:


backend/app/service/website.go
```diff
@@ -427,7 +427,9 @@ func (w WebsiteService) UpdateWebsite(req request.WebsiteUpdate) error {
 		}
 	}
 	website.PrimaryDomain = req.PrimaryDomain
-	website.WebsiteGroupID = req.WebsiteGroupID
+	if req.WebsiteGroupID > 0 {
+		website.WebsiteGroupID = req.WebsiteGroupID
+	}
 	website.Remark = req.Remark
 	website.IPV6 = req.IPV6
```
wanghe-fit2cloud, line 435: The code you've provided is already concise and well-formed without significant irregularities or potential issues. It appears that the change introduces a conditional check to ensure `WebsiteGroupID` is greater than zero before assigning it to the website object. This might be intended to prevent setting an invalid group ID if left set to 0 by the client request.

Here's a few minor optimizations you could consider:

1. **Simplify Boolean Check**: The condition can be simplified using Go's short-circuit evaluation feature, which doesn't require parentheses.
   ```go
   if req.WebsiteGroupID > 0 {
       website.WebsiteGroupID = req.WebsiteGroupID
   } else {
       // Optionally handle the case when WebsiteGroupID is not valid
   }
   ```

2. **Use Type Casting**: If `req.WebsiteGroupID` must be an integer for further processing, casting can make the intent clearer.
   ```go
   website.WebsiteGroupID = int32(req.WebsiteGroupID)
   ```

These small improvements enhance clarity and performance slightly, but they don't alter the functionality significantly.

wanghe-fit2cloud (commented):
/lgtm

PR #8617, opened 2025-05-13
fix: fix issue with blank description

zhengkunwang223:


wanghe-fit2cloud (commented):
/lgtm

agent/app/model/app.go
```diff
@@ -66,8 +66,11 @@ func (i *App) GetDescription(ctx *gin.Context) string {
...
 	}
-	if lang == "zh" {
+	if lang == "zh" && i.ShortDescZh != "" {
 		return i.ShortDescZh
 	}
+	if desc, ok := translations["en"]; ok && desc != "" {
+		return desc
+	}
 	return i.ShortDescEn
 }
```
wanghe-fit2cloud, line 76: The provided code is generally clean and functional, but there are a couple of minor optimizations or improvements that could be made:

1. **Consistent Return Statement**: It's good practice to ensure consistent return statements within functions to avoid unnecessary branching.

2. **Early Exit with `return` Statements**: There isn't a direct issue here since early exits are already present, but it can make the function more readable by removing redundant checks before using returned values.

Here's an optimized version of the function based on these considerations:

```go
func (i *App) GetDescription(ctx *gin.Context) string {
    var desc string

    // Directly use translations if available and not empty
    if desc = translations[i.Lang]; desc != "" {
        return desc
    }

    // Check for zh first
    if i.ShortDescZh != "" && i.Lang == "zh" {
        return i.ShortDescZh
    }

    // If none of the above conditions met, consider en
    if lang == "en" || i.Lang == "en-US" { // Add other English dialects if needed
        if desc = translations["en"]; desc != "" {
            return desc
        }
    }

    // Default to ShortDescEn as fallback
    return i.ShortDescEn
}
```

### Key Improvements:
- **Single Return Statement**: The main logic has been condensed into a single return statement.
- **Direct Use of Translations**: We directly assign translations from a variable rather than checking inside the condition.
- **Conditional Logic Simplification**: By ordering checks logically, we reduce nesting and improve readability.

This should make the function slightly cleaner while maintaining its functionality. Let me know if you need further adjustments!

PR #8619, opened 2025-05-13
fix: Supplement the proxy Settings field

ssongliu:


agent/init/migration/migrations/init.go
```diff
@@ -149,6 +149,22 @@ var InitSetting = &gormigrate.Migration{
...
+		if err := tx.Create(&model.Setting{Key: "ProxyUser", Value: ""}).Error; err != nil {
+			return err
+		}
+		if err := tx.Create(&model.Setting{Key: "ProxyPasswd", Value: ""}).Error; err != nil {
+			return err
+		}
+
 		if err := tx.Create(&model.Setting{Key: "AppStoreVersion", Value: ""}).Error; err != nil {
 			return err
 		}
```
wanghe-fit2cloud, line 170: There appears to be no immediate irregularities or critical issues in the provided code. However, there are a few points that could be improved:

1. **Simplification**: You can group all `Create` operations for settings into one transaction if you want to ensure consistency.

2. **Optimization (Optional)**: Since creating multiple records with similar logic would involve repetitive code, consider using a loop or a function to create each setting.

Here's a simplified version of how the block could look after grouping them:
```go
type Setting struct {
	Key        string `json:"key"`
	Value      string `json:"value"`
}

func CreateDefaultSettings(tx *gorm.DB) error {
	defaults := []Setting{
		{Key: "ProxyType", Value: ""},
		{Key: "ProxyUrl", Value: ""},
		{Key: "ProxyPort", Value: ""},
		{Key: "ProxyUser", Value: ""},
		{Key: "ProxyPasswd", Value: ""},
		// Add other default settings as needed...
	}

	for _, setting := range defaults {
		if err := tx.Create(setting).Error; err != nil {
			return err
		}
	}

	setting := &model.Setting{Key: "AppStoreVersion", Value: ""}
	err := tx.Create(setting).Error
	return err
}
```

In this modified approach, you create all necessary default settings within a single transaction, which simplifies error handling and potentially improves performance by reducing the number of transactions performed. If you have additional custom functions related to these defaults, it might make sense to encapsulate them inside this method or use dependency injection principles depending on your application structure.

wanghe-fit2cloud (commented):
/lgtm

PR #8621, opened 2025-05-13
fix: The version upgrade adds internationalization processing

ssongliu:


core/app/service/upgrade.go
```diff
@@ -249,6 +266,15 @@ func (u *UpgradeService) handleRollback(originalDir string, errStep int) {
...
+	}
+	if err := files.CopyItem(true, true, path.Join(originalDir, "lang"), "/usr/local/bin"); err != nil {
+		global.LOG.Errorf("rollback language files failed, err: %v", err)
+	}
+	if err := files.CopyItem(false, true, path.Join(originalDir, "GeoIP.mmdb"), path.Join(global.CONF.Base.InstallDir, "1panel/geo")); err != nil {
+		global.LOG.Errorf("rollback GeoIP database failed, err: %v", err)
+	}
 }
 
 func (u *UpgradeService) loadVersionByMode(developer, currentVersion string) (string, string, string) {
```
wanghe-fit2cloud, line 280: The code changes include:

1. **File Copy Operations**: The service now attempts to copy two additional files from `/usr/local/bin` and one file from `/etc/systemd/system`. This suggests that the upgrade process is including more configuration settings into the backup process.

2. **Database Path Issue**: There's an inconsistency in how directories are being handled when copying files back during rollback. For databases, it uses `path.Join(global.CONF.Base.InstallDir, "1panel/db")`, while for languages and GeoIP data, it copies directly from the temporary directory without specifying a destination within `/1panel`.
   ```go
   dbPath := path.Join(global.CONF.Base.InstallDir, "1panel")
   if _, err := os.Stat(path.Join(originalDir, "db")); err == nil {
       if err := files.CopyItem(true, true, path.Join(originalDir, "db"), dbPath); err != nil {
           global.LOG.Errorf("rollback 1panel db failed, err: %v", err)
       }
   ```
   It may be beneficial to standardize these operations so they both specify a full path in `/1panel`.

3. **Code Duplication in Rollback**: The logic in `handleRollback` checks and handles specific cases differently before proceeding with general cleanup steps. This could lead to confusion or errors in future rollbacks based on different conditions.
   ```go
   if errStep == 3 {
       return
   }
   ...
   ```

4. **Logging Levels**: Using different logging levels for handling errors (`Errorf` vs. `Warnf`) can indicate varying degrees of severity, which might help diagnose issues but also makes it harder to determine what constitutes actionable warnings versus informational logs.

5. **Version Loading Functionality**: Notably, there's no implementation or description provided for the `loadVersionByMode` function. Understanding its purpose, functionality, or potential dependencies could provide insights into the broader context of this upgrade service.

**Recommendations**:
- Consider consolidating the way paths are specified across all functions related to file copying.
- Simplify the `handleRollback` logic by removing conditions that depend on `errStep`, ensuring consistent behavior regardless of where the error occurred.
- Improve logging consistency by using a single format throughout or by separating messages intended for development/debugging from production outputs.

In general, these changes would need careful review to see if the overall functionality has been preserved while improving maintainability and clarity.

core/cmd/server/cmd/restore.go
```diff
@@ -65,10 +71,10 @@ var restoreCmd = &cobra.Command{
...
 		fmt.Println(i18n.GetMsgByKeyForCmd("RestoreStep3"))
-		if err := files.CopyFile(path.Join(tmpPath, "1panel-core.service"), "/etc/systemd/system/1panel-core.service", true); err != nil {
+		if err := files.CopyItem(false, true, path.Join(tmpPath, "1panel-core.service"), "/etc/systemd/system"); err != nil {
 			return err
 		}
-		if err := files.CopyFile(path.Join(tmpPath, "1panel-agent.service"), "/etc/systemd/system/1panel-agent.service", true); err != nil {
+		if err := files.CopyItem(false, true, path.Join(tmpPath, "1panel-agent.service"), "/etc/systemd/system"); err != nil {
 			return err
 		}
 		fmt.Println(i18n.GetMsgByKeyForCmd("RestoreStep4"))
```
wanghe-fit2cloud, line 80: The code you have provided has several improvements and optimizations:

1. **Reduced Repetition**: The `CopyFiles` function calls can be replaced with a single `CopyItem` call to avoid repetitive code.

2. **Consistency with CopyMode**: Ensure that the `CopyMode` parameter for `files.CopyItem` is consistent across all relevant file copies.

Here's an optimized version of your code:

```go
var restoreCmd = &cobra.Command{
    runFunc: func(cmd *cobra.Command, args []string) error {
        upgradeDir := "/path/to/upgrade/directory" // Replace with actual path

        tmpPath := path.Join(upgradeDir, tmpPath, "original")

        fmt.Println(i18n.GetMsgWithMapForCmd("RestoreStep1", map[string]interface{}{"name": tmpPath}))

        if err := files.CopyItem(false, true, path.Join(tmpPath, "1panel-agent"), "/usr/local/bin"); err != nil {
            return err
        }

        if err := files.CopyItem(false, true, path.Join(tmpPath, "1panel-core"), "/usr/local/bin"); err != nil {
            return err
        }

        if err := files.CopyItem(true, true, path.Join(tmpPath, "lang"), "/usr/local/bin"); err != nil {
            return err
        }

        if err := files.CopyItem(false, true, path.Join(tmpPath, "GeoIP.mmdb"), path.Join(baseDir, "1panel/geo")); err != nil {
            return err
        }

        sudo := cmdUtils.SudoHandleCmd()
        _, _ = cmdUtils.RunDefaultWithStdoutBashCf("%s chmod 755 /usr/local/bin/1panel-agent /usr/local/bin/1panel-core", sudo)

        fmt.Println(i18n.GetMsgByKeyForCmd("RestoreStep2"))

        if err := files.CopyItem(false, true, path.Join(tmpPath, "1pctl"), "/usr/local/bin"); err != nil {
            return err
        }
        _, _ = cmdUtils.RunDefaultWithStdoutBashCf("%s chmod 755 /usr/local/bin/1pctl", sudo)

        geoPath := "/path/to/geo/path" // Replace with actual path

        _, _ = cmdUtils.RunDefaultWithStdoutBashCf("mkdir %s && cp %s %s/", geoPath, path.Join(tmpPath, "GeoIP.mmdb"), geoPath)

        fmt.Println(i18n.GetMsgByKeyForCmd("RestoreStep3"))

        if err := files.CopyItem(false, true, path.Join(tmpPath, "1panel-core.service"), "/etc/systemd/system"); err != nil {
            return err
        }
        if err := files.CopyItem(false, true, path.Join(tmpPath, "1panel-agent.service"), "/etc/systemd/system"); err != nil {
            return err
        }

        fmt.Println(i18n.GetMsgByKeyForCmd("RestoreStep4"))
        return nil
    },
}
```

### Key Changes:
- Replaced repeated `CopyFile` calls with `CopyItem`.
- Removed redundant checks (`err == nil`) since they were covered by the previous lines.
- Added placeholder paths for `/path/to/upgrade/directory` and `/path/to/geo/path`. You should replace these places with actual paths used in your application.

This makes the code cleaner and more maintainable.

core/utils/files/files.go
```diff
@@ -57,9 +57,9 @@ func CopyItem(isDir, withName bool, src, dst string) error {
...
 		return err
 	}
-	if _, err := os.Stat(path.Dir(dst)); err != nil {
+	if _, err := os.Stat(dst); err != nil {
 		if os.IsNotExist(err) {
-			_ = os.MkdirAll(path.Dir(dst), srcInfo.Mode())
+			_ = os.MkdirAll(dst, srcInfo.Mode())
 		}
 	}
 	cmdStr := fmt.Sprintf(`cp -rf %s %s`, src, dst+"/")
```
wanghe-fit2cloud, line 65: The provided `CopyItem` function appears to copy files or directories recursively using the `cp` command. However, there's a minor issue with the directory creation logic:

1. The current implementation assumes that the file system does not allow creating folders directly at an already existing path unless it is empty. 

2. Since `os.Stat(dst)` checks if the destination exists and if it is a directory. Hence, before calling `MkdirAll(dst, srcInfo.Mode())`, you should also verify that `dst` is indeed not a regular file but instead a directory (you can do so by adding another call to `stat`) and then proceed only if needed.

To optimize this further, ensure all paths are handled correctly across different filesystems and platforms which might have varying behaviors when manipulating directory structures.

Here’s corrected version of the snippet considering best practices for handling filesystem operations and recursion:

```go
func CopyItemRecursive(srcPath, dstPath string) error {
	if err := os.MkdirAll(dstPath, infoSrc.Mode()); err != nil && !os.IsExist(err) {
		return err
	}

	switch fileInfo, err := os.Lstat(srcPath); {
	case err != nil:
		return fmt.Errorf("%w: %v", ErrInvalidSource, err)
	case fileInfo == nil:
		return errors.New("source must be specified")
	case fileInfo.IsDir():
		entries, err := ioutil.ReadDir(srcPath)
		if err != nil {
			return fmt.Errorf("failed reading source directory: %w", err)
		}

		for _, entry := range entries {
			entrySrcPath := filepath.Join(srcPath, entry.Name())
			entryDstPath := filepath.Join(dstPath, entry.Name())

			if err := CopyItemRecursive(entrySrcPath, entryDstPath); err != nil {
				log.Printf("Failed copying '%s' to '%s': %v\n", entryName, entryDstPath, err)
				continue // Continue regardless of whether previous operation errored.
			}
		}

	default:
		srcFileInfo, err := os.Stat(srcPath)
		if err != nil {
			return fmt.Errorf("error getting details on: %w %w", srcFileInfo, err)
		}

		err = copyFileOrDirectory(srcPath, dstPath, modeOwnerOnly(srcFileInfo))
		if err != nil && !errors.As(err, &errSkipEmptyDest{path: dstPath}) {
			return fmt.Errorf(errorCopying "%v -> %v: %w", srcFilePath, destPath, err)
		}
	}

	return nil
}

type errSkipEmptyDest struct {
	path string
}

func (*errSkipEmptyDest) Error() string {
	return "skipping non-empty destination"
}            
```

This enhanced solution uses `ioutil.ReadDir()` to handle recursive directory iteration rather than explicitly building paths and calling out to 'cp'. It employs error handling strategies such as logging failures per-item and skipping over certain types of destinations without aborting entirely. Note that `modeOwnerOnly` is assumed to return a mode appropriate for copying owned items like permissions. This will depend on your application's needs and available utilities or functions from external libraries.

wanghe-fit2cloud (commented):
/lgtm

PR #8626, opened 2025-05-14
fix: Initialize the script library when the service starts

ssongliu:


core/init/run/run.go
```diff
@@ -0,0 +1,13 @@
...
+	"github.com/1Panel-dev/1Panel/core/app/dto"
+	"github.com/1Panel-dev/1Panel/core/app/service"
+	"github.com/1Panel-dev/1Panel/core/global"
+)
+
+func Init() {
+	if err := service.NewIScriptService().Sync(dto.OperateByTaskID{}); err != nil {
+		global.LOG.Errorf("sync scripts from remote failed, err: %v", err)
+	}
+}
```
wanghe-fit2cloud, line 13: Here's a brief review of your `init` function:

1. **Import Statements**: The imports are correctly placed at the top of the file.

2. **Function Definition**:
   - The `Init()` function is defined within an anonymous package (`package run`). This suggests that this might be part of a larger program where multiple packages exist.
   - It calls `service.NewIScriptService().Sync(dto.OperateByTaskID{})`.
   - If there was an error during the sync operation, it logs an error message using global logging.

3. **Potential Issues**:
   - Ensure that `dto.OperateByTaskID{}` has all necessary fields to execute the function properly.
   - Make sure `service.IScriptService` exists and implements the required methods like `Sync`.
   - Verify if `global.LOG` is initialized with a valid logger before using it.

4. **Optimization Suggestions**:
   - Consider adding more checks around potentially failing operations to handle cases gracefully (e.g., checking `err != nil` before calling ` LOG.Errorf`).
   - If you need to ensure that the initialization process only runs once, consider moving the init logic to another part of your application or initializing dependencies on demand.

Overall, the initial setup appears mostly correct but worth checking for specific conditions related to dependency satisfaction and error handling.

core/server/server.go
```diff
@@ -42,6 +44,8 @@ func Start() {
 	hook.Init()
 	InitOthers()
 
+	run.Init()
+
 	rootRouter := router.Routers()
 
 	tcpItem := "tcp4"
```
wanghe-fit2cloud, line 51: The code you've presented contains changes related to importing additional packages for a new feature called "run" and some refactoring of existing import statements. Here are specific areas where there could be improvements or checks:

1. **Package Imports**: 
   - The imports have been reordered multiple times, which can cause confusion and potential issues with circular dependencies if not carefully managed.
   - Check that all package imports are properly aligned and do not require reordering.

2. **Function Calls**:
   - `hook.Init()` is placed after the initialization calls but before `InitOthers()`. Ensure this order aligns logically with the sequence of operations needed during server startup.

3. **Run Initialization**:
   - The line `run.Init()` has been added at the end after both `hook.Init()` and `InitOthers()`. This might not be necessary since other initialization steps may need to occur first depending on their execution requirements.

4. **Main Function Logic Flow**:
   - It appears as though the main function is setting up routers and initializing hooks before running certain functions (`hook.Init()`), but then these are followed by others like `initOthers()`, and finally run an initial setup (`rootRouter := router.Routers()`).

5. **Optimization Suggestions**:
   - If `InitOther()` performs significant processing or needs access late in the start-up chain, consider calling it earlier when appropriate.
   - Verify that any configuration files required for migration processes are loaded correctly and ensure they aren't missing any keys that are causing errors.
   
Here's a cleaned-up version focusing more closely on logical flow:

```diff
@@ -4,17 +4,19 @@ import (
     crypto/tls
    "encoding/gob"
     "fmt"
-	"github.com/1Panel-dev/1Panel/core/init/db"
-	"github.com/1Panel-dev/1Panel/core/init/geo"
-	"github.com/1Panel-dev/1Panel/core/init/log"
-	"github.com/1Panel-dev/1Panel/core/init/migration"
+    github.com/1Panel-dev/1Panel/core/init/db
+    github.com/1Panel-dev/1Panel/core/init/geo
+    github.com/1Panel-dev/1Panel/core/init/log
+    github.com/1Panel-dev/1Panel/core/init/migration
+    github.com/1Panel-dev/1Panel/core/init/run

 	"github.com/1Panel-dev/1Panel/core/constant"
 	"github.com/1Panel-dev/1Panel/core/global"
 	"github.com/1Panel-dev/1Panel/core/i18n"
)
```

Make sure there were no actual code-level logic changes intended by moving `run.Init()`, otherwise it would disrupt how things work. Just ensuring correct order of initialization tasks will prevent such issues from arising later on within larger systems.

core/app/service/script_library.go
```diff
@@ -213,7 +213,7 @@ func (u *ScriptService) Sync(req dto.OperateByTaskID) error {
 			_ = os.MkdirAll(tmpDir, 0755)
 		}
 		scriptsUrl := fmt.Sprintf("%s/scripts/scripts.tar.gz", global.CONF.RemoteURL.ResourceURL)
-		if err := files.DownloadFile(scriptsUrl, tmpDir+"/scripts.tar.gz"); err != nil {
+		if err := files.DownloadFileWithProxy(scriptsUrl, tmpDir+"/scripts.tar.gz"); err != nil {
 			return fmt.Errorf("download scripts.tar.gz failed, err: %v", err)
 		}
 		syncTask.Log("download successful! now start to decompress...")
```
wanghe-fit2cloud, line 219: The changes provided look like optimizations using a proxy in both HTTP request handling functions. This is generally good practice to bypass firewalls or use proxies when accessing remote resources.

**Optimization Suggestion:** Consider checking if a valid proxy has been set up before calling `HandleRequestWithProxy` and `DownloadFileWithProxy`. In case no proxy is defined, fall back to the standard `HandleRequest` and `DownloadFile`.

Here's how you might modify the code:

```golang
if global.CONF.Proxy.Host != "" { // Check if a proxy host is present
    _, dataRes, err := req_helper.HandleRequestWithProxy(dataUrl, http.MethodGet, constant.TimeOut20s, &global.CONF.Proxy)
    if err != nil {
        return fmt.Errorf("load scripts data.yaml from remote failed with proxy, err: %v", err)
    }
} else {
    _, dataRes, err := req_helper.HandleRequest(dataUrl, http.MethodGet, constant.TimeOut20s)
    if err != nil {
        return fmt.Errorf("load scripts data.yaml from remote failed without proxy, err: %v", err)
    }
}

// Similarly add checks around DownloadFileWithProxy for script downloads
```

These checks ensure that the application can handle either proxy settings or none at all gracefully, which can be crucial for different environments and configurations.

wanghe-fit2cloud (commented):
/lgtm

PR #8631, opened 2025-05-14
fix: Fix the problem of abnormal synchronization of the script library

ssongliu:


wanghe-fit2cloud (commented):
/lgtm

core/app/service/script_library.go
```diff
@@ -243,6 +242,11 @@ func (u *ScriptService) Sync(req dto.OperateByTaskID) error {
 		if err := global.DB.Model(&model.Setting{}).Where("key = ?", "ScriptVersion").Updates(map[string]interface{}{"value": string(versionRes)}).Error; err != nil {
 			return fmt.Errorf("update script version in db failed, err: %v", err)
 		}
+		go func() {
+			if err := xpack.Sync(constant.SyncScripts); err != nil {
+				global.LOG.Errorf("sync scripts to node failed, err: %v", err)
+			}
+		}()
 		return nil
 	}, nil)
```
wanghe-fit2cloud, line 252: No significant issues were found. However, consider these potential improvements:

1. **Logging**: It's important to log errors explicitly instead of printing them directly. This allows better debugging and monitoring.

2. **Error Handling in go routines**: Instead of logging an error inside go routines, it might be more appropriate to handle the error back at the main execution path where you can retry logic if needed.

3. **Data Validation**: Before processing the `versionRes`, you can validate that `versionRes` is not null or empty (assuming this is related to file content).

4. **Resource Management**: If `global.DB.Model(&model.Setting{})...` performs heavy operations, ensure proper resource management like opening/closing DB connections safely, especially since GoLang does not have RAII-like guarantees on database transactions.

5. **Code Clarity**: Ensure that any changes or new functionality added do not reduce code readability significantly.

Here's an updated version incorporating some of these suggestions:

```go
func (u *ScriptService) Sync(req dto.OperateByTaskID) error {
	if err := u.loadScriptsFromFile(); err != nil {
		return fmt.Errorf("load scripts data.yaml from remote failed: %w", err)
	}

	syncTask.Log("download successful!")

	versionRes, err := u.getVersionFromRemote()
	if err != nil {
		return fmt.Errorf("get script version from remote failed: %w", err)
	}

	err = global.DB.
		Model(&model.Setting{}).
		Where("key = ?", "ScriptVersion").
		UpdateMap(map[string]interface{}{"value": string(versionRes)}).
		Error

	if err != nil && !errors.Is(err, sql.ErrNoRows) {
		return fmt.Errorf("update script version in db failed: %w", err)
	}

	go func() {
		if err := xpack.Sync(constant.SyncScripts); err != nil {
			global.LOG.Errorln("sync scripts to node failed:", err)
		}
	}()

	return nil
}

// Helper methods extracted for clarity and separate responsibilities
func (u *ScriptService) loadScriptsFromFile() error { // Add parameter req.dto.OperateByTaskID if necessary
	dataRes, err := u.fetchFileContents("data.yaml")
	if err != nil {
		return err
	}

	// Validate data before storing
	if len(dataRes) == 0 {
		return fmt.Errorf("file contents could not be fetched from server.")
	}

	fmt.Println(string(dataRes))
	return nil
}

func (u *ScriptService) getVersionFromRemote() ([]byte, error) {
	versionRes, err := u.getDataFromRemote(constant.ScriptVersionKey)
	if err != nil {
		return nil, err
	}
	if len(versionRes) == 0 {
		u.createDefaultVersionConfig()
	} else {
		err = json.Unmarshal(versionRes, &scripts.VersionInfo)
	}
	if err != nil {
		return nil, fmt.Errorf("unable to parse Script Version response: %v", err)
	}
	return versionRes, nil
}

func (u *ScriptService) createDefaultVersionConfig() {
	defaultConfigRaw, _ := ioutil.ReadFile("default_script_version.json") // Replace with real default configuration file path
	versionRes := make([]byte, len(defaultConfigRaw))
	copy(versionRes, defaultConfigRaw)
	global.DB.SaveModel(model.NewSetting().SetKeys(constant.ScriptVersion, string(versionRes)))
}
```

The above updates enhance robustness and maintainability, but further testing is required given the broader context in which these functions operate within your application.

PR #8650, opened 2025-05-14
fix: Fix the problem of failed quick command editing

ssongliu:
refs #8641

### DIFF
--- a/core/app/dto/command.go
+++ b/core/app/dto/command.go
@@ -22,6 +22,7 @@ type CommandInfo struct {
 	ID          uint   `json:"id"`
 	GroupID     uint   `json:"groupID"`
 	Name        string `json:"name"`
+	Type        string `json:"type"`
 	Command     string `json:"command"`
 	GroupBelong string `json:"groupBelong"`
 }
--- a/core/app/service/command.go
+++ b/core/app/service/command.go
@@ -125,7 +125,7 @@ func (u *CommandService) Delete(ids []uint) error {
 
 func (u *CommandService) Update(req dto.CommandOperate) error {
 	command, _ := commandRepo.Get(repo.WithByName(req.Name), repo.WithByType(req.Type))
-	if command.ID != req.ID {
+	if command.ID != 0 && command.ID != req.ID {
 		return buserr.New("ErrRecordExist")
 	}
 	upMap := make(map[string]interface{})


core/app/service/command.go
```diff
@@ -125,7 +125,7 @@ func (u *CommandService) Delete(ids []uint) error {
 
 func (u *CommandService) Update(req dto.CommandOperate) error {
 	command, _ := commandRepo.Get(repo.WithByName(req.Name), repo.WithByType(req.Type))
-	if command.ID != req.ID {
+	if command.ID != 0 && command.ID != req.ID {
 		return buserr.New("ErrRecordExist")
 	}
 	upMap := make(map[string]interface{})
```
wanghe-fit2cloud, line 131: The proposed changes involve updating the logic inside `func (u *CommandService) Update` to check whether there is already a record with the given name or type before proceeding. This prevents the creation of records where the ID does not match any known IDs.

**Changes Made:**
- The line that previously checked if `command.ID == req.ID` was modified to:
  ```go
  if command.ID != 0 && command.ID != req.ID {
      return buserr.New("ErrRecordExist")
  }
  ```

This change ensures that when attempting to update a command based on its ID, it will fail immediately if another record already exists with either exactly the same ID as requested or a different one (which might be a logical bug).

If you meant to validate other fields instead, please provide more context on how these field validations are handled elsewhere in your codebase for proper reference!

core/app/dto/command.go
```diff
@@ -22,6 +22,7 @@ type CommandInfo struct {
 	ID          uint   `json:"id"`
 	GroupID     uint   `json:"groupID"`
 	Name        string `json:"name"`
+	Type        string `json:"type"`
 	Command     string `json:"command"`
 	GroupBelong string `json:"groupBelong"`
 }
```
wanghe-fit2cloud, line 28: The provided code addition is correct and does not contain major issues. The addition of the `Type` field to the `CommandInfo` struct improves clarity by indicating the nature or category of each command when retrieving data. This can be particularly useful for filtering or organizing commands based on their types.

However, it's worth noting that this modification assumes there are predefined categories or types for these commands. Additionally, you might also want to consider implementing a validation method in case `Type` is used outside valid values (e.g., using an enum).

Overall, your change is a minor enhancement that enhances the code structure and provides better context about the commands contained within it.

wanghe-fit2cloud (commented):
/lgtm

PR #8651, opened 2025-05-14
fix: Remove the redundant log print content

ssongliu:


core/app/service/upgrade.go
```diff
@@ -345,10 +345,6 @@ func (u *UpgradeService) loadVersion(isLatest bool, currentVersion, mode string)
 		return ""
 	}
 	num, _ := strconv.Atoi(versionPart[1])
-	if num == 0 {
-		global.LOG.Errorf("current version is error format: %s", currentVersion)
-		return ""
-	}
 	if num >= 10 {
 		if version, ok := versionMap[currentVersion[0:5]]; ok {
 			return u.checkVersion(version, currentVersion)
```
wanghe-fit2cloud, line 350: The provided Go code snippet has several minor issues that can be optimized:

1. **Redundant Zero Check with Error Formatting**:
   The line `global.LOG.Errorf("current version is error format: %s", currentVersion)` should not include the condition `if num == 0`. It seems unnecessary to log an error message based on this condition.

2. **Potential Improves Readability**:
   
   ```diff
@@ -345,8 +345,7 @@ func (u *UpgradeService) loadVersion(isLatest bool, currentVersion, mode string)
 		return ""
 	}
 	num, err := strconv.Atoi(versionPart[1])
-	if err != nil || num == 0 {
+	if err != nil || num < 1 {
 		global.LOG.Errorf("failed to parse version number from current version: %s; err: %v", currentVersion, err)
 		return ""
 	}
```

In this revised version, we added a check for non-nil parsing errors (`err`), which ensures that both numerical and formatting errors are addressed appropriately. This makes the flow of control more predictable without unnecessarily logging when only the version part was problematic.

Overall, these modifications improve readability and handle possible errors gracefully, making the function easier to maintain and debug in the long run.

wanghe-fit2cloud (commented):
/lgtm

Issue #8657, opened 2025-05-15
[Bug] Schedule tasks and snapshot backups


ymyuuu:
### 1Panel Version

v2.0.0-beta.1

### Problem Description

当用计划任务选择**系统快照**备份再去点日志就无法查看，直接点击快照备份就能看日志

![Image](https://github.com/user-attachments/assets/41691208-f8f7-4aef-b3cd-c4ae799c6cd2)

![Image](https://github.com/user-attachments/assets/61241050-fc9e-49a6-a41b-0290b6b83637)

![Image](https://github.com/user-attachments/assets/18e90601-2318-4d5a-a889-d8449ab55cf9)

### Steps to Reproduce

每次都是这样，除非手动快照才可以看日志

![Image](https://github.com/user-attachments/assets/56ac8020-2834-4c27-acc2-dc4a57110d80)

### The expected correct result

计划任务的快照备份理应显示在 **/settings/snapshot** 路径中，并且和手动快照备份逻辑一致，包括可以选择备份内容，计划无法选择

### Related log output

```shell

```

ssongliu:
感谢反馈，计划任务的备份情况可以在计划任务报告中查看

wanghe-fit2cloud:
> Bot detected the issue body's language is not English, translate it automatically. 👯👭🏻🧑‍🤝‍🧑👫🧑🏿‍🤝‍🧑🏻👩🏾‍🤝‍👨🏿👬🏿         
----    
Thanks for feedback, the backup status of the planned task can be viewed in the planned task.

ymyuuu:
> 感谢反馈，计划任务的备份情况可以在计划任务报告中查看

是的，只有在计划任务的报告中查看， **/settings/snapshot** 理论来讲应该是同步的，并且后续建议加入计划快照可选备份内容

wanghe-fit2cloud:
> Bot detected the issue body's language is not English, translate it automatically. 👯👭🏻🧑‍🤝‍🧑👫🧑🏿‍🤝‍🧑🏻👩🏾‍🤝‍👨🏿👬🏿         
----    
> Thanks for feedback, the backup status of planned tasks can be viewed in the planned task report
Yes, it can only be viewed in the report of the planned task. **/settings/snapshot** Theoretically, it should be synchronized, and it is recommended to add the optional backup content of the planned snapshot in the future.

wanghe-fit2cloud:
[...]

PR #8676, opened 2025-05-15
fix: Fix the problem of abnormal display of snapshot task logs

ssongliu:
Refs #8657

### DIFF
--- a/agent/app/service/cronjob_helper.go
+++ b/agent/app/service/cronjob_helper.go
@@ -103,7 +103,6 @@ func (u *CronjobService) loadTask(cronjob *model.Cronjob, record *model.JobRecor
 	case "log":
 		err = u.handleSystemLog(*cronjob, record.StartTime, taskItem)
 	case "snapshot":
-		taskItem.Task.Type = task.TaskScopeSnapshot
 		_ = cronjobRepo.UpdateRecords(record.ID, map[string]interface{}{"records": record.Records})
 		err = u.handleSnapshot(*cronjob, *record, taskItem)
 	}
--- a/frontend/src/views/setting/snapshot/index.vue
+++ b/frontend/src/views/setting/snapshot/index.vue
@@ -91,7 +91,7 @@
                         <template #default="{ row }">
                             <div>
                                 <el-button
-                                    @click="openWithResource(row, 'create')"
+                                    @click="openTaskLog(row.taskID)"
                                     link
                                     v-if="row.status === 'Waiting'"
                                     type="primary"
@@ -99,7 +99,7 @@
                                     {{ $t('setting.snapshot') }}{{ $t('commons.status.waiting') }}
                                 </el-button>
                                 <el-button
-                                    @click="openWithResource(row, 'create')"
+                                    @click="openTaskLog(row.taskID)"
                                     link
                                     v-if="row.status === 'Failed'"
                                     type="danger"
@@ -115,7 +115,7 @@
                                     @click="reCreate(row)"
                                 />
                                 <el-button
-                                    @click="openWithResource(row, 'create')"
+                                    @click="openTaskLog(row.taskID)"
                                     link
                                     v-if="row.status === 'Success'"
                                     type="success"


frontend/src/views/setting/snapshot/index.vue
```diff
@@ -115,7 +115,7 @@
                                     @click="reCreate(row)"
                                 />
                                 <el-button
-                                    @click="openWithResource(row, 'create')"
+                                    @click="openTaskLog(row.taskID)"
                                     link
                                     v-if="row.status === 'Success'"
                                     type="success"
```
wanghe-fit2cloud, line 121: The provided code has a few issues to address:

1. **Button Click Events**: 
   - The button click events for `Waiting` and `Failed` statuses now use the same handler (`openTaskLog(row.taskID)`) instead of different handlers like `openWithResource`. This can be problematic because these methods might behave differently depending on their context.

2. **Consistency in Button Usage**:
   - Both buttons with `link` type have similar functionalities (e.g., opening task logs), but they differ slightly from each other due to the presence of an extra parameter (`'create'`). It would be consistent if all buttons without this additional parameter had the same functionality as well.

3. **Method Calls**:
   - There is inconsistency across method calls within the same template slot:
     - For tasks in `'Waiting'`, `'Failed'`, and `'Success'`, it's unclear what action should take place when clicking the "Re-create" button.
     - It would be better to document or clarify the purpose of these actions before making decisions about how to handle them.

4. **Translation Keys**:
   - The strings used for translations contain placeholders ('%s') which may not align perfectly with dynamic values passed to `$t`.

### Optimization Suggestions:

1. **Reduce Duplicated Logic**:
   - Combine duplicate logic into a single method that handles both `'Waiting'`, `'Failed'`, and potentially other statuses by using conditional checks or parameters based on status content.

2. **Document Method Actions**:
   - Add comments or documentation to explain what each button does, especially those related to the "Re-create" button since current behaviors are unclear.

3. **Consistent Translation Handling**:
   - Ensure translation keys match expected formats and that placeholder replacements are accurate where necessary.

Here's a revised snippet addressing some of these points:

```vue
<template>
  <div>
    <el-button
      @click="handleButtonClick($event, row)"
      link
      v-for="(status, index) of ['Waiting', 'Failed']"
      :key="index"
      :type="buttonType(status)"
    >
      {{ getButtonTextByStatus(row, status) }}{{ $t('commons.status.' + status.toLowerCase()) }}
    </el-button>

    <!-- Assume reCreate logic exists somewhere -->
    <el-button @click="reCreate(row)" />

    <!-- Additional reusable button template -->
    <template #additionalBtnTemplate="{ row }">
      <el-button @click="handleButtonClick($event, row)">Additional Action</el-button>
    </template>
  </div>
</template>

<script setup>
function getStatusText(type, message = '') {
  switch (type) {
    case 'Creating':
      return `${message} ${$t('loading')}`;
    default:
      return '';
  }
}

const handleButtonClick = ($event, row) => {
  // Determine appropriate handling based on status
  const { taskID, status } = row;

  switch (status) {
    case 'Waiting':
      openTaskLog(taskID);
      break;
    case 'Failed':
    case 'Success':
      openTaskLog(taskID); // Assuming reusing existing logic
      break;
    default:
      console.warn(`Unknown status: ${status}`);
  }

  // Prevent default behavior e.g., form submission or navigation
  $event.preventDefault();
};

// Helper functions
...
</script>

<style scoped lang="scss"></style>
```

This refactoring improves code readability, reduces redundancy, and provides clarity around intended functionality, especially regarding the "Re-create" button actions.

agent/app/service/cronjob_helper.go
```diff
@@ -103,7 +103,6 @@ func (u *CronjobService) loadTask(cronjob *model.Cronjob, record *model.JobRecor
 	case "log":
 		err = u.handleSystemLog(*cronjob, record.StartTime, taskItem)
 	case "snapshot":
-		taskItem.Task.Type = task.TaskScopeSnapshot
 		_ = cronjobRepo.UpdateRecords(record.ID, map[string]interface{}{"records": record.Records})
 		err = u.handleSnapshot(*cronjob, *record, taskItem)
 	}
```
wanghe-fit2cloud, line 108: The provided code snippet is from a function that processes Cronjob configurations and handles different types of tasks such as `log` and `snapshot`. Here are some points to consider:

1. **Typo in Error Handling**:
   The error handling on line 108 seems incorrect because it uses an underscore (`_`) instead of returning the actual error value. This can be improved by replacing `_ := cronjobRepo.UpdateRecords(...)` with `err := cronjobRepo.UpdateRecords(...)`.

2. **Consistent Naming**:
   Use consistent naming conventions throughout the method for better readability. For example, you could rename `taskItem` to `item`, but this does not significantly impact the functionality.

3. **Efficiency**:
   Ensure that all operations within each case statement are optimized. For instance, if updating records involves complex logic, consider refactoring them into separate helper functions.

4. **Documentation**: Add comments explaining key sections of the code, especially around new features or significant changes in behavior, to help maintainers understand and review the functionality.

Here's the corrected version of the segment:
```go
case "log":
 err = u.handleSystemLog(*cronjob, record.StartTime, taskItem)
break;
case "snapshot":
 // Assuming snapshotType is already defined somewhere else
// taskItem.Task.Type = task.TaskScopeSnapshot
 _err := cronjobRepo.UpdateRecords(record.ID, map[string]interface{}{"records": record.Records})
 err = u.handleSnapshot(*cronjob, *record, item)
 break; 
```

Additionally, make sure to include these adjustments when committing your code, providing clear commit messages that outline the changes made and why they were necessary.

wanghe-fit2cloud (commented):
/lgtm

Issue #8671, opened 2025-05-15
[Bug] Domain match issue with non-default HTTPS port


wan92hen:
### 1Panel Version

v2.0.0-beta.1

### Problem Description

When Openresty is configured to use a non-default port for HTTPS (other than port 443), the first site can be accessed using any domain names that are not bound to any sites.

No default site is configured.

### Steps to Reproduce

1. Install Openresty and configure it to use a non-default port for HTTPS (for example, port 8443 instead of the default port 443).
2. Create three DNS records for a.domain.name, b.domain.name, and c.domain.name, pointing them to the public IP address of the 1Panel server.
3. Create two static sites for a.domain.name and b.domain.name, and enable HTTPS for them.
4. When accessing https://c.domain.name:8443/, the content of the first site will be displayed.

### The expected correct result

Accessing https://c.domain.name:8443 should result in a 404 error. 

### Related log output

```shell

```

zhengkunwang223:
Thanks for the feedback. We'll try to reproduce the issue.

wanghe-fit2cloud:
v2.0.0-beta.2 版本已发布。
- **需前往【面板设置】页面，开启【预览体验计划】后即可顺利升级至 v2.0.0-beta.2 版本。**
- **需卸载已安装的 OpenResty，全新安装**

PR #8677, opened 2025-05-15
feat: fix issue with error domain match

zhengkunwang223:
Refs https://github.com/1Panel-dev/1Panel/issues/8671

### DIFF
--- a/agent/app/service/nginx_utils.go
+++ b/agent/app/service/nginx_utils.go
@@ -103,6 +103,32 @@ func getNginxParamsByKeys(scope string, keys []string, website *model.Website) (
 	return res, nil
 }
 
+func updateDefaultServerConfig(enable bool) error {
+	nginxInstall, err := getAppInstallByKey("openresty")
+	if err != nil {
+		return err
+	}
+	defaultConfigPath := path.Join(nginxInstall.GetPath(), "conf", "default", "00.default.conf")
+	content, err := os.ReadFile(defaultConfigPath)
+	if err != nil {
+		return err
+	}
+	defaultConfig, err := parser.NewStringParser(string(content)).Parse()
+	if err != nil {
+		return err
+	}
+	defaultConfig.FilePath = defaultConfigPath
+	defaultServer := defaultConfig.FindServers()[0]
+	defaultServer.UpdateListen(fmt.Sprintf("%d", nginxInstall.HttpPort), enable)
+	defaultServer.UpdateListen(fmt.Sprintf("[::]:%d", nginxInstall.HttpPort), enable)
+	defaultServer.UpdateListen(fmt.Sprintf("%d", nginxInstall.HttpsPort), enable, "ssl")
+	defaultServer.UpdateListen(fmt.Sprintf("[::]:%d", nginxInstall.HttpsPort), enable, "ssl")
+	if err = nginx.WriteConfig(defaultConfig, nginx.IndentedStyle); err != nil {
+		return err
+	}
+	return nginxCheckAndReload(string(content), defaultConfigPath, nginxInstall.ContainerName)
+}
+
 func updateNginxConfig(scope string, params []dto.NginxParam, website *model.Website) error {
 	nginxFull, err := getNginxFull(website)
 	if err != nil {
--- a/agent/app/service/website.go
+++ b/agent/app/service/website.go
@@ -1296,44 +1296,36 @@ func (w WebsiteService) ChangeDefaultServer(id uint) error {
 			return err
 		}
 	}
-	if id > 0 {
-		website, err := websiteRepo.GetFirst(repo.WithByID(id))
-		if err != nil {
-			return err
-		}
-		params, err := getNginxParamsByKeys(constant.NginxScopeServer, []string{"listen"}, &website)
-		if err != nil {
-			return err
-		}
-		httpPort, httpsPort, err := getAppInstallPort(constant.AppOpenresty)
-		if err != nil {
-			return err
-		}
-
-		var changeParams []dto.NginxParam
-		for _, param := range params {
-			paramLen := len(param.Params)
-			bind := param.Params[0]
-			var newParam []string
-			if bind == strconv.Itoa(httpPort) || bind == strconv.Itoa(httpsPort) || bind == "[::]:"+strconv.Itoa(httpPort) || bind == "[::]:"+strconv.Itoa(httpsPort) {
-				if param.Params[paramLen-1] == components.DefaultServer {
-					newParam = param.Params
-				} else {
-					newParam = append(param.Params, components.DefaultServer)
-				}
-			}
-			changeParams = append(changeParams, dto.NginxParam{
-				Name:   param.Name,
-				Params: newParam,
-			})
-		}
-		if err := updateNginxConfig(constant.NginxScopeServer, changeParams, &website); err != nil {
-			return err
-		}
-		website.DefaultServer = true
-		return websiteRepo.Save(context.Background(), &website)
+	if err := updateDefaultServerConfig(!(id > 0)); err != nil {
+		return err
 	}
-	return nil
+	if id == 0 {
+		return nil
+	}
+	website, err := websiteRepo.GetFirst(repo.WithByID(id))
+	if err != nil {
+		return err
+	}
+	params, err := getNginxParamsByKeys(constant.NginxScopeServer, []string{"listen"}, &website)
+	if err != nil {
+		return err
+	}
+	var changeParams []dto.NginxParam
+	for _, param := range params {
+		if hasHttp3(param.Params) || hasDefaultServer(param.Params) {
+			continue
+		}
+		newParam := append(param.Params, components.DefaultServer)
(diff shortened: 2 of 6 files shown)


agent/app/service/website.go
```diff
@@ -1296,44 +1296,36 @@ func (w WebsiteService) ChangeDefaultServer(id uint) error {
...
 	}
-	return nil
+	if err := updateNginxConfig(constant.NginxScopeServer, changeParams, &website); err != nil {
+		return err
+	}
+	website.DefaultServer = true
+	return websiteRepo.Save(context.Background(), &website)
 }
 
 func (w WebsiteService) ChangePHPVersion(req request.WebsitePHPVersionReq) error {
```
wanghe-fit2cloud, line 1331: The provided patch contains several changes to the function `ChangeDefaultServer`. The primary changes include:

1. **Optimization Simplification**: The logic for checking if the server is already marked with "default" has been simplified using a single condition: `!(id > 0)`.

2. **Remove Unnecessary Checks**: Comments about specific port checks like HTTP/S are removed because they were unnecessary given the simplification.

3. **Reintroduction of Logic for Setting Default Server**:
   - It now skips parameters that either already have "http3", which is not used in this context, or those that already contain "default".
   - If no such conditions are met, it appends "default" to the parameter list.
   - Finally, it updates the Nginx configuration and sets the default flag on the website record before saving it.

This refactoring makes the code more concise and easier to understand while maintaining its functionality.

agent/app/service/nginx_utils.go
```diff
@@ -103,6 +103,32 @@ func getNginxParamsByKeys(scope string, keys []string, website *model.Website) (
...
+	defaultServer.UpdateListen(fmt.Sprintf("[::]:%d", nginxInstall.HttpsPort), enable, "ssl")
+	if err = nginx.WriteConfig(defaultConfig, nginx.IndentedStyle); err != nil {
+		return err
+	}
+	return nginxCheckAndReload(string(content), defaultConfigPath, nginxInstall.ContainerName)
+}
+
 func updateNginxConfig(scope string, params []dto.NginxParam, website *model.Website) error {
 	nginxFull, err := getNginxFull(website)
 	if err != nil {
```
wanghe-fit2cloud, line 134: The `updateDefaultServerConfig` function appears to be responsible for updating Nginx configuration settings on Docker containers by enabling or disabling listen ports. Here's a concise review of the function:

### Irregularities and Potential Issues

1. **Missing Error Handling**: The function does not handle errors that may occur during file operations (reading, parsing), container interaction (getting installation key), or reloading checks.

2. **Resource Management**: It doesn't explicitly manage resources like temporary files if necessary.

3. **Synchronization**: There isn't any mention of synchronization between different threads, which could lead to concurrent modifications of configurations.

4. **Configuration File Format Handling**: The current implementation assumes specific formats and doesn't validate them for correctness after modification.

5. **Logging**: No logging is provided, making it difficult to track errors and progress.

   ```go
   if err == nil { // Add error handling here
       return nil, err
   }
   ```

### Optimization Suggestions

1. **Caching Configurations**: Cache parsed configurations and paths to avoid repeated reads, especially if these values haven't changed.

2. **Error Recovery**: Implement better recovery strategies in case some updates fail without affecting others.

3. **Concurrency Control**: Use mutexes or other concurrency controls to ensure thread safety when modifying config files.

4. **Input Validation**: Validate input parameters to prevent unexpected behavior and crashes.

Example improvements:

```go
package main

import (
	"bufio"
	"bytes"
	"fmt"
	"os"

	"github.com/openresty/nginx-parser-go/parser"
)

// ... rest of your functions ...

func updateDefaultServerConfig(enable bool) error {
	var (
		nginxInstall *app.Installation
		err          error
		content      []byte
		config       *parser.ConfigASTNode
		server        *parser.ServerASTNode
		defaultConfigPath string = "/usr/local/nginx/conf/default/00.default.conf" // Example path
	)
	
	// Step 1: Fetch nginx install details
	nginxInstall, err = getAppInstallByKey("openresty")
	if err != nil {
		return fmt.Errorf("Failed to fetch NGINX install details: %w", err)
	}

	// Step 2: Read default server configuration
	path := filepath.Join(nginxInstall.GetPath(), "conf", "default", "00.default.conf")
	content, err = ioutil.ReadFile(path)
	if err != nil {
		return fmt.Errorf("Failed to read default server configuration at %s: %w", path, err)
	}
	
	// Step 3: Parse the configuration
	config, err = parser.NewFileParser(string(content)).Parse()
	if err != nil {
		return fmt.Errorf("Failed to parse default server configuration: %w", err)
	}
	
	// Step 4: Modify the configuration
	for _, s := range config.Servers {
		server = &s
		switch s.Address.String() {
			case "", "::":
				s.Listen = append(s.Listen, nginxi.ParseListener{Port: strfmt.Itoa(int(nginxInstall.HttpPort))}
				s.Listen = append(s.Listen, nginxi.ParseListener{Address: "[::]", Port: strfmt.Itoa(int(nginxInstall.HttpPort))})
			case ":":
				s.ListeningAddresses = map[string]interface{}{
					strconv.FormatInt(nginxi.ToInt64(nginxInstall.HttpPort), 10): struct{}{},
				}
			default:
				s.Listeners = append(s.Listeners, nginxi.ParseListener{Address: s.Address.String(), Port: strfmt.Itoa(int(nginxInstall.HttpsPort)), SSL: true})
				s.Listeners = append(s.Listeners, nginxi.ParseListener{Address: "[" + s.Address.String() + "]", Port: strfmt.Itoa(int(nginxInstall.HttpsPort)), SSL: true})
		}
	}
	
	// Step 5: Write updated configuration back
	newContent, marshalErr := json.MarshalIndent(config, "", "\t")
	if marshalErr != nil {
		return fmt.Errorf("Failed to marshal updated configuration: %w", marshalErr)
	}
	os.WriteFile(path, newContent, modePerm) // Ensure proper permissions set

	// Step 6: Apply changes (reload service)
	reloadCmd := fmt.Sprintf("/bin/sh -c 'docker exec -it %s /sbin/nginx -s reload'", nginxInstall.ContainerNam)
	err = Exec.reloadCommand(reloadCmd, 0)
	if err != nil {
		return fmt.Errorf("Failed to reload NGINX after applying changes: %w", err)
	}
	return nil
}
```
These changes improve error handling, enhance readability, and provide more robustness by managing configurations directly rather than relying on strings and arrays. They also include example code for writing the configuration file and executing commands within Docker containers.

agent/app/service/website_utils.go
```diff
@@ -1395,3 +1395,21 @@ func getSystemProxy(useProxy bool) *dto.SystemProxy {
...
+}
+
+func hasDefaultServer(params []string) bool {
+	for _, param := range params {
+		if param == "default_server" {
+			return true
+		}
+	}
+	return false
+}
```
wanghe-fit2cloud, line 1415: The provided code snippets seem to be part of a Go application dealing with network settings, specifically HTTP proxies. The code includes two additional functions:

1. `hasHttp3(params []string)` checks if there is any param indicating support for QUIC (Quick UDP Internet Connections).
2. `hasDefaultServer(params []string)` checks if there is any parameter that indicates using the default server.

These functions iterate over an array of parameters and return `true` if at least one match is found for 'quic' or 'default_server', respectively. Here are some suggested optimizations:

1. **Comments**: Adding comments to explain what each function does and how it operates can enhance readability.
2. **Error Handling**: Since neither function handles errors explicitly in this context, consider adding error handling for situations where the input might not be properly formatted or processed.
3. **Function Name Clarity**: The names `hasHttp3`, `hasDefaultServer` could potentially benefit from being slightly more descriptive, though they already capture their purpose.
4. **Return Value Consistency**: Ensure that both functions consistently return boolean values based on their logic.

Here's a revised version of the code including these suggestions:

```go
// Determine if the connection should use QUIC as an alternative to HTTP/3
func hasQuicAlternative(configParams []string) bool {
	var supportedProtocols string = "http/3"
	for _, p := range configParams {
		if strings.HasPrefix(p, supportedProtocols) {
			return true
		}
	}
	log.Printf("QUIC protocol not supported or missing in configuration.")
	return false
}

// Check if the specified config options include using the default server
func usesDefaultServer(configParams []string) bool {
	defaultServersParam := "use_default_server=true"
	supportsMultipleConfigurations := "supports_multiple_configurations=false"

	multipleConfigSupported := strings.Contains(strings.Join(configParams, ""), supportsMultipleConfigurations)
	defaultServerSetToTrue := strings.Contains(strings.Join(configParams, ""), defaultServersParam)

	if multipleConfigSupported || defaultServerSetToTrue {
		return true
	} else {
		log.Printf("Configuration does not indicate usage of the default server.")
		return false
	}
}
```

### Key Changes:
- Added a logging statement within `hasQuicAlternative` to handle cases where QUIC is unsupported.
- Improved clarity in `usesDefaultServer` function by checking against specific parameters rather than joining all params into a single string.
- Used `strings.HasPrefix` to check if a protocol starts with the supported protocols, making the logic clearer.
- Kept the approach simple and focused on identifying specific conditions without unnecessary complexity.

wanghe-fit2cloud (commented):
/lgtm

PR #8687, opened 2025-05-16
feat: update WebsiteTypes usage

zhengkunwang223:


frontend/src/views/website/website/index.vue
```diff
@@ -334,7 +334,7 @@ const shortcuts = [
         },
     },
 ];
-
+const WebsiteTypes = getWebsiteTypes();
 const loading = ref(false);
 const maskShow = ref(true);
 const createRef = ref();
```
wanghe-fit2cloud, line 340: The provided code has a few inconsistencies:

1. **Variable Names Mismatch**: The variable name `getWebsiteTypes` is used twice with different syntax. It should be changed to `getWebsiteTypesValue`.

2. **Redundant Imports**: The import statement for `website-types` at the beginning of the file can be removed if you're using the imported value directly without an alias.

3. **Potential Improvements**:
   - Consider caching or memoizing `WebsiteTypes` to avoid unnecessary calls each time it's accessed.
   - Ensure that there’s no circular dependency between modules (e.g., not importing `@/global/mimetype` within itself).

Here's the corrected version of your code with potential improvements included:

```javascript
@@ -315,7 +315,7 @@ import { useI18n } from 'vue-i18n';
 import { getAgentGroupList } from '@/api/modules/group';
 import { Group } from '@/api/interface/group';
-import { WebsiteTypes } from '@/global/mimetype';
+import { getWebsiteTypes, websiteTypes } from '@/global/mimetype'; // Use alias for better clarity

const shortcuts = [
@@ -334,7 +334,7 @@ const shortcuts = [
         },
     },
 ];

// Assuming getWebsiteTypes returns Promise which resolves to a list
getWebsiteTypes().then(response => {
    const WebsiteTypes = response.data; // Assign the returned data to WebsiteTypes
});

let WebsiteTypes = ref(websiteTypes); // If necessary, cache the fetched data here

const loading = ref(false);
const maskShow = ref(true);
const createRef = ref();
```

In this revised version:

- A new alias `websiteTypes` is introduced to clarify usage when assigning the result of `getWebsiteTypes`.
- An example is added on how to fetch and store the results of `getWebsiteTypes`, showcasing basic async operations handling. Adjust this according to your actual implementation details.
- Comments are added for clarity.

PR #8691, opened 2025-05-16
fix: Cronjob backup supports multiple selections

ssongliu:
Refs #8669 #8694

### DIFF
--- a/agent/app/service/cronjob_backup.go
+++ b/agent/app/service/cronjob_backup.go
@@ -27,12 +27,17 @@ func (u *CronjobService) handleApp(cronjob model.Cronjob, startTime time.Time, t
 	if cronjob.AppID == "all" {
 		apps, _ = appInstallRepo.ListBy(context.Background())
 	} else {
-		itemID, _ := strconv.Atoi(cronjob.AppID)
-		app, err := appInstallRepo.GetFirst(repo.WithByID(uint(itemID)))
+		appIds := strings.Split(cronjob.AppID, ",")
+		var idItems []uint
+		for i := 0; i < len(appIds); i++ {
+			itemID, _ := strconv.Atoi(appIds[i])
+			idItems = append(idItems, uint(itemID))
+		}
+		appItems, err := appInstallRepo.ListBy(context.Background(), repo.WithByIDs(idItems))
 		if err != nil {
 			return err
 		}
-		apps = append(apps, app)
+		apps = appItems
 	}
 	if len(apps) == 0 {
 		return errors.New("no such app in database!")
@@ -108,7 +113,6 @@ func (u *CronjobService) handleWebsite(cronjob model.Cronjob, startTime time.Tim
 			u.removeExpiredBackup(cronjob, accountMap, record)
 			return nil
 		}, nil, int(cronjob.RetryTimes), time.Duration(cronjob.Timeout)*time.Second)
-		return nil
 	}
 	return nil
 }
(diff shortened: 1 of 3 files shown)


frontend/src/views/cronjob/cronjob/operate/index.vue
```diff
@@ -1241,6 +1274,15 @@ const onSubmit = async (formEl: FormInstance | undefined) => {
...
+        }
+        if (form.websiteList) {
+            form.website = form.websiteList.join(',');
+        }
+        if (form.dbNameList) {
+            form.dbName = form.dbNameList.join(',');
+        }
 
         form.alertCount = form.hasAlert && isProductPro.value ? form.alertCount : 0;
         form.alertTitle =
```
wanghe-fit2cloud, line 1288: There are no irregularities or major issues with the provided code fragment. However, I have optimized the select component to use a single `@change` event handler that checks for a special value (`'all'`) and updates the list accordingly. Here's an example of how you could implement this:

```typescript
<template>
  <el-select
    clearable
    v-model="form.websiteList"
    multiple
    @change="
      handleWebsiteChange('websiteList', form.websiteList)
    ">
    <el-option
      :disabled="webisteOptions.length === 0"
      label="$t('commons.table.all')"
      value="all">
    </el-option>
    <!-- other options -->
  </el-select>
</template>

<script lang="ts">
interface Props {
  // props definition here
}

defineComponent<Props>({
  setup(props) {
    const handleWebsiteChange = (
      type: string,
      values: string[]
    ) => {
      if (values.includes('all')) {
        form[type] = ['all'];
      } else {
        form[type] = [...new Set(values)];
      }
    };

    return {
      handleWebsiteChange,
    };
  },
});
```

In this modified snippet:

1. The `handleWebsiteChange` function takes two parameters: `type`, which can be "websiteList" or another type related to databases or apps; and `values`, the selected options array.

2. Inside the function, it checks if any option is set to `'all'`. If so, it sets the corresponding field in the form to `['all']`.

3. Otherwise, it converts the current selection array into a unique set using `...new Set(values)` to avoid duplicates before updating the selected fields in the form.

This approach consolidates the logic for handling both website and application selections, making the code cleaner and more maintainable.

agent/app/service/cronjob_backup.go
```diff
@@ -354,11 +361,13 @@ func loadWebsForJob(cronjob model.Cronjob) []model.Website {
...
-		weblist = append(weblist, webItem)
+	websites := strings.Split(cronjob.Website, ",")
+	var idItems []uint
+	for i := 0; i < len(websites); i++ {
+		itemID, _ := strconv.Atoi(websites[i])
+		idItems = append(idItems, uint(itemID))
 	}
+	weblist, _ = websiteRepo.GetBy(repo.WithByIDs(idItems))
 	return weblist
 }
```
wanghe-fit2cloud, line 373: ### Code Differences, Issues, and Optimization Suggestions

1. **Multiple Databases/Websites Handling**:
    - All three sections (`loadDBsForJob`, `handleApp`, and `handleWebsite`) perform similar operations of splitting comma-separated IDs into individual components. However, they use different repositories and handling logic.

   **Suggestion**: Extract this common logic to a separate helper function that handles the conversion of string slice of IDs into their corresponding models or repository queries.

2. **Error Handling**:
    - The code contains multiple error returns without checking if subsequent operations depend on successful completion of previous ones.
  
   **Suggestion**: Ensure all functions return an appropriate error type so that calling functions can easily propagate errors. Handle these errors appropriately at each level of the function call chain.

3. **Code Duplication**:
    - The duplicated logic in `handleApp` and `loadWebsForJob` is redundant and should be abstracted out for better maintainability.

   **Suggestion**: Create reusable functions like `loadEntitiesByIds(crons.Job, stringSliceOfIDs)` that take a job structure and a list of strings (representing identifiers), converts them using appropriate converters, then retrieves the entities from the relevant repositories.

4. **Performance Considerations**:
    - If `appInstallRepo.ListBy()` supports bulk retrieval more efficiently than repeatedly querying by ID, consider optimizing accordingly. For example, query the apps in chunks rather than individually.

5. **Comments and Documentation**:
    - There seems to be inconsistent use of comments across the file. Add more detailed documentation where necessary to clarify complex logic blocks.

   
In summary:

* **Abstract common logic to utility functions:** This reduces redundancy and improves maintainability.
* **Ensure proper error propagation:** Improve error handling in higher-level calls.
* **Optimize data retrieval performance:** Use efficient batch operations when possible.
* **Improve readability with consistent comments/documentation:** Enhance the comprehension of the codebase.


These suggestions will contribute towards cleaner and more robust code.

frontend/src/api/interface/cronjob.ts
```diff
@@ -30,6 +30,10 @@ export namespace Cronjob {
         files: Array<Item>;
         sourceDir: string;
 
+        websiteList: Array<string>;
+        appIdList: Array<string>;
+        dbNameList: Array<string>;
+
         sourceAccounts: Array<string>;
         downloadAccount: string;
         sourceAccountIDs: string;
```
wanghe-fit2cloud, line 39: The code snippet you provided appears to be from TypeScript, likely part of a larger application dealing with cronjobs. Here are some points for review:

### Irregularities/Potential Issues

1. **Variable Naming**: The use of `Array<T>` instead of generic array syntax (`T[]`) is recommended in modern JavaScript/TypeScript. This can lead to better type inference.

2. **Missing Documentation**: There's no documentation or comments explaining the purpose of each variable or the intended usage of this namespace.

3. **Unused Variables and Parameters**: Since there are no additional fields added beyond existing ones, they seem redundant unless used elsewhere.

4. **Potential Confusion with Existing Properties**: It's unclear why these new arrays (`websiteList`, `appIdList`, `dbNameList`) exist if `sourceDir` is also an array. They might serve a different purpose and should be documented clearly.

### Optimization Suggestions

1. **Avoid Redundant Arrays**:
   - If `sourceDir` already serves the purpose, consider removing the `files` array since it seems duplicative.

2. **Use Generic Arrays**:
   - Replace all instances of `Array<T>` with `T[]`. For example:
     ```typescript
     export namespace Cronjob {
         sourceAccounts: string[];
         downloadAccount: string;
         sourceAccountIDs: string[];
     }
     ```

### Updated Code Example

Here's how your updated code could look:

```typescript
export namespace Cronjob {
    // These properties align with the original description but without unnecessary redundancy.
    public sourceDirectories: string[];
    public downloadAccount: string;

    // Consider consolidating paths into a single property if appropriate.
    public pathDetails?: { sourceDirectory: string; ...additionalDetails };

    // Additional logic related to processing directories/files/accounts...
}
```

### Conclusion

While the current implementation doesn’t introduce significant errors, following best practices (such as using generics) and clarifying the purpose and scope of variables and parameters will make the code more maintainable and understandable.

wanghe-fit2cloud (commented):
/lgtm

PR #8698, opened 2025-05-16
fix: Set the default value of Developer Mode to enable

ssongliu:


core/init/migration/migrations/init.go
```diff
@@ -270,3 +270,13 @@ var AddTaskDB = &gormigrate.Migration{
...
+
+var UpdateDeveloperMode = &gormigrate.Migration{
+	ID: "20240516-update-developer-mode",
+	Migrate: func(tx *gorm.DB) error {
+		if err := tx.Model(&model.Setting{}).Where("key = ?", "DeveloperMode").Update("value", constant.StatusEnable).Error; err != nil {
+			return err
+		}
+		return nil
+	},
+}
```
wanghe-fit2cloud, line 282: The provided code snippets are part of a GORM migration process using the gormigrate library. Here's an analysis and some suggestions:

1. **Code Differences**:
   - There is no difference between `AddTaskDB` and `UpdateDeveloperMode`.

2. **Potential Issues**:
   - Neither function handles rollback scenarios, which can lead to data inconsistencies if anything fails during execution.

3. **Optimization Suggestions**:
   - Consider adding a transaction around both migrations to ensure they either fail as one whole unit (all changes apply simultaneously) or succeed completely (no partial updates).

Here’s an optimized version incorporating these considerations:

```go
import (
	"context"
)

var AddTaskDB = &gormigrate.Migration{
	ID: "20240516-add-task-db",
		return tx.AutoMigrate(
			&model.Task{},
		)
	},
}

var UpdateDeveloperMode = &gormigrate.Migration{
	ID: "20240516-update-developer-mode",
		if err := tx.WithContext(context.Background()).Model(&model.Setting{}).
				Where(map[string]interface{}{
					"key":  "DeveloperMode",
				}).Update("value", constant.StatusEnable).Error; err != nil {
			return err
		}
		return nil
	},
}

// Ensure that you wrap all migrations in transactions
func runAllMigrations(dbURL string) error {
	db, err := gorm.Open(sqlite.Open(dbURL), &gorm.Config{})
	if err != nil {
		fmt.Println(err)
		return err
	}
	migator := gormigrate.New(db, gormigrate.DefaultOptions,
		gormigrate.Multi(migrations...))

	err = migator.Run(ctx.Background(), gormigrate.Up)
	if err != nil && !errors.Is(err, sql.ErrNoRows) {
		fmt.Println(err)
		return err
	}
	if _, _err := db.Exec(`PRAGMA foreign_keys=ON;`).Error; _err != nil {
		log.Errorln(_err)
	}
	return nil
}
```

### Key Points:
- Always wrap database operations within transactions using contexts and functions like `db.WithContext`.
- Check for common errors such as SQL errors (`sql.ErrNoRows`) when handling migrations.
- Optionally add logging to track the success or failure of migration steps. 

This approach ensures better reliability and consistency across database interactions.

core/init/migration/migrate.go
```diff
@@ -16,6 +16,7 @@ func Init() {
 		migrations.InitTerminalSetting,
 		migrations.InitGoogle,
 		migrations.AddTaskDB,
+		migrations.UpdateDeveloperMode,
 	})
 	if err := m.Migrate(); err != nil {
 		global.LOG.Error(err)
```
wanghe-fit2cloud, line 22: The changes you've made appear to be adding a new migration function `UpdateDeveloperMode` to the list of migrations when initializing `m`. This is generally fine, but let's go through a quick review:

1. **Regularity Check**: The addition of `migrations.UpdateDeveloperMode` seems syntactically correct within the context of the `Init` function.

2. **Potential Issues**:
   - Ensure that there are no bugs or syntax errors associated with the `UpdateDeveloperMode` function itself.
   
3. **Optimization Suggestions**:
   - If `UpdateDeveloperMode` involves significant database operations, consider breaking it down into smaller steps to improve readability and ensure all changes are applied correctly.
   - Verify the transaction management if necessary, especially in production environments.

In summary, it looks like everything is set up appropriately, so further testing would be required depending on the complexity of `UpdateDeveloperMode` and the overall application requirements.

Issue #8712, opened 2025-05-17
[Bug] Default group setting issue in Website module


tenyon61:
### Contact Information

凌霞News8/tenyon/tenyon@cqbo.com

### 1Panel Version

v2.0.0-beta.2

### Problem Description

网站模块设置默认分组BUG

### Steps to Reproduce

1. 网站
2. 分组
3. 新增分组
4. 设置默认

### The expected correct result

只有一个默认分组

### Related log output

```shell
无需
```

### Additional Information

![Image](https://github.com/user-attachments/assets/4c924501-cb27-4f40-8a2c-36a0e2dff6f8)

zhengkunwang223:
已复现 后续版本修复

wanghe-fit2cloud:
> Bot detected the issue body's language is not English, translate it automatically. 👯👭🏻🧑‍🤝‍🧑👫🧑🏿‍🤝‍🧑🏻👩🏾‍🤝‍👨🏿👬🏿         
----    
Reproduced Result Version Fixed

wanghe-fit2cloud:
v2.0.0-beta.3 版本已发布。

PR #8717, opened 2025-05-17
fix: fix issue with set default group

zhengkunwang223:
Refs https://github.com/1Panel-dev/1Panel/issues/8712


agent/app/repo/group.go
```diff
@@ -69,3 +70,9 @@ func (g *GroupRepo) WithByWebsiteDefault() DBOption {
 		return g.Where("is_default = ? AND type = ?", 1, "website")
 	}
 }
+
+func (g *GroupRepo) CancelDefault(groupType string) error {
+	return global.DB.Model(&model.Group{}).
+		Where("is_default = ? AND type = ?", 1, groupType).
+		Updates(map[string]interface{}{"is_default": 0}).Error
+}
```
wanghe-fit2cloud, line 78: The modified `CancelDefault` method includes several improvements:

1. **Corrected Syntax**: Updated the SQL update to set the `is_default` to 0 correctly.

2. **Improved Formatting**: Added spacing and newlines for better readability.

3. **Added Error Handling**: Ensure that the function returns an explicit error.

Here is the corrected method:
```go
func (g *GroupRepo) CancelDefault(groupType string) error {
	err := global.DB.Model(&model.Group{}).
		Where("is_default = ? AND type = ?", 1, groupType).
		Updates(map[string]interface{}{"is_default": 0})
	if err != nil {
		return err
	}
	return nil
}
```

**Suggestions**:
- Consider adding logging within this method to track cancellation operations.
- Add a test case in your integration or unit tests to verify the functionality of cancelling default groups.

This change ensures proper functionality while adhering to standard Go conventions.

agent/app/service/group.go
```diff
@@ -81,6 +81,11 @@ func (u *GroupService) Delete(id uint) error {
...
 
 func (u *GroupService) Update(req dto.GroupUpdate) error {
+	if req.IsDefault {
+		if err := groupRepo.CancelDefault(req.Type); err != nil {
+			return err
+		}
+	}
 	upMap := make(map[string]interface{})
 	upMap["name"] = req.Name
 	upMap["is_default"] = req.IsDefault
```
wanghe-fit2cloud, line 91: The provided code looks largely correct, but there is one issue to address. The `CancelDefault` method in the repository returns an error, which needs to be checked when calling it:

```go
if req.IsDefault {
    if err := groupRepo.CancelDefault(req.Type); err != nil {
        log.Println("Error canceling default:", err) // Add this line to log errors
        return err
    }
}
```

Additionally, using named returning parameters can improve readability and maintainability of the function:

```go
func (u *GroupService) Update(req dto.GroupUpdate) error {
    upMap := make(map[string]interface{})

    if req.IsDefault {
        if err := groupRepo.CancelDefault(req.Type); err != nil {
            log.Println("Error canceling default:", err)
            return err
        }
    }

    upMap["name"] = req.Name
    upMap["is_default"] = req.IsDefault

    _, err := u.groupRepository.Update(id, upMap)
    if err != nil {
        return fmt.Errorf("failed to update group: %w", err)
    }

    return nil
}
```

This version includes logging of the error if `cancelDefault` fails, and uses the idiomatic way to return an error from the function. Make sure that the `groupRepository`'s `Update` method handles the expected input format and returns its own error if something goes wrong during execution.

agent/app/service/website.go
```diff
@@ -1378,6 +1378,7 @@ func (w WebsiteService) ChangePHPVersion(req request.WebsitePHPVersionReq) error
 			return buserr.New("ErrPHPResource")
 		}
 		website.RuntimeID = req.RuntimeID
+		website.AppInstallID = 0
 		phpProxy := fmt.Sprintf("127.0.0.1:%s", runtime.Port)
 		website.Proxy = phpProxy
 		server.UpdatePHPProxy([]string{website.Proxy}, "")
```
wanghe-fit2cloud, line 1384: The code snippet you've provided makes some adjustments to a `Website` object when handling requests related to changing its PHP version. Here's an analysis of the changes:

1. **Setting `AppInstallID`**: 
   ```go
   website.AppInstallID = 0
   ```
   This seems unnecessary if `runtimeID` should be used instead since it appears that setting `appInstallID` is meant to reflect some association between the runtime ID and application installation details.

2. **Handling Errors**:
   The function checks for errors from various methods like `CheckPHPResources()` but does not handle them consistently. If `CheckPHPResources()` returns an error, it exits early and sends back `buserr.New("ErrPHPResource")`. However, after calling this method, no further error handling is done, which might lead to silent failures unless another error occurs later in the function execution.

3. **Comments**:
   There are comments explaining each step, such as "UpdatePHPProxy([]string{website.Proxy}, "")", but they could potentially add noise if not essential or relevant to all lines of code.

4. **Function Structure**:
   It looks like there might be additional logic beyond the scope of what's shown here, including database updates, log entries, etc., which could be handled after confirming that these other steps succeed without any unexpected errors.

Considering the above points, here are optimizations or improvements suggestions based on the context available:

- Simplify the code: Ensure only necessary fields are updated and avoid redundant assignments.

- Improve error handling: Add more detailed logging or use structured error propagation where appropriate to provide better feedback about what went wrong during different stages of processing.

- Refactor if applicable: Look into whether `ChangePHPVersion` needs to interact with multiple components of the system concurrently, as it involves network configurations and potential side effects.

If you have access to more parts of the codebase, you may want to review how these functions are called elsewhere, any associated business validations, and ensure consistency across similar methods in your service layer. Also, consider adding tests specific to this `ChangePHPVersion` functionality to cover typical edge cases thoroughly.

wanghe-fit2cloud (commented):
/lgtm

Issue #8703, opened 2025-05-16
[Bug] The compressed password and exclusion rules for the scheduled task are invalid


libs-cloud:
### Contact Information

admin@cxle.cn

### 1Panel Version

v2.0.0-beta.2

### Problem Description

计划任务设置网站备份设置了密码 文件则破损 排除路径规则无效

### Steps to Reproduce

创建计划任务站点备份，备份至123云盘的webdav，设置密码后文件破损失效，排除指定路径规则失效
路径排除↓
![Image](https://github.com/user-attachments/assets/bea3668d-7221-49d8-9256-ac458cf65080)
压缩密码↓
![Image](https://github.com/user-attachments/assets/8e766a03-6ef3-421e-809b-c285f528c19d)
选了压缩密码后的压缩包打开后↓
![Image](https://github.com/user-attachments/assets/0ecb19fa-6a1e-4b34-b44d-f9ff90aba136)
排除路径后还在第一代版本没这个问题↓
![Image](https://github.com/user-attachments/assets/baa7a69a-088a-43f8-8cfd-228279b3cca8)

### The expected correct result

上述问题能正常的使用就行，谢谢您小兄弟

### Related log output

```shell

```

wanghe-fit2cloud:
压缩密码借助 OpenSSL 的能力生成，不支持使用 WinRAR 解压，仅在恢复网站时使用该密码。

wanghe-fit2cloud:
排除规则不生效的问题，先在本地环境复现一下。

libs-cloud:
现在这个版本是不是不能排除目录？

wanghe-fit2cloud:
> Bot detected the issue body's language is not English, translate it automatically. 👯👭🏻🧑‍🤝‍🧑👫🧑🏿‍🤝‍🧑🏻👩🏾‍🤝‍👨🏿👬🏿         
----    
Is it impossible to exclude directories in this version now?

ssongliu:
> 现在这个版本是不是不能排除目录？

这个版本支持排除目录的，这里规则设置可能有点不太对

如果 cache 是网站下的一级目录的话，可以尝试一下直接写 `cache,*.log` 就行


加密的压缩文件可以通过系统文件中解压，直接解压是无法解压开的

wanghe-fit2cloud:
[...]

PR #8736, opened 2025-05-19
fix: Fix the problem of the exclusion directory failure in cronjob

ssongliu:
Refs #8703


agent/utils/cmd/cmdx.go
```diff
@@ -64,12 +64,18 @@ func (c *CommandHelper) RunBashCWithArgs(arg ...string) error {
...
-	_, err := c.run("bash", "-c", fmt.Sprintf(command, arg...))
-	return err
+	std, err := c.run("bash", "-c", fmt.Sprintf(command, arg...))
+	if err != nil {
+		return fmt.Errorf("handle failed, std: %s, err: %v", std, err)
+	}
+	return nil
 }
 
 func (c *CommandHelper) RunWithStdout(name string, arg ...string) (string, error) {
```
wanghe-fit2cloud, line 81: The provided code has improvements compared to its previous version but could be further optimized:

1. **Error Handling**: The functions `RunBashC`, `RunBashCWithArgs`, and `RunWithStdout` now include an explicit return of `nil` when no error occurs, providing clarity about the result.

2. **Debugging Information**: By capturing `std` (standard output), each function provides detailed feedback in case of failure (`fmt.Errorf`). This can aid in debugging by showing both standard output and any errors that occurred.

3. **Consistent Return Type**: All three functions return a single `error`. While not strictly necessary, it's generally good practice to have consistent return types across similar methods within a package.

Here is the improved version of the code with these changes:

```go
func (c *CommandHelper) RunBashC(command string) error {
	std, err := c.run("bash", "-c", command)
	if err != nil {
		return fmt.Errorf("run bash failed for command '%s': stderr: %s, stdout: %s, error: %v",
			command, stderr, stdout, err)
	}
	return nil
}

func (c *CommandHelper) RunBashCWithArgs(arg ...string) error {
	std, err := c.run("bash", "-c", strings.Join(arg, " "))
	if err != nil {
		return fmt.Errorf("run bash with args failed for args %v: stderr: %s, stdout: %s, error: %v",
			arg, stderr, stdout, err)
	}
	return nil
}

func (c *CommandHelper) RunWithStdout(name string, arg ...string) (string, error) {
	stdout, stderr, err := c.rawRunWithStdout(name, arg...)
	if err != nil {
		return "", fmt.Errorf("raw run with stdout failed for name '%s' and args %v: stderr: %s, stdout: %s, error: %v",
			name, arg, stderr, stdout, err)
	}
	return stdout, nil
}
```

### Key Changes:
- Added a comprehensive log message in each error-returning function, including `stderr` and `stdout`.
- Used `strings.Join` in `RunBashCWithArgs` to safely construct the argument string.
- Included `name` in the raw run logging message to help identify which specific command was being executed, even if multiple commands are involved.

These modifications enhance the robustness and readability of the code while maintaining consistency among related functions.

agent/app/service/snapshot_create.go
```diff
@@ -476,7 +476,7 @@ func snapPanelData(snap snapHelper, req dto.SnapshotCreate, targetDir string) er
 		}
 		excludes = append(excludes, "."+strings.ReplaceAll(ignore, rootDir, ""))
 	}
-	err := snap.FileOp.TarGzCompressPro(false, rootDir, path.Join(targetDir, "1panel_data.tar.gz"), "", strings.Join(excludes, ";"))
+	err := snap.FileOp.TarGzCompressPro(false, rootDir, path.Join(targetDir, "1panel_data.tar.gz"), "", strings.Join(excludes, ","))
 	snap.Task.LogWithStatus(i18n.GetMsgByKey("SnapCompressPanel"), err)
 	if err != nil {
 		return err
```
wanghe-fit2cloud, line 482: There is a small difference between using semicolons and commas in the `strings.Join` function call inside `TarGzCompressPro`. The original code uses `";"` which might cause some file systems to treat it as an invalid character or improperly split files. It is advisable to use `,` instead for consistency with standard separators used in Linux-like environments.

The corrected line should be:
```go
err := snap.FileOp.TarGzCompressPro(false, global.Dir.LocalBackupDir, path.Join(targetDir, "1panel_backup.tar.gz"), "", strings.Join(excludes, ","))
```

This change ensures that directories with spaces in their names will not cause issues during compression, as comma is more commonly recognized as a valid separator.

frontend/src/lang/modules/ms.ts
```diff
@@ -988,8 +988,7 @@ const message = {
         ntp_helper: 'Anda boleh mengkonfigurasi pelayan NTP di halaman Tetapan Pantas Alat.',
         app: 'Aplikasi sandaran',
         website: 'Laman web sandaran',
-        rulesHelper:
-            'Apabila terdapat pelbagai peraturan pengecualian mampatan, gunakan [,] sebagai pemisah. Contoh: *.log,*.sql',
+        rulesHelper: 'Menyokong pelbagai peraturan pengecualian, dipisahkan dengan koma Inggeris , contoh: *.log,*.sql',
         lastRecordTime: 'Waktu pelaksanaan terakhir',
         all: 'Semua',
         failedRecord: 'Rekod kegagalan',
```
wanghe-fit2cloud, line 994: The code snippet appears to be related to a configuration interface for an application that involves setting preferences such as NTP servers, applications, websites, rules for filtering logs and SQL files, and records of failures. There are no significant errors or issues in this code.

However, there is one potential improvement:

**Rule Helper Suggestion:**

Instead of using the British comma `,` (which is used in United Kingdom), consider switching it to the American comma `,` (commonly used in North America) for consistency across different user interfaces. This might make the documentation more readable for users from both regions.

Here's the updated line with the suggested change:

```javascript
rulesHelper: 'Menyokong pelbagai peraturan pengecualian, dipisahkan dengan koma Amerika , contoh: *.log,*.sql'
```

Overall, the code is clear and well-documented, so there doesn't appear to be any need for further optimizations at this time apart from ensuring regional consistency.

PR #8740, opened 2025-05-19
fix: Fix the problem of abnormal default check items in snapshots

ssongliu:


agent/app/service/snapshot.go
```diff
@@ -120,7 +120,7 @@ func (u *SnapshotService) LoadSnapshotData() (dto.SnapshotData, error) {
 	}
 	for i, item := range itemBackups {
 		if item.Label == "app" {
-			data.BackupData = append(itemBackups[:i], itemBackups[i+1:]...)
+			itemBackups = append(itemBackups[:i], itemBackups[i+1:]...)
 		}
 		if item.Label == "system_snapshot" {
 			itemBackups = append(itemBackups[:i], itemBackups[i+1:]...)
```
wanghe-fit2cloud, line 126: The code contains several issues that need to be addressed:

1. The `data` variable is not defined within this function scope, which means it will cause an error if you try to assign to `data.BackupData`. You should declare and initialize the `data` variable before using it.

2. Even if the `data` variable was properly declared, modifying the slice `itemBackups` directly inside the loop without creating a copy of the slice would lead to unintended behavior because slices reference underlying arrays, and changes made to one affect its original array.

3. Consider adding return values or logging statements outside of the function to better understand where the execution flow could potentially fail.

4. For performance reasons, consider copying elements from an existing list rather than creating new lists with `append()` each time.

Here's a corrected version of the code along with suggested improvements:

```go
func (u *SnapshotService) LoadSnapshotData() (dto.SnapshotData, error) {
	var data dto.SnapshotData

	// ... rest of your code ...

	for i, item := range itemBackups {
		switch item.Label {
		case "app":
			newItemBackups := make([]BackupBackup, len(itemBackups)-1)
			copy(newItemBackups[:i], itemBackups[:i])
			copy(newItemBackups[i:], itemBackups[i+1:])
			itemBackups = newItemBackups
		case "system_snapshot":
			newItemBackups := make([]BackupBackup, len(itemBackups)-1)
			copy(newItemBackups[:i], itemBackups[:i])
			copy(newItemBackups[i:], itemBackups[i+1:])
			itemBackups = newItemBackups
		}
	}

	return data, nil
}
```

In this revised version:
- A new local variable `newItemBackups` is used to store the modified slice.
- The `make()` function ensures that we allocate enough space for the reduced length of `itemBackups`.
- The `copy()` function efficiently copies elements from the old slice to the new slice.
- This approach avoids mutating the original slice and prevents unexpected modifications.

PR #8741, opened 2025-05-19
feat: add bind_domain and ip_limit res

zhengkunwang223:
Refs #8707

### DIFF
--- a/core/app/api/v2/helper/helper.go
+++ b/core/app/api/v2/helper/helper.go
@@ -1,6 +1,8 @@
 package helper
 
 import (
+	"fmt"
+	"github.com/1Panel-dev/1Panel/core/cmd/server/res"
 	"net/http"
 
 	"github.com/1Panel-dev/1Panel/core/app/dto"
@@ -70,5 +72,25 @@ func CheckBindAndValidate(req interface{}, c *gin.Context) error {
 
 func ErrResponse(ctx *gin.Context, code int) {
 	ctx.JSON(code, nil)
+	ctx.Abort()
+}
+
+func ErrWithHtml(ctx *gin.Context, code int, scope string) {
+	if code == 444 {
+		ctx.String(444, "")
+		ctx.Abort()
+		return
+	}
+	file := fmt.Sprintf("html/%d.html", code)
+	if code == 200 && scope != "" {
+		file = fmt.Sprintf("html/200_%s.html", scope)
+	}
+	data, err := res.ErrorMsg.ReadFile(file)
+	if err != nil {
+		ctx.String(http.StatusInternalServerError, "Internal Server Error")
+		ctx.Abort()
+		return
+	}
+	ctx.Data(code, "text/html; charset=utf-8", data)
 	ctx.Abort()
 }
--- a/core/middleware/bind_domain.go
+++ b/core/middleware/bind_domain.go
@@ -1,7 +1,6 @@
 package middleware
 
 import (
-	"errors"
 	"strings"
 
 	"github.com/1Panel-dev/1Panel/core/app/api/v2/helper"
@@ -28,11 +27,8 @@ func BindDomain() gin.HandlerFunc {
 		}
 
 		if domains != status.Value {
-			if LoadErrCode() != 200 {
-				helper.ErrResponse(c, LoadErrCode())
-				return
-			}
-			helper.ErrorWithDetail(c, 311, "ErrInternalServer", errors.New("domain not allowed"))
+			code := LoadErrCode()
+			helper.ErrWithHtml(c, code, "err_domain")
 			return
 		}
 		c.Next()
--- a/core/middleware/ip_limit.go
+++ b/core/middleware/ip_limit.go
@@ -1,7 +1,6 @@
 package middleware
 
 import (
-	"errors"
 	"github.com/1Panel-dev/1Panel/core/utils/common"
 	"strings"
 
@@ -33,10 +32,7 @@ func WhiteAllow() gin.HandlerFunc {
 				return
 			}
 		}
-		if LoadErrCode() != 200 {
-			helper.ErrResponse(c, LoadErrCode())
-			return
-		}
-		helper.ErrorWithDetail(c, 310, "ErrInternalServer", errors.New("IP address not allowed"))
+		code := LoadErrCode()
+		helper.ErrWithHtml(c, code, "err_ip_limit")
 	}
 }


core/app/api/v2/helper/helper.go
```diff
@@ -72,3 +74,23 @@ func ErrResponse(ctx *gin.Context, code int) {
...
+	}
+	data, err := res.ErrorMsg.ReadFile(file)
+	if err != nil {
+		ctx.String(http.StatusInternalServerError, "Internal Server Error")
+		ctx.Abort()
+		return
+	}
+	ctx.Data(code, "text/html; charset=utf-8", data)
+	ctx.Abort()
+}
```
wanghe-fit2cloud, line 96: There are no significant irregularities or major performance concerns in this Go code snippet. Here are some minor improvements:

1. **Imports**: It's good practice to organize imports alphabetically and use the fully qualified names (e.g., `github.com/1Panel-dev/1Panel/core/cmd/server/res`). This can help avoid ambiguities if there are multiple packages with similar name structures.

2. **Function Documentation**: Adding function comments can improve readability and maintenance by explaining what each function does, its parameters, and return values.

3. **Error Handling**: The `ErrWithHtml` function handles different error codes more robustly. However, you might consider adding logic for other potential errors related to reading the HTML file or handling invalid inputs.

4. **Return Type Consistency**: In the `ErrResponse` function, returning `nil` is unnecessary since it will be overwritten later when setting JSON data. Consider changing the return type to an empty interface (`interface{}`).

Here’s a slightly improved version of the code with these considerations:

```go
package helper

import (
	"fmt"
	"io/ioutil"
	"net/http"

	"github.com/1Panel-dev/1Panel/core/cmd/server/res"
	"github.com/gin-gonic/gin"
)

// Returns a generic HTTP response with a provided status code.
// If the status code is 444, returns an empty string without calling Abort().
//
// Usage:
// ctx.JSON(http.StatusNotFound, gin.H{"message": "Not Found"})
func ErrResponse(ctx *gin.Context, statusCode int) {
	ctx.JSON(statusCode, gin.H{})
	if statusCode != 444 {
		ctx.Abort()
	}
}

// Handles errors by returning custom HTML templates based on their status codes.
// Customizes behavior for scope-specific HTML responses.
//
// Parameters:
// - ctx: gin context.
// - statusCode: HTTP status code to handle.
// - scope: Identifier for specific scopes that may affect customizing HTML files.
//
// Usage example:
// helper.ErrWithHtml(c, http.StatusBadRequest, "validation_error")
func ErrWithHtml(ctx *gin.Context, statusCode int, scope string) {
	switch statusCode {
	case 400:
		ctx.HTML(http.StatusBadRequest, "bad_request.html", gin.H{})
	case 500:
		ctx.JSON(http.StatusInternalServerError, gin.H{"message": "Internal Server Error"})
	default:
		ctx.String(http.StatusInternalServerError, "Invalid Status Code")
		return
	}
	ctx.Abort()

	htmlFilePath := fmt.Sprintf("html/%d.html", statusCode)
	if statusCode == 200 && scope != "" {
		htmlFilePath = fmt.Sprintf("html/200_%s.html", scope)
	}

	buf, err := ioutil.ReadFile(htmlFilePath)
	if err != nil {
		statusCode = http.StatusInternalServerError // Default error scenario
		msg := "Failed to read custom HTML template"
		if err.Error() == "file not found" {
			msg += ". Falling back to default message."
		}
		ctx.AbortWithStatusJSON(http.StatusOK, gin.H{
			"status_code": statusCode,
			"error":      msg,
		})
		return
	} else {
		ctx.Data(http.StatusOK, "text/html; charset=utf-8", buf)
	}
}
```

These enhancements make the code cleaner and potentially easier to maintain, especially as it grows in complexity.

core/middleware/bind_domain.go
```diff
@@ -28,11 +27,8 @@ func BindDomain() gin.HandlerFunc {
...
-			if LoadErrCode() != 200 {
-				helper.ErrResponse(c, LoadErrCode())
-				return
-			}
-			helper.ErrorWithDetail(c, 311, "ErrInternalServer", errors.New("domain not allowed"))
+			code := LoadErrCode()
+			helper.ErrWithHtml(c, code, "err_domain")
 			return
 		}
 		c.Next()
```
wanghe-fit2cloud, line 34: The primary issue with this code is that it attempts to use `helper.ErrWithDetail`, which requires an error and a detail message, but the line where this function is called does not include one. A proper fix would be to add a detailed error message when calling helper.ErrWithDetail if necessary.

Additionally, it seems like there's a mistake in the comment about using custom HTTP codes. The constant reference (LoadErrCode()) should either refer to an existing variable or return a correct status code instead of returning 0 without explaining what's supposed to happen.

Lastly, the line "if LoadErrCode() != 200" is redundant since LoadErrCode presumably returns 200 on success, making the condition unnecessary unless specific logic exists elsewhere in the codebase related to other statuses returned from LoadErrCode().

core/middleware/ip_limit.go
```diff
@@ -33,10 +32,7 @@ func WhiteAllow() gin.HandlerFunc {
...
 		}
-		if LoadErrCode() != 200 {
-			helper.ErrResponse(c, LoadErrCode())
-			return
-		}
-		helper.ErrorWithDetail(c, 310, "ErrInternalServer", errors.New("IP address not allowed"))
+		code := LoadErrCode()
+		helper.ErrWithHtml(c, code, "err_ip_limit")
 	}
 }
```
wanghe-fit2cloud, line 38: I see a couple of changes and optimizations you've made to the `WhiteAllow` middleware function:

1. **Variable Removal**: You removed an unnecessary import statement for `errors`, which is not used later.

2. **Error Handling Refinement**:
   - The code now uses a constant from a helper function `LoadErrCode()` instead of hardcoding the error code 200.
   - If this constant does not equal 200, it returns an internal server error with specific details.

3. **HTML Rendering**: Instead of directly handling the error detail using the `helper.ErrorWithDetail()`, they use `helper.ErrWithHtml(...)` method that might render an HTML message based on the status code.

4. **Security Considerations**: Using HTML-based error responses can be risky due to XSS concerns if unvalidated inputs are passed into them. It's advisable to consider sanitizing HTML output when necessary.

Here’s the updated code snippet highlighting these improvements:

```go
@@ -33,18 +32,7 @@ func WhiteAllow() gin.HandlerFunc {
 				return
 			}
 		}

+		code := LoadErrCode()
+
 		switch code {

-		case common.CodeSuccessAuth:
-			// Continue processing further logic here
-			break

 		default:

-			if code != common.CodeSuccessAuth { // Check for non-successful auth or other relevant cases
-
-				helper.ErrorWithHtml(c, int64(code), "err_ip_limit") // Assuming helper.ErrWithHtml sets HTTP Status Code
-				return
-			}

-			helper.SuccessJson(c, "", map[string]string{
-				"ok": true,
-			})
-			common.LogAction(c, nil)
-
+			helper.ErrorWithHtml(c, code, "err_ip_limit")

 		} // End switch
 }
```

### Potential Issues Identified:
- **Hardcoded Success Code Value**: Although you correctly handle different cases, ensure that `common.CodeSuccessAuth` (or similar) properly maps to its intended value before comparison.
  
This modification should be safe to make but always review the security aspects around how your application handles user input to prevent potential vulnerabilities related to rendering HTML content.

PR #8784, opened 2025-05-22
fix: code editor font color

lan-yonghui:


frontend/src/views/host/file-management/code-editor/index.vue
```diff
@@ -424,9 +424,9 @@ const config = reactive<EditorConfig>({
...
     base: 'vs',
     inherit: true,
-    rules: [{ token: '', foreground: 'FFFFFF' }],
+    rules: [{ token: '' }],
     colors: {
-        'editor.background': '#FFFFFF',
+        'editor.background': '#f8f6f6',
         'minimap.background': '#f4f4f4',
         'scrollbar.shadow': '#e1e1e1',
         'scrollbarSlider.background': '#e1e1e1',
```
wanghe-fit2cloud, line 432: The changes you've made to the color `editor.background` appear to be an oversight or correction. In a dark theme, it's common practice for the background color of editors to match the system contrast ratio with text, typically making them lighter than pure white. The new value (`#f8f6f6`) is more appropriate for most dark themes.

For best practices in dark themes:

1. Ensure that the background color contrasts well with the default font color.
2. Use a light version of your main editor color (like `#f5f5f5` instead of plain dark gray `#333333`).

So, I would recommend reverting back to the previous line:
```
colors: {
    'editor.background': '#FFFFFF',
```

wanghe-fit2cloud (commented):
/lgtm

PR #8790, opened 2025-05-23
fix: Fix the issue of abnormal parsing of Chinese nodes

ssongliu:


core/middleware/operation.go
```diff
@@ -32,7 +33,8 @@ func OperationLog() gin.HandlerFunc {
 		source := loadLogInfo(c.Request.URL.Path)
 		pathItem := strings.TrimPrefix(c.Request.URL.Path, "/api/v2")
 		pathItem = strings.TrimPrefix(pathItem, "/api/v2/core")
-		currentNode := c.Request.Header.Get("CurrentNode")
+		currentNodeItem := c.Request.Header.Get("CurrentNode")
+		currentNode, _ := url.QueryUnescape(currentNodeItem)
 		record := &model.OperationLog{
 			Source:    source,
 			Node:      currentNode,
```
wanghe-fit2cloud, line 40: The provided code has several minor issues that could be addressed:

1. **Incorrect Usage of `strings.TrimSpace`:** The function should only be trimming from the start if desired. It's currently removing both leading and trailing spaces.

   **Correction:** Replace:
   ```go
   pathItem := strings.TrimPrefix(c.Request.URL.Path, "/api/v2")
   ```
   With:
   ```go
   pathItem := strings.TrimPrefix(c.Request.URL.Path, "/api/v2/", -1) // Remove first occurrence of "/api/v2/"
   ```

2. **Use of URL Encoding with `url.QueryUnescape`:** Since "CurrentNode" is supposed to be an exact match based on your comment, there isn't much value in using `url.QueryEscape` here because it will escape characters which might not need escaping (like slashes).

   **Correction:** If you genuinely want all parameters to be unescaped but keep "CurrentNode" as exact match, then proceed with:
   ```go
   currentNodeItem := c.Request.Header.Get("CurrentNode")
   nodeParts :=(strings.SplitN(currentNodeItem, "=", 2))
   currentNodeValue := ""
   if len(nodeParts) == 2 && !nodeParts[0].HasSuffix("+") {
       currentNodeValue = decodeURIComponentWithDefault(nodeParts[1], "")
   }
   currentNode, _ = url.QueryUnescape(currentNodeValue)
   }

   func decodeURIComponent(s string, defVal interface{}) string {
       return url.QueryEscape(s)
   }
  ```
   Otherwise, if you don’t intend to perform any decoding after taking out “/api/v2”, just remove `decodeURIComponentWithDefault`.

These changes ensure that string manipulations align better with common use cases where header names must retain their original values unless explicitly stated otherwise through encoding methods like `QueryEscape` or `URL.Encode`.

wanghe-fit2cloud (commented):
/lgtm

PR #8791, opened 2025-05-23
ref: api auth

lan-yonghui:


core/middleware/session.go
```diff
@@ -12,7 +12,8 @@ import (
 
 func SessionAuth() gin.HandlerFunc {
 	return func(c *gin.Context) {
-		if strings.HasPrefix(c.Request.URL.Path, "/api/v2/core/auth") {
+		apiReq := c.GetBool("API_AUTH")
+		if strings.HasPrefix(c.Request.URL.Path, "/api/v2/core/auth") || apiReq {
 			c.Next()
 			return
 		}
```
wanghe-fit2cloud, line 19: The provided code snippet has no major issues. However, there is an inconsistency between checking if the `API_AUTH` flag was set via context (in this case it's hardcoded to true) versus manually inspecting the request URL path with `strings.HasPrefix`. If this behavior change needs to be maintained across different execution contexts or depends on a configuration setting rather than being hardcoded, consider using either approach consistently throughout.

core/init/router/router.go
```diff
@@ -79,14 +79,14 @@ func Routers() *gin.Engine {
...
 	PrivateGroup := Router.Group("/api/v2/core")
 	PrivateGroup.Use(middleware.SetPasswordPublicKey())
 	for _, router := range rou.RouterGroupApp {
 		router.InitRouter(PrivateGroup)
 	}
 
-	Router.Use(middleware.ApiAuth())
 	Router.Use(Proxy())
 	Router.NoRoute(func(c *gin.Context) {
 		if !security.HandleNotRoute(c) {
```
wanghe-fit2cloud, line 92: The changes to the `Routers` function do not have any significant irregularities, but there is one potential issue:

In the second use statement of middleware.ApiAuth(), it should be removed as it duplicates what's already been done when PrivateGroup is defined.

Optimization suggestion: Consider consolidating multiple API authentication checks into one middleware if they serve similar purposes (e.g., ensuring access tokens), rather than applying them repeatedly in different routes or group definitions.

wanghe-fit2cloud (commented):
/lgtm

Issue #8742, opened 2025-05-19
[Bug] Child node CPU occupation exception


jiaxudonggit:
### Contact Information

jiaxudongmail@qq.com

### 1Panel Version

v2.0.0-beta.2

### Problem Description

1panel-agent 进程CPU占用异常，导致服务器CPU占用100%。

<img width="642" alt="Image" src="https://github.com/user-attachments/assets/ca95e39e-e1df-494e-a229-b4d881024dce" />

<img width="1287" alt="Image" src="https://github.com/user-attachments/assets/8706900a-0145-4744-bd1a-6ecd5bcc2742" />

<img width="880" alt="Image" src="https://github.com/user-attachments/assets/5158c05e-2447-4b1a-a800-caf51ab9c141" />

### Steps to Reproduce

-

### Related log output

```shell

```

wanghe-fit2cloud:
想确认下，你有没有尝试手动重启一下 1panel-agent 进程？重启之后 CPU 占用是否恢复正常？
有没有发现是执行了某些操作后导致 CPU 占用突然升高并持续不降？
这些信息有助于我们进一步排查原因。

jiaxudonggit:
> 想确认下，你有没有尝试手动重启一下1panel-agent进程？重启之后CPU占用是否正常？ 有没有发现是执行了某些操作后导致CPU占用突然升高并且持续不降？ 这些信息有助于我们进一步排查原因。

1. 重启面板后恢复正常了
2. CPU升高之前应该是安装了一些应用，以及恢复了应用的备份

wanghe-fit2cloud:
> Bot detected the issue body's language is not English, translate it automatically. 👯👭🏻🧑‍🤝‍🧑👫🧑🏿‍🤝‍🧑🏻👩🏾‍🤝‍👨🏿👬🏿         
----    
> Want to confirm that you have tried to manually restart the 1panel-agent process? Is the CPU occupancy normal after restarting? Have you found that after performing some operations, the CPU usage suddenly increases and continues to decline? This information helps us further troubleshoot the cause.

1. After restarting the panel, it returns to normal
2. Before the CPU is raised, some applications should have been installed and the application backups should be restored.

wanghe-fit2cloud:
[...]

PR #8792, opened 2025-05-23
fix: fix issue with agent  cpu usage error

zhengkunwang223:
Refs https://github.com/1Panel-dev/1Panel/issues/8742

### DIFF
--- a/agent/utils/cmd/cmdx.go
+++ b/agent/utils/cmd/cmdx.go
@@ -2,6 +2,8 @@ package cmd
 
 import (
 	"bytes"
+	"context"
+	"errors"
 	"fmt"
 	"log"
 	"os"
@@ -89,7 +91,17 @@ func (c *CommandHelper) RunWithStdoutBashCf(command string, arg ...interface{}) 
 }
 
 func (c *CommandHelper) run(name string, arg ...string) (string, error) {
-	cmd := exec.Command(name, arg...)
+	var cmd *exec.Cmd
+	var ctx context.Context
+	var cancel context.CancelFunc
+
+	if c.timeout != 0 {
+		ctx, cancel = context.WithTimeout(context.Background(), c.timeout)
+		defer cancel()
+		cmd = exec.CommandContext(ctx, name, arg...)
+	} else {
+		cmd = exec.Command(name, arg...)
+	}
 
 	customWriter := &CustomWriter{taskItem: c.taskItem}
 	var stdout, stderr bytes.Buffer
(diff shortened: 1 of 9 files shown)


agent/utils/cmd/cmdx.go
```diff
@@ -122,25 +134,15 @@ func (c *CommandHelper) run(name string, arg ...string) (string, error) {
...
-		case err := <-done:
-			if err != nil {
-				return handleErr(stdout, stderr, err)
-			}
+		}
+		if err != nil {
+			return handleErr(stdout, stderr, err)
 		}
 		return stdout.String(), nil
 	}
```
wanghe-fit2cloud, line 148: There are few minor issues and opportunities for improvement in the provided `run` function within the CLI application:

### Issues found:
1. **Resource Management**: The use of goroutines to wait on the command's completion is fine when using a timeout but unnecessary if the timeout is not set. It can lead to additional complexity.

   ```go
   // Goroutine usage here is unnecessary because we're handling timeouts separately
   ```

2. **Context Handling**: The initialization and cancellation logic can be simplified.

### Improvements suggested:
1. **Remove Irrelevant Code Block**:
   - Remove the `ctx`, `cancel`, and related logic as they aren't necessary without a specific timeout.

   ```diff
   - var ctx context.Context
   - var cancel context.CancelFunc

   - if c.timeout != 0 {
   -    ctx, cancel = context.WithTimeout(context.Background(), c.timeout)
   -    defer cancel()
   -    cmd = exec.CommandContext(ctx, name, arg...)
   - } else {
   -    cmd = exec.Command(name, arg...)
   - }
   ```

2. **Refactor Timeout Logic**:
   - Simplify the timeout handling by checking if an error from running the command indicates a timeout.

   ```go
   err := cmd.Run()
   if errors.Is(err, os.ErrTimeout) || errors.As(err, (*net.Error)(nil)) { // Assuming net.Error handles other network-related timeouts
       return "", buserr.New("ErrCmdTimeout")
   }
   ```

3. **General Cleanup**:
   - Ensure that any variables declared at the start of the function have been properly initialized before use.

### Modified Functionality:
The revised `run` function eliminates the complex context management and simplifies the timeout logic while still supporting both with and without a timeout:

```go
func (c *CommandHelper) run(name string, arg ...string) (string, error) {
 cmd := exec.Command(name, arg...)

 customWriter := &CustomWriter{taskItem: c.taskItem}
 var stdout, stderr bytes.Buffer
 if _, err := io.Copy(&stdout, cmd.Stdout); err != nil {
 	log.Printf("copying stdout failed: %v", err)
 }
 if _, err := io.Copy(&stderr, cmd.Stderr); err != nil {
 	log.Printf("copying stderr failed: %v", err)
 }

 err := cmd.Wait()

 if err == nil {
 	return stdout.String(), nil
 } else {
 	return handleErr(stdout, stderr, err)
 }
```

These improvements make the function cleaner and more maintainable, especially without relying on goroutine-based async operations for non-timeout cases.

frontend/src/lang/modules/en.ts
```diff
@@ -2821,6 +2821,7 @@ const message = {
             statusCode: 'Status code',
             manage: 'Management',
             autoRestart: 'Auto Restart',
+            EXITED: 'Exited',
         },
     },
     xpack: {
```
wanghe-fit2cloud, line 2827: There is an extra closing parenthesis at the end of the `EXITED` key-value pair for the `'statusMessages'` object. No other immediate irregularities or issues were identified in the code snippet provided.

frontend/src/lang/modules/ja.ts
```diff
@@ -2687,6 +2687,7 @@ const message = {
             statusCode: 'ステータスコード',
             manage: '管理',
             autoRestart: '自動再起動',
+            EXITED: '終了しました',
         },
     },
     xpack: {
```
wanghe-fit2cloud, line 2693: The `xpack` section seems to be missing. This may not cause immediate errors but is recommended to include it for completeness if there are plans to extend functionality related to Elasticsearch X-Pack.

Overall, the code looks clean and follows standard syntax conventions. However, if you have specific areas from which you'd like assistance with further optimization or maintenance, let me know!

wanghe-fit2cloud (commented):
/lgtm

PR #8796, opened 2025-05-23
feat: change cmd timeout handle for core

zhengkunwang223:


core/utils/cmd/cmdx.go
```diff
@@ -112,25 +124,15 @@ func (c *CommandHelper) run(name string, arg ...string) (string, error) {
...
-		case err := <-done:
-			if err != nil {
-				return handleErr(stdout, stderr, err)
-			}
+		}
+		if err != nil {
+			return handleErr(stdout, stderr, err)
 		}
 		return stdout.String(), nil
 	}
```
wanghe-fit2cloud, line 138: The provided code contains several potential issues and areas for improvement:

1. **Context Management**: The use of `context` is not consistently applied throughout the function. Specifically, when a timeout is configured using `c.timeout`, the `command.Ctx` should also be set.

2. **Error Handling in Start Function**: There's an intentional return statement missing after starting the command with `cmd.Start()` on lines 105-106 (commented out).

3. **Resource Allocation**: Allocating large structures (`bytes.Buffer`) can consume resources quickly. This might become problematic if many concurrent tasks are running.

4. **Return Value for Error**: A more specific error message could be added to make it easier to debug.

Here's a revised version of the relevant section considering these points:

```go
func (c *CommandHelper) run(name string, arg ...string) (string, error) {
	var cmd *exec.Cmd
	var ctx context.Context
	var cancel context.CancelFunc

	if c.timeout != 0 {
		ctx, cancel = context.WithTimeout(context.Background(), c.timeout)
		defer cancel()
		cmd = exec.CommandContext(ctx, name, arg...)
	} else {
		cmd = exec.Command(name, arg...)
	}

	customWriter := &CustomWriter{taskItem: c.taskItem}
	var stdout, stderr bytes.Buffer

	if c.timeout != 0 {
		if err := cmd.Start(); err != nil {
			return "", fmt.Errorf("failed to start command '%s': %w", name, err)
		}
		done := make(chan error, 1)
		go func() {
			log.Printf("[DEBUG] Waiting for command '%s' to finish...", name)
			done <- cmd.Wait()
			cancel()
			customWriter.Flush()
		}()
		after := time.After(c.timeout)
		select {
		case <-after:
			err := cmd.Process.Kill()
			if err != nil && !errors.Is(err, os.ErrProcessDone) { // Check if process was already done
				return "",
					fmt.Errorf("timed out waiting for command '%s': killing process returned error:%v", name, err)
			}
			return "", errors.New(fmt.Sprintf("CmdTimedOut%s\nSTDOUT:\n%s\nSTDERR:\n%s",
				cmd.String(),
				stdout,
				stderr))
		case err := <-done:
			if err != nil && !errors.Is(err, os.ErrProcessDone) { // Check if process was already done
				return "",
					fmt.Errorf("process errored while executing command '%s', stopping it due to context cancellation:%v", name, err)
			}
			return stdout.String(), nil
		}
	} else {
		err := cmd.Run()

		// Ensure we properly wait for cancellations before handling errors
		cancel()
		if err = cmd.Wait(); err != nil {
			return handleErr(stdout, stderr, err)
		}
		
		return stdout.String(), nil
	}
}
```

### Key Changes:
- Added detailed logging messages for better debugging.
- Updated the error handling mechanism to log the actual exit status of the command.
- Ensured that all commands are canceled immediately upon timeout or completion.
- Improved clarity in error messages.

wanghe-fit2cloud (commented):
/lgtm

PR #8797, opened 2025-05-23
fix: Fix the problem of Docker restarting caused by proxy synchroniza…

ssongliu:
Refs #8793

### DIFF
--- a/agent/app/service/docker.go
+++ b/agent/app/service/docker.go
@@ -222,15 +222,18 @@ func (u *DockerService) UpdateConf(req dto.SettingUpdate) error {
 		}
 	}
 	if len(daemonMap) == 0 {
-		_ = os.Remove(constant.DaemonJsonPath)
-		if err := restartDocker(); err != nil {
-			return err
+		if len(file) == 0 {
+			return nil
 		}
-		return nil
+		_ = os.Remove(constant.DaemonJsonPath)
+		return restartDocker()
 	}
 	newJson, err := json.MarshalIndent(daemonMap, "", "\t")
 	if err != nil {
 		return err
+	}
+	if string(newJson) == string(file) {
+		return nil
 	}
 	if err := os.WriteFile(constant.DaemonJsonPath, newJson, 0640); err != nil {
 		return err
@@ -239,10 +242,7 @@ func (u *DockerService) UpdateConf(req dto.SettingUpdate) error {
 		return err
 	}
 
-	if err := restartDocker(); err != nil {
-		return err
-	}
-	return nil
+	return restartDocker()
 }
 func createIfNotExistDaemonJsonFile() error {
 	if _, err := os.Stat(constant.DaemonJsonPath); err != nil && os.IsNotExist(err) {


agent/app/service/docker.go
```diff
@@ -222,25 +222,31 @@ func (u *DockerService) UpdateConf(req dto.SettingUpdate) error {
...
 		return err
 	}
 
-	if err := restartDocker(); err != nil {
-		return err
+	if withRestart {
+		return restartDocker()
 	}
 	return nil
 }
```
wanghe-fit2cloud, line 246: The changes made to `DockerService` include:
1. The addition of a parameter `withRestart` to the `UpdateConf()` method, allowing developers to specify whether to restart Docker after updating its configuration files.
2. Simplification: When no changes were detected (`len(daemonMap) == 0`), the file was removed before attempting to restart Docker because it's not necessary in this condition.
3. Optimization: Removed redundant checks like `if err := os.Remove(constant.DaemonJsonPath); err != nil { return err }`. If there is an error during removal, it should prevent further processing but won't be returned by the function until the next line.

This improves the functionality and robustness of the Docker service while maintaining clarity in how updates are handled.

agent/app/api/v2/docker.go
```diff
@@ -72,7 +72,7 @@ func (b *BaseApi) UpdateDaemonJson(c *gin.Context) {
 		return
 	}
 
-	if err := dockerService.UpdateConf(req); err != nil {
+	if err := dockerService.UpdateConf(req, true); err != nil {
 		helper.InternalServer(c, err)
 		return
 	}
```
wanghe-fit2cloud: There is a minor change in the `UpdateDaemonJson` function from version 2021 to 2025:
In version 2021:
```go
dockerService.UpdateConf(req)
```
In version 2025, the argument `true` was added:
```go
dockerService.UpdateConf(req, true)
```

This addition changes the behavior of `dockerService.UpdateConf`, which may mean different outcomes depending on how `dokerService.UpdateConf` handles its arguments after this change.

As an alternative, consider removing all parameters if it's not intentional to pass any additional arguments while updating the daemon JSON. If passing `true` is necessary for a specific purpose, ensure documentation or comments clarify why `true` is being used and if there might be side effects due to optional or unexpected default values set within `dockerService.UpdateConf`.

agent/app/service/docker.go
```diff
@@ -222,27 +222,27 @@ func (u *DockerService) UpdateConf(req dto.SettingUpdate) error {
...
 	}
 
-	if err := restartDocker(); err != nil {
-		return err
-	}
-	return nil
+	return restartDocker()
 }
 func createIfNotExistDaemonJsonFile() error {
 	if _, err := os.Stat(constant.DaemonJsonPath); err != nil && os.IsNotExist(err) {
```
wanghe-fit2cloud, line 248: The code has several issues:

1. **Potential Memory Leak**: The file handle for the `file` variable is not closed after being used in the conditional statement before it's checked if its contents match `newJson`. This should be fixed.

2. **Incorrect Error Handling for File Existence Check**: The current implementation checks if the file exists but doesn't use the returned error to return an appropriate result. Consider checking for errors explicitly.

3. **Duplicate Logic with Restart Docker Function**: There's redundant logic where `restartDocker()` is called multiple times, which can be simplified.

Here’s the corrected version of the code snippet along with optimizations:

```go
  daemonMap := make(map[string]string)

  // ... existing code remains unchanged ...

  newJson, err := json.MarshalIndent(daemonMap, "", "\t")
  if err != nil {
    return fmt.Errorf("failed to marshal JSON: %v", err)
  }

  fileExists, err := os.Stat(constant.DaemonJsonPath)
    return err
  }

  defer func() {
    if fh := recover(); fh != nil || err := fileClose(file); err != nil {
      fmt.Printf("Error encountered closing file: %s\n", err)
    }
  }()

  var content []byte

  if fileExists != nil && !os.IsExist(fileExists.Err()) {
    fileInfo, err := os.ReadFile(constant.DaemonJsonPath)
    if err != nil {
      return fmt.Errorf("failed to read file: %v", err)
    }
    content = fileInfo.Bytes()
  } else {
    content = make([]byte, 0)
  }

  if string(newJson) == string(content) {
    return nil
  }

  if err = os.WriteFile(constant.DaemonJsonPath, newJson, 0640); err != nil {
    return fmt.Errorf("failed to write to file: %v", err)
  }

  if err := validateDockerConfig(); err != nil {
    return fmt.Errorf("invalid configuration: %v", err)
  }

  return restartDocker()
}

// Assuming this function handles closing files properly
func fileClose(file io.Closer) error {
  if c, ok := file.(io.ReadCloser); ok {
    _ = c.Close()
  }
  return file.Close()
}
```

### Key Changes:
- Corrected memory leak by ensuring the file handle is closed with a deferred call.
- Improved handling of file existence and reading/writing operations.
- Simplified the conditional logic related to restarting Docker.

PR #8828, opened 2025-05-26
fix: Fix issue with duplicate nodes appearing after adding a node.

zhengkunwang223:


frontend/src/components/docker-proxy/dialog.vue
```diff
@@ -33,38 +33,37 @@ const em = defineEmits(['update:withDockerRestart', 'submit']);
...
+            emit();
             return;
         }
         if (res.data.proxyDocker === '') {
-            em('update:withDockerRestart', false);
-            em('submit');
+            emit();
             return;
         }
         open.value = true;
```
wanghe-fit2cloud, line 69: The provided code seems mostly correct but could potentially be improved with some minor adjustments to ensure clarity and optimize performance where feasible:

### Improvements Proposed

1. **Inline Functions**: Instead of using a separate `emit` function, you can directly call `em` from inside conditions without defining an additional function. This keeps the logic more concise.

2. **Try-Catch Block Simplification**: Since both branches in the `try-catch` block end up calling `emit()`, there is no need for duplicate lines after the error capture.

Here's how these improvements would look:

```javascript
const acceptParams = async (props: DialogProps): Promise<void> => {
    if (props.syncList.indexOf('SyncSystemProxy') === -1) {
        em('update:withDockerRestart', false);
        em('submit');
        return;
    }

    try {
        if (res.data.proxyType === '' || res.data.proxyType === 'close') {
            em('update:withDockerRestart', false);
            em('submit');
            return;
        }

        let searchXSetting;
        const xpackModules = import.meta.glob('../../xpack/api/modules/setting.ts', { eager: true });
        if (xpackModules['../../xpack/api/modules/setting.ts']) {
            searchXSetting = xpackModules['../../xpack/api/modules/setting.ts']['searchXSetting'] || {};
            const res = await searchXSetting();
            
            if (!res) {
                em('update:withDockerRestart', false);
                em('submit');
                return;
            }
 
            if (res.data.proxyDocker === '') {
                em('update:withDockerRestart', false);
                em('submit');
                return;
            }
           
            open.value = true;
        }
    } catch (error) {
        em('update:withDockerRestart', false);
        em('submit');
    }
}
```

### Additional Notes

- The changes should not affect the correctness of the functionality.
- These improvements enhance readability by reducing redundancy while maintaining maintainability.

Issue #8809, opened 2025-05-24
[Bug] OpenResty Upstream Editor Error


kohirei:
### 1Panel Version

v2.0.0-beta.3

### Problem Description

创建网站中负载均衡编辑器错误，使用调试工具

发送请求带有空变量
{"websiteID":1,"name":"test1","operate":"edit","servers":[{"server":"111.180.189.224","weight":"","flag":""}],"algorithm":"default","flag":""}

未触发输入动作，正常情况下
{"websiteID":1,"name":"test1","operate":"edit","servers":[{"server":"111.180.189.224","flag":""}],"algorithm":"default","flag":""}

### Steps to Reproduce

![Image](https://github.com/user-attachments/assets/12b237d2-5a8d-46a7-8196-7579d43bb728)

点击确认报错
请求参数错误: json: cannot unmarshal string into Go struct field NginxUpstreamServer.servers.weight of type int

### Related log output

```shell

```

wanghe-fit2cloud:
感谢反馈，我们先在本地环境复现一下这个问题。

zhengkunwang223:
感谢反馈 已复现 后续版本修复

wanghe-fit2cloud:
> Bot detected the issue body's language is not English, translate it automatically. 👯👭🏻🧑‍🤝‍🧑👫🧑🏿‍🤝‍🧑🏻👩🏾‍🤝‍👨🏿👬🏿         
----    
Thanks for the feedback Reproduced Subsequent version fix

wanghe-fit2cloud:
v2.0.0-beta.4 版本已发布。

PR #8829, opened 2025-05-26
fix: Fix issue with update loadbalance error

zhengkunwang223:
Refs #8809

### DIFF
--- a/frontend/src/views/website/website/config/basic/load-balance/operate/index.vue
+++ b/frontend/src/views/website/website/config/basic/load-balance/operate/index.vue
@@ -210,6 +210,17 @@ const submit = async (formEl: FormInstance | undefined) => {
         if (!valid) {
             return;
         }
+        for (const server of item.value.servers) {
+            if (!server.weight || server.weight == '') {
+                server.weight = 0;
+            }
+            if (!server.maxFails || server.maxFails == '') {
+                server.maxFails = 0;
+            }
+            if (!server.maxConns || server.maxConns == '') {
+                server.maxConns = 0;
+            }
+        }
         loading.value = true;
         try {
             if (item.value.operate === 'edit') {


frontend/src/views/website/website/config/basic/load-balance/operate/index.vue
```diff
@@ -210,6 +210,17 @@ const submit = async (formEl: FormInstance | undefined) => {
...
+            if (!server.maxFails || server.maxFails == '') {
+                server.maxFails = 0;
+            }
+            if (!server.maxConns || server.maxConns == '') {
+                server.maxConns = 0;
+            }
+        }
         loading.value = true;
         try {
             if (item.value.operate === 'edit') {
```
wanghe-fit2cloud, line 226: The code snippet provided appears to be part of an asynchronous form submission function in a Vue.js application, specifically related to handling server configuration data. Here's a brief analysis:

1. **Variable Declaration**: The `submit` function is defined to handle the form submission logic.

2. **Input Validation Check**:
   ```javascript
   if (!valid) {
       return; // If input validation fails, return early
   }
   ```
   - This line ensures that further processing only occurs if the input is valid.

3. **Server Weight Handling**:
   ```javascript
   for (const server of item.value.servers) {
       if (!server.weight || server.weight == '') {
           server.weight = 0; // Assign default zero weight if it's missing or empty
       }
       if (!server.maxFails || server.maxFails == '') {
           server.maxFails = 0; // Assign default zero maxFails if it's missing or empty
       }
       if (!server.maxConns || server.maxConns == '') {
           server.maxConns = 0; // Assign default zero maxConns if it's missing or empty
       }
   }
   ```
   - These lines ensure that `weight`, `maxFails`, and `maxConns` values are set to zero if they are not present or empty strings. This could improve user experience by applying defaults where unspecified.

4. **Loading Indication**:
   ```javascript
   loading.value = true; // Set loading indicator to True
   try {
       if (item.value.operate === 'edit') {
           await instanceEditFormRef?.update(item.value);
       } else if (item.value.operate === 'add') {
           await instanceAddFormRef?.save();
       } else {
           console.log('operate err');
       }
       successMessage.value("Configuration has been saved."); // Show success message upon successful operation
       router.go(-1); // Navigate back to previous page after saving
   } catch (error) { // Handle error on save attempt
       errorMessage.value(`Failed to ${item.value.operate}: ${JSON.stringify(error.message.split("\n")[0])}`); // Log failure with specific error information
       reload(); // Reload the current route
   } finally {
       loading.value = false; // Reset loading indicator when done
   }
   ```
   - This block manages form updates based on whether the operation is edit or add, handles errors gracefully using a try-catch block, logs messages appropriately, navigates back to the previous route after a successful update, and resets the loading state afterward.

Overall, the code is mostly well-structured and follows standard conventions, but there are two optimization points you might consider:

1. **Destructuring Objects**:
   To avoid checking `undefined` multiple times, especially for nested properties like objects accessed via `.value`, destructuring can streamline the access:
   ```javascript
   for (const { servers } of item.value) {
       servers.forEach(server => {
           const { weight = 0, maxFails = 0, maxConns = 0 } = server;
           // Use these destructured variables here instead of accessing them repeatedly
       });
   }
   ```

2. **Prevent Redundant Checks**:
   Although redundant checks here (e.g., converting empty strings to numbers and setting to zero), this behavior is intended and correct for certain applications. If strict type checking were needed, it’s best handled separately from business logic.

In summary, the main considerations here are readability for maintenance purposes and ensuring robustness against potentially malformed inputs. The suggested optimizations can enhance performance slightly while maintaining functionality.

wanghe-fit2cloud (commented):
/lgtm

PR #8846, opened 2025-05-27
fix: fix edit docker proxy restart

lan-yonghui:


frontend/src/views/setting/panel/proxy/index.vue
```diff
@@ -160,7 +157,7 @@ const submitChangePassword = async (formEl: FormInstance | undefined) => {
         }
         if (isMasterProductPro.value && (params.proxyDocker || proxyDockerVisible.value)) {
             dockerProxyRef.value.acceptParams({
-                syncList: 'SyncSystemProxy',
+                syncList: 'SyncSystemProxyDocker',
             });
         } else {
             loading.value = true;
```
wanghe-fit2cloud, line 163: 1. The `confirmDialogRef` should be removed since it's not used anywhere in the component after it was defined.

2. In `submitChangePassword`, use `proxyDockerVisible.value` instead of the old variable name.

These changes ensure that the code remains consistent and free from unnecessary dependencies or bugs.

frontend/src/components/docker-proxy/dialog.vue
```diff
@@ -38,6 +38,10 @@ const emit = () => {
     em('submit');
 };
 const acceptParams = async (props: DialogProps): Promise<void> => {
+    if (props.syncList.includes('SyncSystemProxyDocker')) {
+        open.value = true;
+        return;
+    }
     if (props.syncList.indexOf('SyncSystemProxy') === -1) {
         emit();
         return;
```
wanghe-fit2cloud, line 47: The provided code snippet appears to be part of a dialog management system where `syncList` is an array that contains different synchronization options. The function `acceptParams` is designed to handle these options and perform various actions based on the contents of `syncList`.

### Potential Issues:
1. **Undefined Variable**: There's a reference to `open`, which might not have been initialized earlier in the code. If `open` should be a Vue ref or boolean variable, it needs to be defined before being used.

2. **Code Duplication**: Conditional checks inside the function can sometimes lead to duplicate blocks. For example, checking if `'SyncSystemProxy'` is present in `syncList` multiple times could use common logic to avoid redundancy.

3. **Async/Return Behavior**: The function currently uses `await` for asynchronous operations (`emit()`), but there is no mention of awaiting anything else. This may affect how synchronous tasks behave when combined with other asynchronous calls.

4. **Potential Infinite Loop**: If `props.syncList` includes `'SyncSystemProxyDocker'`, it will set `open.value` to `true` and return immediately without waiting for further operations. However, this behavior isn't immediately clear from the context.

### Optimization Suggestions:

1. **Consistent Variable Declaration**:
   Ensure that all variables referenced within functions are properly declared. If `open` is meant to be a reactive value (e.g., using Vue.js's reactivity system), you should declare it as such at its initialization point.

```javascript
// Assuming open is supposed to be a Vue ref
const {ref} = require('@vue/reactivity');

const open = ref(false);
```

2. **Refactoring Conditional Checks**:
   Consolidate similar conditional checks into reusable logic.

```javascript
function includeInSyncList(list, item) {
    return list.includes(item);
}

if (!includeInSyncList(props.syncList, 'SyncSystemProxy')) {
    emit();
}
```

3. **Avoid Unnecessary Returns**:
   Consider removing unnecessary returns if they don’t add functionality. Here’s a more streamlined version:

```javascript
const acceptParams = async (props: DialogProps): Promise<void> => {
    if (includeInSyncList(props.syncList, 'SyncSystemProxyDocker')) {
        open.value = true;
    } else if (props.syncList.indexOf('SyncSystemProxy') === -1) {
        emit();
    }
};
```

These suggestions aim to improve readability, maintainability, and potentially performance by organizing complex logic better.

PR #8896, opened 2025-06-03
fix: Fix the problem of unfiltered types in the backup database of cr…

ssongliu:
…onjob

### DIFF
--- a/agent/app/service/cronjob_backup.go
+++ b/agent/app/service/cronjob_backup.go
@@ -309,12 +309,13 @@ func loadDbsForJob(cronjob model.Cronjob) []DatabaseHelper {
 	var dbs []DatabaseHelper
 	if cronjob.DBName == "all" {
 		if cronjob.DBType == "mysql" || cronjob.DBType == "mariadb" {
-			mysqlItems, _ := mysqlRepo.List()
+			databaseService := NewIDatabaseService()
+			mysqlItems, _ := databaseService.LoadItems(cronjob.DBType)
 			for _, mysql := range mysqlItems {
 				dbs = append(dbs, DatabaseHelper{
 					ID:       mysql.ID,
 					DBType:   cronjob.DBType,
-					Database: mysql.MysqlName,
+					Database: mysql.Name,
 					Name:     mysql.Name,
 				})
 			}


agent/app/service/cronjob_backup.go
```diff
@@ -309,12 +309,13 @@ func loadDbsForJob(cronjob model.Cronjob) []DatabaseHelper {
...
+			mysqlItems, _ := databaseService.LoadItems(cronjob.DBType)
 			for _, mysql := range mysqlItems {
 				dbs = append(dbs, DatabaseHelper{
 					ID:       mysql.ID,
 					DBType:   cronjob.DBType,
-					Database: mysql.MysqlName,
+					Database: mysql.Name,
 					Name:     mysql.Name,
 				})
 			}
```
wanghe-fit2cloud, line 321: The provided code snippet seems to be attempting to load database information based on certain criteria from an interface-based repository (`MySQLRepo`), but it makes two mistakes that need correction:

1. **Misuse of Database Helper Method**: The line `dbs = append(dbs, DatabaseHelper{...})` implies that there is a method called `LoadItems` available within the `databaseService`. However, this method doesn't exist in the given code context.

2. **Improper Naming Convention**: The variable name `databaseService` should likely refer to a service or factory type related to loading databases instead of directly accessing its methods.

To correct these issues without modifying other parts of the function unnecessarily, you can refactor the code as follows:

```go
func loadDbsForJob(cronjob model.Cronjob) []DatabaseHelper {
    var dbs []DatabaseHelper

    if cronjob.DBName == "all" {
        if cronjob.DBType == "mysql" || cronjob.DBType == "mariadb" {
            // Assuming MySQLRepo has LoadAll() instead of List()
            mysqlItems, _ := mysqlRepo.LoadAll()

            // Alternatively, create a separate service for MySQL operations:
            dbService := NewDBServiceInterface() // Replace with actual implementation
            items, err := dbService.ListMysql()
            if err != nil {
                log.Println("Error fetching MySQL data:", err)
            } else {
                for _, item := range items {
                    dbs = append(dbs, DatabaseHelper{
                        ID:       item.ID,
                        DBType:   cronjob.DBType,
                        Database: item.Name, // Adjust field according to your DB struct definition
                        Name:     item.Name,
                    })
                }
            }

        } else { ... }
    } else { ... }
}
```

### Key Points:
- If `mysqlRepo` has a more convenient method like `LoadAll()` which returns all databases, then use that.
- Ensure naming conventions align appropriately (e.g., using `NewIDatabaseService`, creating `dbService` for clarity).
- Properly handle errors when fetching data from services or repositories.

wanghe-fit2cloud (commented):
/lgtm

PR #8898, opened 2025-06-03
fix: Fix the problem of failed creation of child node snapshots

ssongliu:
Refs #8897

### DIFF
--- a/agent/app/service/snapshot_create.go
+++ b/agent/app/service/snapshot_create.go
@@ -268,11 +268,7 @@ func loadDbConn(snap *snapHelper, targetDir string, req dto.SnapshotCreate) erro
 		_ = taskDB.Where("id = ?", req.TaskID).Delete(&model.Task{}).Error
 	}
 	if !req.WithOperationLog {
-		err = snap.snapCoreDB.Exec("DELETE FROM operation_logs").Error
-		snap.Task.LogWithStatus(i18n.GetMsgByKey("SnapDeleteOperationLog"), err)
-		if err != nil {
-			return err
-		}
+		_ = snap.snapCoreDB.Exec("DELETE FROM operation_logs").Error
 	}
 	if !req.WithLoginLog {
 		err = snap.snapCoreDB.Exec("DELETE FROM login_logs").Error


agent/app/service/snapshot_create.go
```diff
@@ -268,11 +268,7 @@ func loadDbConn(snap *snapHelper, targetDir string, req dto.SnapshotCreate) erro
...
 	if !req.WithOperationLog {
-		err = snap.snapCoreDB.Exec("DELETE FROM operation_logs").Error
-		snap.Task.LogWithStatus(i18n.GetMsgByKey("SnapDeleteOperationLog"), err)
-		if err != nil {
-			return err
-		}
+		_ = snap.snapCoreDB.Exec("DELETE FROM operation_logs").Error
 	}
 	if !req.WithLoginLog {
 		err = snap.snapCoreDB.Exec("DELETE FROM login_logs").Error
```
wanghe-fit2cloud, line 274: The code looks mostly clean and functional. However, the deletion of `operation_logs` if `withOperationLog` is false can be optimized slightly. Instead of using error checking for delete operations, one could simply check if the number of deleted rows is zero to ensure at least some records were cleared.

Here's an improved version:

```diff
    @@ -268,11 +268,7 @@ func loadDbConn(snap *snapHelper, targetDir string, req dto.SnapshotCreate) erro
 		_ = taskDB.Where("id = ?", req.TaskID).Delete(&model.Task{}).Error
 	}
 	if !req.WithOperationLog {
-		err = snap.snapCoreDB.Exec("DELETE FROM operation_logs").Error
-		snap.Task.LogWithStatus(i18n.GetMsgByKey("SnapDeleteOperationLog"), err)
-		if err != nil || res.RowsAffected == 0 {
-			return err
-		}
+		res := snap.snapCoreDB.Exec("DELETE FROM operation_logs")
+		if err := res.Error; err != nil || res.RowsAffected == 0 {
+			return err
  	 }
```

This change reduces redundancy by combining the execution of the query with the error handling in one line, while still verifying that at least one row was deleted (as indicated by non-zero RowsAffected after executing the DELETE statement).

Additionally, I'm not sure where `res` is used later in the function. If it's intended to use the result from the DELETE operation, this might need further modification based on its actual usage elsewhere in the code.

wanghe-fit2cloud (commented):
/lgtm

PR #8926, opened 2025-06-05
fix: Fix some snapshot anomaly issues

ssongliu:
Refs #8917
Refs #8924
Refs #8897

### DIFF
--- a/agent/app/service/snapshot.go
+++ b/agent/app/service/snapshot.go
@@ -118,15 +118,14 @@ func (u *SnapshotService) LoadSnapshotData() (dto.SnapshotData, error) {
 	if err != nil {
 		return data, err
 	}
-	for i, item := range itemBackups {
-		if item.Label == "app" {
-			itemBackups = append(itemBackups[:i], itemBackups[i+1:]...)
-		}
-		if item.Label == "system_snapshot" {
-			itemBackups = append(itemBackups[:i], itemBackups[i+1:]...)
+	i := 0
+	for _, item := range itemBackups {
+		if item.Label != "app" && item.Label != "system_snapshot" {
+			itemBackups[i] = item
+			i++
 		}
 	}
-	data.BackupData = itemBackups
+	data.BackupData = itemBackups[:i]
 
 	return data, nil
 }
@@ -329,7 +328,7 @@ func loadPanelFile(fileOp fileUtils.FileOp) ([]dto.DataTree, error) {
 			itemData.IsDisable = true
 		case "clamav", "download", "resource":
 			panelPath := path.Join(global.Dir.DataDir, itemData.Label)
-			itemData.Children, _ = loadFile(panelPath, 5, fileOp)
+			itemData.Children, _ = loadFile(panelPath, 3, fileOp)
 		case "apps", "backup", "log", "db", "tmp":
 			continue
 		}
@@ -376,7 +375,6 @@ func loadFile(pathItem string, index int, fileOp fileUtils.FileOp) ([]dto.DataTr
 	if err != nil {
 		return data, err
 	}
-	i := 0
 	for _, fileItem := range snapFiles {
 		itemData := dto.DataTree{
 			ID:      uuid.NewString(),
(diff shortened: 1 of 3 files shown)


agent/app/service/snapshot.go
```diff
@@ -400,7 +400,6 @@ func loadFile(pathItem string, index int, fileOp fileUtils.FileOp) ([]dto.DataTr
 			itemData.Size = uint64(fileInfo.Size())
 		}
 		data = append(data, itemData)
-		i++
 	}
 	return data, nil
 }
```
wanghe-fit2cloud, line 405: There is an issue in the `LoadSnapshotData` function where the labels "app" and "system_snapshot" are removed from the list, but there's no corresponding logic to replace these with valid options if they exist. The `loadPanelFile` function has issues related to handling different types of items ('clamav', 'download', 'resource') differently compared to others ('apps', 'backup', 'log', 'db', 'tmp'). Additionally, the `loadFile` function does not handle directories correctly since it doesn't add them as children nodes when appropriate. Finally, the use of variable names can be improved for clarity.

agent/app/service/snapshot_create.go
```diff
@@ -267,10 +267,14 @@ func loadDbConn(snap *snapHelper, targetDir string, req dto.SnapshotCreate) erro
...
+		snap.Task.LogWithStatus(i18n.GetMsgByKey("SnapDeleteOperationLog"), err)
+		if err != nil {
+			return err
+		}
 	}
-	if !req.WithLoginLog {
+	if !req.WithLoginLog && global.IsMaster {
 		err = snap.snapCoreDB.Exec("DELETE FROM login_logs").Error
 		snap.Task.LogWithStatus(i18n.GetMsgByKey("SnapDeleteLoginLog"), err)
 		if err != nil {
```
wanghe-fit2cloud, line 280: There are no significant issues with the provided code snippet. The changes made to handle deletion of `operation_logs` and `login_logs` when requested depend on whether they are enabled (`WithOperationLog`) and if it's a master server (`global.IsMaster`). This structure enhances code clarity and ensures that the necessary operations are performed only under specific conditions. However, since these operations involve potentially deleting sensitive data, careful consideration should be given to how this logic is implemented in a production environment. If error handling can be moved elsewhere more efficiently without repeating, further optimizations may be beneficial, but overall, the current implementation seems well-structured.

frontend/src/views/setting/snapshot/create/index.vue
```diff
@@ -147,7 +147,7 @@
         <template #footer>
             <el-button @click="drawerVisible = false">{{ $t('commons.button.cancel') }}</el-button>
             <el-button @click="prev" v-if="nowIndex !== 0">{{ $t('commons.button.prev') }}</el-button>
-            <el-button type="primary" v-if="nowIndex === 4" @click="submitAddSnapshot">
+            <el-button type="primary" v-if="nowIndex === 4" :disabled="loading" @click="submitAddSnapshot">
                 {{ $t('commons.button.create') }}
             </el-button>
             <el-button @click="next" v-else>{{ $t('commons.button.next') }}</el-button>
```
wanghe-fit2cloud, line 153: In the provided code snippet, there is an issue with the button label that appears after the fourth step of the process (index 4). The button is intended to show "Create Snapshot", but it incorrectly has `type="primary"` applied when you are on the last step instead. This can be confusing for users.

To fix this problem and ensure consistency across all steps, you should change `v-if="nowIndex === 4"` to use `:disabled` instead, so that the button remains enabled throughout all steps except the final one where only the "Next" button would appear:

```html
<el-button type="primary" v-if="nowIndex < 3" @click="submitAddSnapshot"> 
```

This modification ensures that the "Create Snapshot" button is disabled on the fifth step (`nowIndex === 4`), maintaining proper functionality and user interface.

Additionally, it's a good practice to avoid having conditional classes if they can be handled using conditional rendering directly in Vue components, rather than trying to force them into different types based on conditions.

wanghe-fit2cloud (commented):
/lgtm

PR #8929, opened 2025-06-05
fix: Fix issue with custom rules being lost after restarting OpenResty.

zhengkunwang223:
Refs https://github.com/1Panel-dev/1Panel/issues/8923

### DIFF
--- a/agent/app/service/app_install.go
+++ b/agent/app/service/app_install.go
@@ -264,12 +264,6 @@ func (a *AppInstallService) Operate(req request.AppInstalledOperate) error {
 		}
 		return syncAppInstallStatus(&install, false)
 	case constant.Restart:
-		if install.App.Key == "openresty" {
-			websites, _ := websiteRepo.GetBy()
-			if len(websites) > 0 {
-				_ = createAllWebsitesWAFConfig(websites)
-			}
-		}
 		out, err := compose.Restart(dockerComposePath)
 		if err != nil {
 			return handleErr(install, err, out)
--- a/frontend/src/views/website/runtime/php/config/index.vue
+++ b/frontend/src/views/website/runtime/php/config/index.vue
@@ -8,6 +8,9 @@
     >
         <template #content>
             <el-tabs tab-position="left" v-model="index">
+                <el-tab-pane :label="$t('php.containerConfig')" name="6">
+                    <Container :id="runtime.id" v-if="index == '6'"></Container>
+                </el-tab-pane>
                 <el-tab-pane :label="$t('website.updateConfig')" name="0">
                     <Config :id="runtime.id" v-if="index == '0'"></Config>
                 </el-tab-pane>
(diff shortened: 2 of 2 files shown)


frontend/src/views/website/runtime/php/config/index.vue
```diff
@@ -44,7 +44,7 @@ import PHP from './php-fpm/index.vue';
 import Performance from './performance/index.vue';
 import Container from './container/index.vue';
 
-const index = ref('0');
+const index = ref('6');
 const open = ref(false);
 const runtime = ref({
     name: '',
```
wanghe-fit2cloud, line 50: There is an inconsistency in the `ref` values used between the two instances of `<el-tabs>` sections:

**Before Change:** 
```vue
<el-tab-pane :label="$t('website.updateConfig')" name="0">
```

**After Change:**
```vue
<el-tab-pane :label="'FPM ' + $t('website.source')" name="3">
```

The `name` attributes "0" and "3" do not match the label text "$t('php.containerConfig')". This will cause the tabs to be labeled differently from their content.

Additionally, you might consider setting default values based on certain conditions (e.g., if user preferences or data dictate a specific tab should open by default). 

Optimizations could involve reducing repetition within components like `<PHP>`, especially since they appear twice with different types. Also, ensure that Vue's reactive references (`ref`) are correctly set up across all related components to maintain state consistency.

agent/app/service/app_install.go
```diff
@@ -264,12 +264,6 @@ func (a *AppInstallService) Operate(req request.AppInstalledOperate) error {
...
 	case constant.Restart:
-		if install.App.Key == "openresty" {
-			websites, _ := websiteRepo.GetBy()
-			if len(websites) > 0 {
-				_ = createAllWebsitesWAFConfig(websites)
-			}
-		}
 		out, err := compose.Restart(dockerComposePath)
 		if err != nil {
 			return handleErr(install, err, out)
```
wanghe-fit2cloud, line 269: There is an issue with this code related to the restart operation when `install.App.Key` equals "openresty". The function attempts to get all websites using `websiteRepo.GetBy()`, which assumes there's a method like that defined on the repository interface.

Here's what needs fixing:

Replace it with the actual implementation from your website repository. Assuming you need data from a specific table or query condition, modify it accordingly:

```go
if install.App.Key == "openresty" {
  // Replace this line with the correct API call from your database
  websites, err := websiteRepo.GetOpenRestyRelatedWebsites()
  
  if err != nil || len(websites) == 0 {
    // Handle error appropriately (e.g., log or update installation status)
    return syncAppInstallStatus(&install, true) // Indicate failure to perform the desired task    
  }

  _ = createAllWebsitesWAFConfig(websites)
}

out, err := compose.Restart(dockerComposePath)
```

This ensures better correctness depending on how your website retrieval logic works and prevents runtime errors caused by missing or incorrect function calls.

wanghe-fit2cloud (commented):
/lgtm

PR #8941, opened 2025-06-06
fix: add port check for php runtime port config

zhengkunwang223:


agent/app/repo/runtime.go
```diff
@@ -64,8 +64,13 @@ func (r *RuntimeRepo) WithNotId(id uint) DBOption {
...
-		return g.Where("port REGEXP ?", regexPattern)
+		return g.Debug().Where(
+			"port = ? OR port LIKE ? OR port LIKE ? OR port LIKE ?",
+			portStr,
+			portStr+",%",
+			"%,"+portStr,
+			"%,"+portStr+",%",
+		)
 	}
 }
```
wanghe-fit2cloud, line 76: There are no apparent issues with the code in this diff. However, you can consider the following optimizations:

1. Replace `WHERE field REGEXP 'pattern'` with `SELECT id FROM table WHERE field LIKE %pattern%`. This will be more efficient on databases that support it.

2. You might want to add error handling for cases where port is not an integer or out of range, depending on your application's requirements.

3. Consider using a separate option function for each DBOption if it helps readability or maintainability.

In terms of performance, since there may be many records, optimizing the query could potentially improve response times.

agent/app/service/runtime.go
```diff
@@ -1075,6 +1079,7 @@ func (r *RuntimeService) UpdatePHPContainer(req request.PHPContainerConfig) erro
 	for k, v := range newMap {
 		envs[k] = v
 	}
+	envs["PANEL_APP_PORT_HTTP"] = runtime.Port
 	envStr, err := gotenv.Marshal(envs)
 	if err != nil {
 		return err
```
wanghe-fit2cloud, line 1085: The provided code snippets appear to be part of a function in a Go application that manages PHP containers within Docker using a runtime service. Here are some points and suggestions:

### Code Differences and Potential Issues
1. **Duplicate `hostPorts` Variable Creation**: You have an unnecessary variable `hostPorts` declared at line 1088 with the same purpose as what was used earlier on line 2. It can be removed without affecting functionality.

   ```go
- var hostPorts []string
```

2. **Empty Loop Body**: The inner loop body is empty from line 1074 to 1088. This might suggest it's meant as a placeholder or comment but could also indicate logic errors if not commented out properly.

3. **Hardcoded Default Port Check**: At line 1062, there’s a check for default port 9000 which suggests some sort of validation against predefined ports like this. Ensure that you handle this appropriately considering whether such checks should only apply during certain stages (e.g., development vs production).

   ```go
-	if export.ContainerPort == 9000 {
+	if export.ContainerPort == 9000 && !req.Config.Debloated { // Adjust based on actual use case
       return buserr.New("ErrPHPPortIsDefault")
   }
   ```

4. **Error Handling Logic**:
   - In line 1055, after calling `checkRuntimePortExist`, the error handling doesn't seem complete since it only returns non-nil errors.
   
   ```go
   -	return err
   ```
   
   Consider adding more comprehensive error handling depending on your business requirements (e.g., logging errors, retry mechanisms etc.).

5. **Missing Comments/Docstrings**: There are no comments explaining key parts, even though they look essential according to your mention about needing to review the changes. Adding clear docstrings can help others understand the intent better.

6. **Environment Variable Addition**: At line 1079, `"PANEL_APP_PORT_HTTP"` environment variable is added directly inside the configuration object (`runtime`). It would typically need to be updated dynamically during deployment or reconfiguration rather than hardcoded.

   ```go
   -	envs["PANEL_APP_PORT_HTTP"] = runtime.Port
   ```

### Optimization Suggestions

1. **Code Readability**: While the current structure may initially be intuitive, readability improvements might enhance maintainability over time. Split larger functions into multiple smaller ones where applicable.

   ```go
   func validateHostPortConstraints(port int) *buserr.BusErrorHandler {
       if port < 1 || port > 65535 {
           return buserr.New("Invalid port number: must be between 1 and 65535.")
       }
       // Add more specific constraints if necessary
       return nil
   }

   // Usage example:
   if err := validateHostPortConstraints(export.HostPort); err != nil {
       return err
   }
   ```

2. **Concurrency Enhancements**: If dealing with multiple container updates concurrently, consider implementing concurrency safe operations to avoid race conditions.

3. **Performance Tuning**:
   - Use buffered channels instead of unbuffered ones in multi-threaded environments when sending responses back to clients.
   - Optimize database queries related to checking port existence by pre-indexing critical fields.
   - Avoid repeated calls to external APIs unless absolutely necessary due to network latency costs.

### Conclusion
By addressing these issues, the overall code will become more robust, readable, and efficient. Additionally, incorporating additional features or refactoring existing functionalities based on further analysis and understanding could lead to a more sophisticated solution tailored to the needs of the application running this code base.

wanghe-fit2cloud (commented):
/lgtm

PR #8965, opened 2025-06-09
fix: Resolve the error in reading container logs

zhengkunwang223:


agent/app/service/container.go
```diff
@@ -945,7 +945,7 @@ func collectLogs(params dto.StreamLog, messageChan chan<- string, errorChan chan
 		message := scanner.Text()
 		select {
 		case messageChan <- message:
-		case <-time.After(time.Second):
+		case <-time.After(5 * time.Second):
 			errorChan <- fmt.Errorf("message channel blocked")
 			return
 		}
```
wanghe-fit2cloud, line 951: The provided code snippet has an issue where `errorChan` is written to after it has already been closed, which can lead to a runtime panic if any goroutines attempt to write to it afterwards. The correct approach would be to check whether `errorChan` is still open before writing to it.

Additionally, there's some redundancy in the use of `select`, since both conditions inside the select block (`<-messageChan`) and `<-time.After()` handle the same message processing logic. This could potentially make the code less readable. A more efficient way would be to have a separate channel specifically for handling timeout errors or using a timer that closes itself directly when a message is read from the other channel within its interval:

```go
func collectLogs(params dto.StreamLog, messageChan chan<- string, errorChan chan<- error) {
    scanner := bufio.NewScanner(io.Reader)
    go func() {
        defer scanner.Close()

        timer := time.NewTimer(5 * time.Second)
        defer timer.Stop()

        for message := scanner.Text(); ; message = strings.TrimSpace(scanner.Text()), timer.Reset(5*time.Second) {
            select {
            case messageChan <- message:
                return // Exit loop on successful message transmission
            case <-errorChan:
                return // Exit loop if channel is closed (e.g., due to context cancellation)
            case <-timer.C: // Timeout reached before log message received
                errorChan <- fmt.Errorf("message channel blocked")
                return
            }
        }
    }()
}
```

This change ensures better resource management by closing the timer upon receiving a valid log message or encountering an immediate error condition. It also removes redundant logic within the loop.

wanghe-fit2cloud (commented):
/lgtm

PR #8966, opened 2025-06-09
fix: Resolve the error in file management

zhengkunwang223:


frontend/src/components/system-upgrade/index.vue
```diff
@@ -3,28 +3,28 @@
...
-                    <el-link :underline="false" class="version" type="primary" @click="copyText(version)">
+                    <el-link underline="never" class="version" type="primary" @click="copyText(version)">
                         {{ version }}
                     </el-link>
                     <el-badge is-dot class="-mt-0.5" :hidden="version === 'Waiting' || !globalStore.hasNewVersion">
-                        <el-link class="ml-2" :underline="false" type="primary" @click="onLoadUpgradeInfo">
+                        <el-link class="ml-2" underline="never" type="primary" @click="onLoadUpgradeInfo">
                             {{ $t('commons.button.update') }}
                         </el-link>
                     </el-badge>
```
wanghe-fit2cloud, line 30: Some minor issues can be addressed here:

1. **`<el-link>` Tag Attributes**:
   - Ensure that `type="primary"` is consistent across all `<el-link>` instances for consistency.
   - The use of `:underline="false"` or `underline="never"` might depend on other styling requirements but should not affect functionality.

2. **Button Text Translations**:
   - Verify that translations (`$t()` calls) remain accurate and up-to-date with current content.

3. **Link Classes**:
   - The `class="flex flex-wrap items-center"` is applied twice which seems redundant; ensure the styles are correctly set elsewhere if necessary.

4. **Badge Display Condition**:
   - Confirm that `version !== 'Waiting' || !globalStore.hasNewVersion` makes sense logically—ensure it's intended to hide the badge under certain conditions.

These checks should help identify simple issues before proceeding further with implementation or deployment.

frontend/src/views/host/file-management/index.vue
```diff
@@ -706,9 +706,10 @@ const searchFile = async () => {
...
     if (isHidden.value) {
-        data.value = res.data.items.filter((item) => !item.isHidden);
+        const items = res.data.items || [];
+        data.value = items.filter((item) => !item.isHidden);
     } else {
-        data.value = res.data.items;
+        data.value = res.data.items || [];
     }
     dirNum.value = data.value.filter((item) => item.isDir).length;
     fileNum.value = data.value.filter((item) => !item.isDir).length;
```
wanghe-fit2cloud, line 715: The only change between versions is that `false` has been replaced with `"never"` in the `underline` attribute of two `<el-link>` components. This change does not introduce any new features, bugs, or performance optimizations.

**Summary:** 
There are no major changes to be addressed in this code snippet. The update to use `"never"` instead of `false` for linking styles should have no impact on functionality, but it's good practice to ensure consistent coding style across the project.

If there were other specific areas needing review or concern, those would need to be identified separately before making decisions.

frontend/src/components/file-list/index.vue
```diff
@@ -60,7 +60,7 @@
                             ></svg-icon>
 
                             <template v-if="!row.isCreate">
-                                <el-link :underline="false" @click="open(row)">
+                                <el-link underline="never" @click="open(row)">
                                     {{ row.name }}
                                 </el-link>
                             </template>
```
wanghe-fit2cloud, line 66: The provided code snippet appears to be a Vue.js template with modifications aimed at improving the appearance and functionality of certain components using Element UI library elements like `svg-icon`, `el-link`, and `el-popover`. The changes include changing the value of the `underline` property from `"false"` to `"never"`, which is not an actual property of Element UI's `el-link` component.

Here are some observations:

1. **Underline Property**: In Element UI, the `link-underline` prop can have three values: `"always"`, `"never"`, or `"hover"`.
   - `"always"` means the link will always have underline.
   - `"never"` means the link should never have underline.
   - `"hover"` means the link will have underline only on hover.

2. **Correct Usage of Link Underlining**:
   To correctly set the link underlining for both the external links in the `paths.slice(1, -1)` iteration and the name displayed when not creating (`row.isCreate`), you should use `"hover"` or remove the `underline` attribute entirely if no specific styling is desired.

3. **Recommended Changes**:
   Assuming you want the default behavior of underline (like `<a>...</a>` tags do) instead of disabling it completely, you can change the attributes like this:

```diff
@@ -22,7 +22,7 @@
                     <div class="hidden-paths">
                         <div v-for="(item, index) in paths.slice(1, -1)" :key="index">
                             <svg-icon :class="'table-icon'" iconName="p-file-folder"></svg-icon>
-                            <el-link :underline="false" @click="jump(index + 1)">{{ item }}</el-link>
+                            <el-link hover @click="jump(index + 1)">{{ item }}</el-link>
                         </div>
                     </div>
                 </el-popover>
@@ -60,7 +60,7 @@
                             ></svg-icon>
 
                             <template v-if="!row.isCreate">
-                                <el-link :underline="false" @click="open(row)">
+                                <el-link hover @click="open(row)">
                                     {{ row.name }}
                                 </el-link>
                             </template>
```

However, if you specifically don't want any underlines, then the most accurate update would be removing the entire `underline` attribute altogether:

```diff
@@ -22,7 +22,6 @@
                     <div class="hidden-paths">
                         <div v-for="(item, index) in paths.slice(1, -1)" :key="index">
                             <svg-icon :class="'table-icon'" iconName="p-file-folder"></svg-icon>
-                            <el-link :underline="false" @click="jump(index + 1)">{{ item }}</el-link>
+                            <el-link @click="jump(index + 1)">{{ item }}</el-link>
                         </div>
                     </div>
                 </el-popover>
@@ -60,7 +59,6 @@
                             ></svg-icon>
 
                             <template v-if="!row.isCreate">
-                                <el-link :underline="false" @click="open(row)">
+                                <el-link @click="open(row)">
                                     {{ row.name }}
                                 </el-link>
                             </template>
```

Choose the one that best fits your design requirements. If the intention was to disable the underline, the second version is more direct; if keeping it as "auto" or "default," the first version should work.

wanghe-fit2cloud (commented):
/lgtm

PR #8968, opened 2025-06-09
fix: Delete the redundant migrations

ssongliu:


core/init/migration/migrations/init.go
```diff
@@ -274,16 +275,6 @@ var AddTaskDB = &gormigrate.Migration{
...
-		if err := tx.Model(&model.Setting{}).Where("key = ?", "DeveloperMode").Update("value", constant.StatusEnable).Error; err != nil {
-			return err
-		}
-		return nil
-	},
-}
-
 var AddXpackHideMenu = &gormigrate.Migration{
 	ID: "20250529-add-xpack-hide-menu",
 	Migrate: func(tx *gorm.DB) error {
```
wanghe-fit2cloud, line 280: There seems to be some redundant imports near the bottom of the file that can cause confusion. Here's an optimized version:

```go
package migrations

import (
	"encoding/json"
	"fmt"
	"path"
	"strings"
	"time"

	dbModel "github.com/1Panel-dev/1Panel/core/app/model"
	constant "github.com/1Panel-dev/1Panel/core/constant"
)

var AddTaskDB = &gormigrate.Migration{
	ID: "20240516-add-task-db",
		return dbModel.Task{}.DropTableIfExists().
			CreateTables(dbModel.Task{})
	},
}

// var UpdateDeveloperMode is unused and can be deleted.
```

Additionally, adding `dbModel` before each call simplifies the usage since it clearly indicates which package they belong to.

This ensures the code is clean and easier to read while maintaining functionality.

core/init/migration/migrate.go
```diff
@@ -16,7 +16,6 @@ func Init() {
 		migrations.InitTerminalSetting,
 		migrations.InitGoogle,
 		migrations.AddTaskDB,
-		migrations.UpdateDeveloperMode,
 		migrations.AddXpackHideMenu,
 	})
 	if err := m.Migrate(); err != nil {
```
wanghe-fit2cloud, line 21: The `UpdateDeveloperMode` function call should be removed from the migration list unless that functionality is intended to be executed within a specific migration step. This could potentially lead to missing updates if migrations are executed independently.

wanghe-fit2cloud (commented):
/lgtm

PR #8971, opened 2025-06-09
feat: change default proxy.conf

zhengkunwang223:


core/utils/security/security.go
```diff
@@ -56,7 +56,12 @@ func CheckSecurity(c *gin.Context) bool {
...
 	c.Writer.WriteHeader(http.StatusOK)
-	_, _ = c.Writer.Write(web.IndexByte)
+	data, err := web.IndexHtml.ReadFile("index.html")
+	if err != nil {
+		c.String(http.StatusInternalServerError, "index.html not found")
+		return
+	}
+	_, _ = c.Writer.Write(data)
 	c.Writer.Flush()
 }
```
wanghe-fit2cloud, line 67: The changes you've made seem largely correct, although there is a small issue with the return type of `ToIndexHtml`. If it should write HTML directly to HTTP response instead of returning an error or boolean value, consider modifying its type accordingly.

Here's your updated code:

```go
func ToIndexHtml(c *gin.Context) {
    c.Writer.Header().Set("Content-Type", "text/html; charset=utf-8")
    c.Writer.WriteHeader(http.StatusOK)

    data, err := web.IndexHtml.ReadFile("index.html")
    if err != nil {
        c.String(http.StatusInternalServerError, "index.html not found")
        return
    }

    _, _ = c.Writer.Write(data)
	c.Writer.Flush()
}
```

### Potential Improvements:
1. **Error Handling**: Ensure proper error handling for other cases (e.g., directory traversal attacks).
2. **Performance**: For larger files, consider streaming the file content rather than reading it all into memory at once.
3. **Caching**: Implement caching strategies if this method is executed frequently by different clients.
4. **Security**: Be cautious when loading static assets from disk, especially in production environments.

These enhancements ensure better robustness and performance of the function.

core/cmd/server/web/web.go
```diff
@@ -8,8 +8,5 @@ var IndexHtml embed.FS
 //go:embed assets/*
 var Assets embed.FS
 
-//go:embed index.html
-var IndexByte []byte
-
 //go:embed favicon.png
 var Favicon embed.FS
```
wanghe-fit2cloud, line 12: There is no apparent issue with the provided code. Here are some minor improvements for better clarity:

```go
//go:embed assets/*
var Assets embed.FS

//go:embed favicon.png
var Favicon embed.FS
```

These two lines can be merged for cleaner syntax while maintaining the same functionality. It's a matter of preference and style choice.

**Note**: The `IndexByte` variable was removed because embedding multiple files of the same name within an embed directive results in only one file being embedded and accessible via that name. If you were to include `index.html` separately later on, it would conflict with this single entry under "assets". 

If preserving both versions (`Assets` for other HTML files and `IndexByte` for a specific index.html) is necessary, you don't need to remove anything - they simply share different names but contain the same set of files.

wanghe-fit2cloud (commented):
/lgtm

PR #8988, opened 2025-06-10
fix: Fix the problem of abnormal acquisition of system update versions

ssongliu:


core/app/service/upgrade.go
```diff
@@ -309,7 +309,7 @@ func (u *UpgradeService) loadVersionByMode(developer, currentVersion string) (st
 		return betaVersionLatest, "", latest
 	}
 	if current[:5] == currentVersion[:5] {
-		return betaVersionLatest, current, ""
+		return betaVersionLatest, "", ""
 	}
 	return betaVersionLatest, "", latest
 }
```
wanghe-fit2cloud, line 315: The provided code has several potential issues and improvements that might affect its functionality:

1. **Redundant Return Values**: The function returns three values but only uses two of them (`betaVersionLatest`, `current`). It should return one or all three values based on how they are used in the calling context.

2. **Error Handling**: There is no explicit error handling for parsing integers or dealing with non-string inputs, which could handle unexpected data gracefully.

3. **Variable Scope**: Unused variables like `num` and `developer` can be removed to improve readability and maintainability.

4. **Comments**: Add more comments to describe what each part of the code does, especially where conditions depend on specific substring comparisons.

5. **Efficiency**: The use of repeated substring comparisons using `strings.HasPrefix()` can be optimized if these substrings do not change after initialization.

Here's an improved version of the code considering some of these points:

```go
func (u *UpgradeService) loadVersionByMode(currentVersion string) (string, bool, bool) {
    versionPart := strings.Split(currentVersion, ".")

    // Determine if it matches beta patterns
    matchBetaPattern := func(prefix, suffix string) bool {
        return strings.HasPrefix(currentVersion, prefix+suffix)
    }

    switch {
    case len(versionPart) < 3:
        return betaVersionLatest, false, true
    case num > 10 && matchBetaPattern("v", "pre-release"):
        return betaVersionLatest, true, true
    case num == 10 && matchBetaPattern("v", ""):
        return betaVersionLatest, false, true
    default:
        return betaVersionLatest, matchBetaPattern("", ""), false
    }
}
```

### Key Changes:
- **Function Signature**: Changed the function signature to return `(version string, matchedPreRelease bool, matchedStable bool)` based on the logic.
- **Removed Redundant Returns**: Only the version pattern that fits best for the given condition is returned.
- **Improved Comments**: Added explanation for why certain blocks are executed.
- **Logical Simplification**: Combined related checks into one line using a helper function `matchBetaPattern` for clearer logic.

This approach reduces redundancy and improves clarity while maintaining the necessary functionality.

wanghe-fit2cloud (commented):
/lgtm

PR #8993, opened 2025-06-10
feat: change 1panel v1 docs url

zhengkunwang223:


frontend/src/store/modules/global.ts
```diff
@@ -49,7 +49,7 @@ const GlobalStore = defineStore({
             state.themeConfig.theme === 'dark' ||
             (state.themeConfig.theme === 'auto' && window.matchMedia('(prefers-color-scheme: dark)').matches),
         isDarkGoldTheme: (state) => state.themeConfig.primary === '#F0BE96' && state.isProductPro,
-        docsUrl: (state) => (state.isIntl ? 'https://docs.1panel.pro' : 'https://1panel.cn/docs'),
+        docsUrl: (state) => (state.isIntl ? 'https://docs.1panel.pro' : 'https://1panel.cn/docs/v1'),
     },
     actions: {
         setOpenMenuTabs(openMenuTabs: boolean) {
```
wanghe-fit2cloud, line 55: There appears to be an issue with modifying the `docsUrl` getter function in your Vuex store. Currently, it's returning different URLs based on whether `state.isIntl` is true or false, but this does not account for the version change from v1 to v2.

### Suggested Changes:

```diff
@@ -49,7 +49,7 @@ const GlobalStore = defineStore({
             state.themeConfig.theme === 'dark' ||
             (state.themeConfig.theme === 'auto' && window.matchMedia('(prefers-color-scheme: dark)').matches),
         isDarkGoldTheme: (state) => state.themeConfig.primary === '#F0BE96' && state.isProductPro,
-        docsUrl: (state) => (state.isIntl ? 'https://docs.1panel.pro' : 'https://1panel.cn/docs'),
+        docsUrl: (state) => (state.isIntl ? 'https://docs.1panel.pro' : `https://1panel.cn/docs/v${parseInt(state.versionNumber[1])}`), // Assuming state.versionNumber contains a string representation of the current version
     },
```

### Explanation:

The new suggestion assumes that there is a property called `versionNumber` within your state which represents the current application version number as a dotted decimal string (`"vX.Y"`). This allows you to dynamically construct the documentation URL depending on the specific version being used instead of always returning the same fixed URL.

Replace `"v1"` in the example with `'v' + parseInt(state.versionNumber[1])`, assuming `state.versionNumber` looks like `["", "1"]`. The first element of `versionNumber` will be empty by default when `versionNumber` is undefined or null, avoiding any potential errors with indexing. Make sure to adjust this logic if your implementation structure differs from what I have described here.

wanghe-fit2cloud (commented):
/lgtm

Issue #9241, opened 2025-06-23
[Bug] Error occurred while viewing logs in the Python runtime environment


kuilei0926:
### 1Panel Version

 专业版 v2.0.2

### Problem Description

scanner error: bufio.Scanner: token too long

/api/v2/containers/search/log?compose=/opt/1panel/runtime/python/python/docker-compose.yml&since=all&tail=100&follow=true&operateNode=local

event: error
data: scanner error: bufio.Scanner: token too long

event: error
data: scanner error: bufio.Scanner: token too long

### Steps to Reproduce

python运行环境查看日志或者docker容器查看日志

### Related log output

```shell

```

zhengkunwang223:
感谢反馈 我们先复现一下

wanghe-fit2cloud:
v2.0.3 版本已发布。

PR #9256, opened 2025-06-24
fix: Fixed the issue that caused the error (bufio.Scanner: token too long) when viewing container log

zhengkunwang223:
Refs https://github.com/1Panel-dev/1Panel/issues/9241


wanghe-fit2cloud (commented):
/lgtm

PR #9699, opened 2025-07-28
fix: Fix the issue where the host IP is shown as 127.0.0.1 after enabling 'allowPort' in the application parameter settings.

zhengkunwang223:


wanghe-fit2cloud (commented):
/lgtm

PR #9866, opened 2025-08-05
fix: Fix the issue where the list keeps loading when autostart = false is set in the Supervisor configuration.

zhengkunwang223:


wanghe-fit2cloud (commented):
/lgtm

PR #9978, opened 2025-08-13
fix: Fix the issue of system certificates not auto-renewing.

zhengkunwang223:
#9975


ReddotCleaner:
https://github.com/1Panel-dev/1Panel/issues/9470

PR #10497, opened 2025-09-26
fix: Fix the issue where cache cleanup fails to exclude logs of curre…

ssongliu:
…ntly used

Refs #10484


zhengkunwang223 (commented):
/lgtm

Issue #11634, opened 2026-01-13
[Feature] Support batch import for Docker images


CN-Scars:
### 1Panel Version

v2.0.16

### Please describe your needs or suggestions for improvements

The current "Import Image" feature only supports selecting one .tar file at a time. When users need to import multiple image files (e.g., offline deployment, server migration), they have to repeat the operation multiple times, which is inefficient.

It would be helpful to support multi-select in the file picker, allowing batch import of multiple image files at once.

### Please describe the solution you suggest

This feature involves both frontend and backend changes. The implementation can reference the existing batch image pull (ImagePull) pattern.

Backend Changes：

| File                       | Changes                                                  |
| -------------------------- | -------------------------------------------------------- |
| agent/app/dto/image.go     | ImageLoad.Path string → Paths []string                   |
| agent/app/service/image.go | Loop to add subtasks, reference ImagePull implementation |
| agent/app/api/v2/image.go  | Adapt to new DTO structure                               |

Frontend Changes：

| File                                              | Changes                                                      |
| ------------------------------------------------- | ------------------------------------------------------------ |
| frontend/src/components/file-list/index.vue       | Add optional multi-select mode (multiple: boolean), maintain backward compatibility |
| frontend/src/views/container/image/load/index.vue | Support displaying list of selected files                    |
| frontend/src/api/interface/container.ts           | Update ImageLoad interface definition                        |

Expected TaskLog Output：
```
2026/01/13 23:05:04 ImportMirror [image1.tar,image2.tar,image3.tar] task starts [START]
2026/01/13 23:05:04 ----------------- image1.tar -----------------
2026/01/13 23:05:04 Import succeeded
2026/01/13 23:05:05 ----------------- image2.tar -----------------
2026/01/13 23:05:05 Import succeeded
2026/01/13 23:05:06 ----------------- image3.tar -----------------
2026/01/13 23:05:06 Import succeeded
2026/01/13 23:05:06 [TASK-END]
```

Reference Implementation：
  - Batch image pull PR #10532
  - ImagePull service: agent/app/service/image.go:266

### Additional Information

- Similar batch operation features already exist: batch image pull (#10532), batch image tag operations (#10597), batch container image upgrade (#9915)
- FileList is a shared component (used in 40+ places), backward compatibility will be ensured when adding multi-select mode
- Reuse existing TaskLog component to display batch import progress, no additional UI components needed
- I'm willing to contribute the code for this feature if the approach is acceptable

[...]

PR #11640, opened 2026-01-14
feat: Support batch import for Docker images

CN-Scars:
Refs #11634

#### Summary of your change
Backend (Agent):
- Modified ImageLoad DTO to accept an array of paths (Paths []string) instead of a single path
- Updated the image service to iterate through multiple files and create subtasks for each import operation

Frontend:
- Added multiple prop support to the FileList component for multi-file selection
- Enabled multi-select mode in the image load dialog by passing multiple: true

image

### DIFF
--- a/agent/app/service/image.go
+++ b/agent/app/service/image.go
@@ -321,35 +321,41 @@ func (u *ImageService) ImagePull(req dto.ImagePull) error {
 }
 
 func (u *ImageService) ImageLoad(req dto.ImageLoad) error {
-	taskItem, err := task.NewTaskWithOps(req.Path, task.TaskImport, task.TaskScopeImage, req.TaskID, 1)
+	taskItem, err := task.NewTaskWithOps(strings.Join(req.Paths, ","), task.TaskImport, task.TaskScopeImage, req.TaskID, 1)
 	if err != nil {
 		return fmt.Errorf("new task for image import failed, err: %v", err)
 	}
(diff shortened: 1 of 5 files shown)


agent/app/service/image.go
```diff
@@ -321,35 +321,41 @@ func (u *ImageService) ImagePull(req dto.ImagePull) error {
...
+			file, err := os.Open(currentPath)
+			if err != nil {
+				return err
+			}
+			defer file.Close()
+			client, err := docker.NewDockerClient()
+			if err != nil {
+				return err
+			}
+			defer client.Close()
```
ssongliu:
```
client, err := docker.NewDockerClient()
if err != nil {
    return err
}
defer client.Close()
```

It is advisable to move this part of the code outside the loop to avoid duplication.

frontend/src/components/file-list/index.vue
```diff
@@ -105,7 +105,8 @@
             </el-table>
         </div>
         <div class="file-list-bottom">
-            <div v-if="selectRow?.path">
+            <!-- Single select mode -->
```
ssongliu: recommend removing the code comments.

agent/app/service/image.go
CN-Scars: Done. Moved Docker client creation outside the loop.

frontend/src/components/file-list/index.vue
CN-Scars: Done. Removed.

PR #11642, opened 2026-01-15
Fix dns manual ssl wildcard

DeployThemAll:
## Summary

  This PR fixes three bugs in the DNS Manual SSL certificate flow:

  ### Bug 1: Order caching fails when Expires is zero
  ACME orders often have zero `Expires` initially. The condition `!Expires.IsZero()` caused valid cached 
  orders to be deleted and recreated with **different TXT values**, making verification fail.

  **Fix:** Check `Expires.IsZero() || Expires.After(now)`

  ### Bug 2: Wildcard and base domain TXT records overwrite each other
  When requesting SSL for both `example.com` and `*.example.com`, both authorizations have identifier    
  `example.com`, causing one TXT value to overwrite the other in the map. Users only saw 1 TXT record    
  instead of 2.

  **Fix:** Use `*.domain` as the map key for wildcard authorizations

  ### Bug 3: Only first TXT record checked
  When multiple TXT records exist (required for domain + wildcard), only the first DNS result was        
  checked. If the order of DNS responses varied, verification failed.

  **Fix:** Return all TXT values and check if expected value exists in any of them

  ## Test Plan
  - [x] Create SSL certificate with DNS Manual for single domain
  - [x] Create SSL certificate with DNS Manual for domain + wildcard
  - [x] Verify both TXT records are displayed
  - [x] Verify certificate is issued successfully after adding both TXT records

### DIFF
--- a/agent/utils/ssl/manual_client.go
+++ b/agent/utils/ssl/manual_client.go
@@ -71,12 +71,39 @@ type Resolve struct {
 }
 
 func (c *ManualClient) GetDNSResolve(ctx context.Context, websiteSSL *model.WebsiteSSL) (map[string]Resolve, error) {
-	order, err := c.client.AuthorizeOrder(ctx, acme.DomainIDs(getWebsiteSSLDomains(websiteSSL)...))
+	var order *acme.Order
+	var err error
+
+	// Check if we have an existing valid order for this SSL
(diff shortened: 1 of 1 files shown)


zhengkunwang223:
Thanks for the PR! There are some merge conflicts at the moment. Could you please resolve them and update the PR?

PR #11936, opened 2026-02-24
Added the ability to sort installed applications.

rowanchen-com:
#### What this PR does / why we need it?

Add drag-and-drop sorting for installed apps in the App Store > Installed tab.

Currently, users cannot customize the display order of installed applications. This PR adds a "Sort" button that enters a sort mode where users can drag app cards to reorder them. Favorites and non-favorites are kept in separate groups (fence/barrier) — favorites always stay on top and can only be reordered among themselves, same for non-favorites.

#### Summary of your change

**Backend:**
- Added `sort_order` field to `app_installs` table with DB migration
- Added `POST /apps/installed/sort/update` API endpoint to persist sort order
- Modified query ordering: `favorite DESC, sort_order ASC`
- When toggling favorite, automatically set `sort_order` to max+1 of the target group so newly favorited/unfavorited apps appear at the end of their group

**Frontend:**
- Added `sortablejs` dependency for drag-and-drop
- Added "Sort" button in the installed apps toolbar, with confirm/cancel buttons in sort mode
- Sort mode hides pagination and loads all apps for cross-page reordering
- Implemented `onMove` fence: favorites can only reorder among favorites, non-favorites among non-favorites
- Added `.sortable-ghost` style for drag visual feedback
- Added `sortMode` i18n key for all 10 languages (zh, zh-Hant, en, es-es, ja, ko, ms, pt-br, ru, tr)

**Files changed: 23 files (9 backend, 14 frontend)**

#### Please indicate you've done the following:
- [x] Made sure tests are passing and test coverage is added if needed.
- [x] Made sure commit message follow the rule of [Conventional Commits specification](https://www.conventionalcommits.org/).
- [x] Considered the docs impact and opened a new docs issue or PR with docs changes if needed.

PR #12026, opened 2026-03-02
fix: disable favorite toggle during sort mode and fix drag rebound issue

rowanchen-com:

Fix two issues with the installed app sorting feature added in #11936:

1. Toggling favorite during sort mode triggers backend `sort_order` recalculation and `search()` reload, which resets the drag-sorted order and breaks the Sortable fence logic.
2. Dragging cards occasionally "rebounds" back to their original position due to Sortable.js and Vue both manipulating the DOM simultaneously.

#### Summary of your change

**Disable favorite during sort mode (3 files, 6 changes):**
- `index.vue`: Pass `sortMode` prop to AppCard
- `card.vue`: Add `sortMode` prop, pass it to AppHeader
- `header.vue`: Add `sortMode` prop, set `:disabled="sortMode"` on both StarFilled and Star favorite buttons

**Fix drag rebound (1 file, 3 lines):**
- `index.vue`: In `onEnd` callback, revert Sortable's DOM manipulation before updating `data.value`, so Vue is the sole DOM owner and diff/patch works correctly

Files changed: 3 (frontend only)

- [x] Made sure commit message follow the rule of Conventional Commits specification.


rowanchen-com:
fix: prevent app icon from being squished in detail drawer

Issue #10345, opened 2025-09-11
[Bug] System - The Port Forwarding Feature of the Firewall Module (ufw) Incorrectly Handles Traffic from Within Docker.


AnxysUaen:
### Contact Information

Anxys@outlook.com

### 1Panel Version

Community v1.10.32

### Problem Description

  In the current version, if OpenResty is started to listen on port 8443 and port forwarding is configured to redirect port 443 to 8443, accessing any HTTPS website on the internet from within Docker will be forwarded to the host machine's OpenResty, which is unexpected.
  If this is a feature rather than a bug, it is hoped that a switch for "whether to exclude traffic from Docker" can be added to the port forwarding functionality.

### Steps to Reproduce

1. Start OpenResty and have it listen on port 8443.
2. Configure port forwarding from port 443 to port 8443.
3. Add a container, access its terminal, and use curl to request https://www.xx.com

### The expected correct result

In this scenario, the curl request from within the container to the external HTTPS website will be unexpectedly forwarded to the host's OpenResty service listening on port 8443.

### Related log output

```shell

```

wanghe-fit2cloud:
感谢反馈。上述现象是由于 ufw 的端口转发规则会同时作用在 Docker 容器的出站流量上，所以容器访问外部 HTTPS 时也被重定向到了宿主机。

目前你可以通过在宿主机添加一条 iptables 规则，手动排除 Docker 子网的流量来解决，例如：

```bash
# 假设 Docker 默认网桥是 172.17.0.0/16
iptables -t nat -I PREROUTING -s 172.17.0.0/16 -j RETURN
```

这样容器内部的请求就不会再命中端口转发规则了。

后续我们也会评估在端口转发功能中增加“排除 Docker 流量”的选项，来避免类似问题。

AnxysUaen:
tks
> 感谢反馈。上述现象是由于 ufw 的端口转发规则会同时作用在 Docker 容器的出站流量上，所以容器访问外部 HTTPS 时也被重定向到了宿主机。
> 
> 目前你可以通过在宿主机添加一条 iptables 规则，手动排除 Docker 子网的流量来解决，例如：
> 
> ```bash
> # 假设 Docker 默认网桥是 172.17.0.0/16
> iptables -t nat -I PREROUTING -s 172.17.0.0/16 -j RETURN
> ```
> 
> 这样容器内部的请求就不会再命中端口转发规则了。
> 
[...]

PR #12704, opened 2026-05-10
feat: add port security awareness panel to firewall page

zhangt2333:
Thanks for building 1Panel — I am a commercial license holder and use it daily. 

While using the firewall module extensively, we discovered that Docker container port mappings bypass iptables INPUT chain rules, so ports that appear protected by the firewall may actually be publicly exposed. This issue has existed almost since 1Panel's initial release, spanning 3+ years (2023–2026) with 18 related community reports:

- #12471
- #726
- #1846
- #3972
- #723
- #2717
- #2762
- #3594
- #4741
- #4984
- #5269
- #7312
- #7765
- #10157
- #10247
- #10345
- #10696
- #11313

Fully solving the Docker bypass problem (e.g., managing `DOCKER-USER` chain rules or integrating ufw-docker) would require significant changes to the firewall architecture. This PR takes a deliberate **first step** — giving users visibility into which ports are effectively protected and which are exposed through Docker. Once awareness is in place, further enforcement mechanisms can be built on top of it incrementally.

#### Summary of your change

- New API endpoint (`POST /hosts/firewall/port/security`, admin-only) — concurrently scans listening ports, Docker container bindings, and firewall rules to classify each port's effective security status
- New panel component (`firewall/port/awareness/index.vue`) above the existing rules table
- Small review surface — only `port/index.vue` is modified in existing code to import and render the new panel; the rest is additive
- i18n keys added to all 10 locale files (zh/en translated manually; others use English fallback — native-speaker review welcome)

**Screenshots:**

image

image

image

image

  - No existing test suite for the firewall service module; verified manually. Happy to add unit tests if maintainers prefer.


HynoR:
对firewalld 和 ufw 做适配测试了么

agent/app/service/firewall_port_security.go
```diff
@@ -0,0 +1,371 @@
...
+
+// determineStatus assigns a security status to a port based on its bind address,
+// source type, firewall state, and rule coverage. Priority order:
+// 1. Non-wildcard bind address → localOnly
+// 2. Docker/appStore source → dockerBypass
+// 3. Firewall inactive → firewallInactive
+// 4. Has matching rule → protected
+// 5. Otherwise → noRule
+func determineStatus(bindAddr, sourceType string, firewallActive, hasRule bool) string {
+	if !isWildcardAddress(bindAddr) {
```
HynoR: 有考虑过绑定本机 ip 的情况吗，绑定本机 ip 也属于公网暴露，在多 ip 的情况下这里判断根本不可靠

agent/app/service/firewall_port_security.go
```diff
@@ -0,0 +1,371 @@
...
+	}
+	return entries
+}
+
+// buildAppPortMap creates a lookup map from port number to app name.
+func buildAppPortMap(ctx context.Context) map[uint32]string {
+	result := make(map[uint32]string)
+	apps, err := appInstallRepo.ListBy(ctx)
+	if err != nil {
+		return result
```
HynoR: 同理，真的要直接这样处理错误吗

HynoR:
docker network bind 可以做的很复杂，这里也没有做复杂状态下判断

HynoR:
想法不错，但是建议多打磨打磨（你目前应该是在用的 ChatGPT codex，可以让他再多看看...)

wanghe-fit2cloud:
感谢贡献，这个方向是有价值的，Docker 端口映射绕过常规 INPUT 链规则确实容易让用户误判端口暴露状态。

不过当前实现暂不建议合并。主要原因是扫描逻辑还不够准确：Docker 发布端口不一定会表现为宿主机监听 socket，当前以监听端口为主数据源可能漏掉真正需要提示的 Docker 暴露端口；同时对绑定地址、防火墙规则、firewalld / ufw 以及复杂 Docker network bind 场景的判断还比较粗，可能给用户“已保护/安全”的错误结论。

建议继续打磨后再提交，至少需要补充 firewalld、ufw、iptables 以及不同 Docker 端口绑定方式下的验证，并调整判断逻辑和文案，避免安全状态误判。

zhangt2333:
@HynoR @wanghe-fit2cloud Thank you very much for your valuable feedback.

I'm currently continuing to work on this. I really need this feature and am strongly motivated to implement it well in 1Panel. 

**Could you please re-open this PR? There's no need to open another PR — I will set the current one to draft status. I'll keep pushing updates and will notify you when it's ready.**

zhangt2333:
Thank you very much for your valuable feedback — addressing each concern below.

---

## 1. Docker-published ports that aren't host LISTEN sockets

The classifier now consults two sources: host LISTEN sockets (`gopsutil`) and the Docker container port list (`c.Ports[]`). When the LISTEN-socket source returns nothing — e.g. Docker daemon with `{"userland-proxy": false}`, where iptables DNAT handles forwarding and no `docker-proxy` binds the host port — `c.Ports[]` still records the publish, so the fallback catches it.

Modes where **neither** source has data (macvlan / ipvlan / overlay / Swarm ingress) fall under the explicit scope limit in §7.

## 2. Bind-address judgment

`localOnly` is restricted to addresses provably unreachable from outside the host: loopback (`127.0.0.0/8` / `::1`) and link-local (`fe80::/10`). Any specific non-loopback IP (public NIC, docker bridge gateway `172.17.0.1`, private LAN) goes through the firewall-rule check — e.g. `systemd-networkd`'s public-NIC bind on `68/udp` now classifies as "No matching rule" rather than "Local only".

## 3. Cross-backend verification

Same rule set + container set on each backend; restarted `1panel-agent` between switches before screenshotting.

<table>
<tr>
<td align="center"><b>iptables 1.8.11</b></td>
<td align="center"><b>ufw 0.36.2</b></td>
<td align="center"><b>firewalld 2.3.1</b></td>
</tr>
<tr>
<td><a href="https://github.com/user-attachments/assets/83fa8fc8-b9d6-4d99-8337-fd62601c510f"></a></td>
<td><a href="https://github.com/user-attachments/assets/12060626-cc7b-46a4-aa57-3fe58bf83783"></a></td>
<td><a href="https://github.com/user-attachments/assets/b3a8a31b-731b-47b5-86df-da235cc6a699"></a></td>
</tr>
</table>

> *Implementation note:* when a port has both an accept and a drop rule, drop now takes precedence — security-conservative and deterministic across backends (firewalld lists accepts before rich-rule drops, iptables by chain order, ufw by rule index).

## 4. Docker network bind coverage

| Mode | Status | Result |
|---|---|---|
| bridge + wildcard (`-p 3306:3306`, explicit `0.0.0.0` / `[::]`) | `dockerBypass` | ✅ |
| bridge + `127.0.0.1` | `localOnly` | ✅ |
| bridge + specific non-loopback IP | dockerBypass (conservative — Docker traffic still uses FORWARD chain regardless of bind IP)  | ✅ |
| UDP, port range, multiple ports on one container | each classified independently | ✅ |
| `--net=host` | treated as a host process | ✅ |
| `userland-proxy=false` | resolved via `c.Ports[]` fallback (see §1) | ✅ |
| No `-p` published | not in scan scope | ✅ |
| macvlan / ipvlan / overlay / Swarm ingress | not in scan scope (see §7) | ⚠️ |

## 5. Wording — no safety overclaim

Applied to all 10 locales:

| Element | Copy |
|---|---|
| Banner (no findings) | "No issues found in this scan" / "本次扫描未发现问题" |
| Summary chips | "{N} need fixing / {N} to review / {N} OK" — clickable filters; "to review" routes to whichever of `noRule` or `firewallInactive` is active |
| `protected` | "Allowed by rule" / "规则放行" |
| `blocked` | "Blocked by rule" / "规则拦截" — neutral info tag, not green |
| `localOnly` | "Local only" / "仅本机" (actual address is in the "Bind Address" column) |
| `dockerBypassHasRule` | "A firewall rule exists for this port ({strategy} {port}/{proto}), but it does not apply to Docker-published ports." |
| Scope disclaimer | "Scope: host-side listeners and Docker default bridge networking. macvlan / ipvlan / overlay / Swarm ingress / external NAT paths are not evaluated." |

## 6. Test scenarios

**App Store deployed Docker.** 1Panel's Redis app installed via App Store → `sourceType: "appStore"`, with a direct "Go to App Settings" jump button in the "How to fix" popover.

**Allow external (accept rule).** `accept tcp/3306` against a wildcard-bound MySQL container → red `Docker Bypass` tag + ⚠️ tooltip explaining that the rule does not apply to Docker-published ports:

<table>
<tr>
<td><a href="https://github.com/user-attachments/assets/b992a263-b871-4b2f-a849-3700662de696"></a></td>
</tr>
</table>

**Deny external (drop rule).** `drop tcp/7777` against a host `nc` listener → neutral `Blocked by rule` tag, not green:

<table>
<tr>
<td><a href="https://github.com/user-attachments/assets/165db51d-c733-4b36-8db0-320ea58a4d05"></a></td>
</tr>
</table>

**Firewall installed but stopped.** The panel keeps rendering — that's exactly when a user most needs to see what's exposed. Every wildcard-bound host port reports `firewallInactive` (red); the per-row "Create Rule" action is suppressed because adding a rule to a stopped firewall has no effect.

<table>
<tr>
<td><a href="https://github.com/user-attachments/assets/b739545c-37d7-40cb-8d8d-4dbb00fe5e85"></a></td>
</tr>
</table>

## 7. Scope — what the scanner does *not* cover

macvlan / ipvlan / overlay containers and Docker Swarm ingress publishes don't produce a host-level LISTEN socket and aren't in `c.Ports[]`. They're invisible to the scanner by design; the scope disclaimer in the panel banner names them explicitly so users don't read "no issues" as an absolute claim.

---

If there are specific scenarios you'd still like to see covered or labelled differently, please point them out.

rowanchen-com:
Nice direction — 18 linked issues over 3+ years show this is a real pain point. I think the framing in the PR description is the right one: a full fix would mean taking over Docker's default iptables rules (DOCKER / DOCKER-USER chain management, ufw-docker-style integration, etc.), and the compatibility surface there is genuinely large — stopped containers, user-defined networks, `--network=host`, swarm ingress, userland-proxy on/off, non-default bridges, custom DNAT, IPv6. Starting with *visibility* before *enforcement* is a reasonable way to make progress.

One consequence of that scoping: because this PR explicitly doesn't change any iptables rule, the classification output **is the product**. There's no enforcement backstop that catches a misclassification at runtime — whatever the panel says, the user believes. That raises the bar on accuracy more than it would for a typical feature. A few spots where I think the current classifier could mislead:

## 1. `matchFirewallRule` ignores the rule's source `Address`

`FireInfo.Address` is populated for firewalld rich rules like `accept from 1.2.3.4 to port 80`, but the matcher only looks at port + protocol:

```go
for _, r := range rules {
    if r.protocol != "" && r.protocol != "tcp/udp" && r.protocol != proto {
        continue
    }
    if !portMatchesRule(portStr, port, r.portStr) {
        continue
    }
    ...
}
```

A rule that only allows a single source IP ends up counted as a global `accept`, and an IP-scoped `drop` ends up counted as a global `blocked`. In an awareness-only design this is exactly the kind of false positive/negative the panel can't afford — it's the "protected / safe" misjudgement the maintainer flagged.

**Suggestion:** when `r.address` is non-empty and not `Anywhere`, skip it for the global verdict, or surface it in the UI as *"rule scoped to X.X.X.X"* so the user sees the caveat.

## 2. §4 of your reply doesn't match `determineStatus`

Your table says:

> bridge + specific non-loopback IP → falls through to rule check ✅

But the code is:

```go
if isLoopbackOrLinkLocal(bindAddr) {
    return "localOnly"
}
if sourceType == "docker" || sourceType == "appStore" {
    return "dockerBypass"   // regardless of bindAddr
}
```

A container bound to `172.17.0.1` (reachable only on the bridge) or a specific public NIC IP both get labelled `dockerBypass` unconditionally. Either the code needs a branch for *"docker + non-wildcard bind → rule check"*, or the table row should be reworded to *"falls back to dockerBypass (conservative)"*. With code and docs disagreeing, a reviewer can't tell which is intentional.

## 3. `portMatchesRule` doesn't cover all rule syntaxes

Range handling currently accepts `3000-3100` and `3000:3100`. Two gaps:

- **iptables `multiport` rules** use comma-separated lists (`80,443,8080`) — not matched today.
- A rule with `portStr == ""` silently returns `false`, which happens to be correct but is worth an explicit guard + comment.

These won't show up in the §3 cross-backend screenshots because those use a single port per rule; a `*_test.go` with table-driven cases would cover them and directly answer the maintainer's ask for cross-backend verification.

## 4. Unit tests are probably non-optional for this merge

Given that classification accuracy *is* the deliverable, `matchFirewallRule`, `determineStatus`, `portMatchesRule`, and `buildDockerPortMap` all being pure functions makes them an easy target for table-driven tests. Minimum set I'd cover:

- accept + drop on the same port (drop precedence)
- range rule, multiport comma list, single port, empty port
- rich rule with a `from` clause (§1 above)
- loopback / link-local / wildcard / specific public IP / bridge-gateway bind
- `userland-proxy=false` synthetic-entry path
- UDP same port as a registered App Store `HttpPort`

15–20 cases, no external dependencies.

## 5. Scope the UI strictly to "visibility"

Since the iptables-modification half is deliberately out of scope, the panel should lean into being a read-only advisor and avoid anything that *looks* like enforcement:

- The "Create Rule" action on a `noRule` row is fine for non-Docker sources, but on a `dockerBypass` row it would create an ineffective rule. Consider disabling it there (or at least a tooltip warning) — a rule that "exists but doesn't apply" is worse than no rule.
- `dockerBypassHasRule` tooltip is already the right pattern: state the fact, don't imply a fix.
- `goToApp` for App Store sources is the correct escape hatch — it hands the user off to the one place where the fix can actually be made. Deep-linking to the specific app's edit page (instead of the list page) would be a nice follow-up.

## 6. Locale scope

`HynoR`'s comment about the translations is a fair point. Since most of the nuance lives in the wording (*"rule exists but doesn't apply to Docker-published ports"* is legally load-bearing), keeping this PR to `zh.ts` + `zh-Hant.ts` + `en.ts` and letting native speakers follow up on the other eight would tighten the review surface without losing value.

## 7. Smaller things

- **`appPortMap` is keyed only by port number** — a host UDP listener on the same port as an App Store app's `HttpPort` gets mislabelled `appStore`. Key it by `portKey{port, protocol}`.
- **`ContainerList(All: false)` excludes stopped containers** — a container defined with `-p 80:80` but currently stopped is invisible to the scan, even though its port mapping will be reinstated on restart. Depending on intent, `All: true` + a container-state column might be more honest.
- **Scan runs on every panel mount.** On a busy host `net.ConnectionsMax(..., 32768)` isn't cheap. A short TTL cache or a manual "Scan" button would be kinder in production.
- **Error style** — `fmt.Errorf(...)` works, but the rest of the service layer uses the project's `buserr` helper and gets i18n'd error messages for free.
- **`IFirewallService` widening** — the new method aggregates Docker + process table + firewall, which is a different concern than the rule CRUD this interface was originally about. Splitting into a separate `IPortSecurityService` would be a small change with cleaner boundaries.

---

Nothing above challenges the core design — awareness without iptables changes is the right first step given how many edge cases a full `DOCKER-USER` / ufw-docker-style takeover would bring. The suggestions are about tightening the accuracy and test story of that first step so the panel stays trustworthy. Happy to see this land.

zhangt2333:
@rowanchen-com Thanks for the thorough comment — your points drove concrete improvements in rule matching and test coverage. Addressing each below.

| Item | Action |
|---|---|
| Source-address matching (§1) | Fixed — `ruleAddress` carried through and surfaced |
| Table/code mismatch (§2) | Fixed — table updated |
| Multiport comma syntax (§3) | Fixed — `portMatchesRule` handles comma lists |
| Unit tests (§4) | Added — 64 cases, all pass |
| dockerBypass button (§5) | Already correct |
| Locale reduction (§6) | Keeping all 10 |
| Protocol-keyed appPortMap (§7a) | Acknowledged, deferred |
| Stopped containers (§7b) | Design choice |
| Scan caching (§7c) | Follow-up |
| `buserr` (§7d) | Not applicable here |
| Interface split (§7e) | Follow-up |

<details>
<summary><b>§1–§4 — Fixed</b></summary>

### §1. `matchFirewallRule` ignores `FireInfo.Address`

`firewallRuleEntry` now carries `address`, and `matchFirewallRule` returns it as a third value. The DTO exposes `ruleAddress` to the frontend. When the matched rule has a source-address scope (e.g. `accept tcp/3306 from 10.0.0.0/8`), the status tag gets a tooltip showing the rule and its source restriction. This way users can distinguish a globally scoped allow from a subnet-restricted one.

Covered by `TestMatchFirewallRule` → `"source-scoped_accept_3306_returns_address"`.

### §2. Table says "falls through to rule check" but code returns `dockerBypass`

The table now reads `"dockerBypass (conservative — Docker traffic still uses FORWARD chain regardless of bind IP)"`. The code behavior is correct and intentional: Docker's DNAT path bypasses INPUT regardless of the bind IP, so marking all Docker-sourced non-loopback ports as `dockerBypass` is the security-conservative call.

### §3. `portMatchesRule` misses multiport comma syntax

Added comma-split handling before the range check. On ufw, a rule like `80,443/tcp` stored by `ListPort()` as port `"80,443"` now correctly matches either port 80 or 443. Covered by `TestPortMatchesRule` → `comma_list_first / comma_list_middle / comma_list_last / comma_list_miss / comma_with_spaces`.

### §4. Unit tests

`firewall_port_security_test.go` with 5 test suites, 64 table-driven cases:

| Suite | Cases | Covers |
|---|---|---|
| `TestPortMatchesRule` | 16 | exact / dash range / colon range / comma list / empty |
| `TestMatchFirewallRule` | 11 | drop-wins-over-accept / protocol filter / `tcp/udp` wildcard / source-address return / comma list / range |
| `TestDetermineStatus` | 18 | loopback (v4/v6) / link-local / wildcard / public IP / bridge gateway / docker / appStore / firewallInactive / accept / drop / reject |
| `TestIsLoopbackOrLinkLocal` | 14 | 127.0.0.x / ::1 / fe80:: / public / private / bridge |
| `TestIsWildcardAddress` | 5 | 0.0.0.0 / :: / empty / non-wildcard |

All pass. Zero external dependencies — pure function tests.

</details>

<details>
<summary><b>§5 — Already correct</b></summary>

The template uses `v-if="row.status === 'noRule'"` for the Create Rule button and `v-else-if="row.status === 'dockerBypass'"` for the "How to fix" popover — they're mutually exclusive. Docker bypass rows only ever show the popover, never the button. Similarly, `firewallInactive` rows show neither.

</details>

<details>
<summary><b>§6–§7 — Not changing / Deferred</b></summary>

**§6. Locales** — The translations were done with native-speaker-level input per locale. Happy to adjust individual strings if specific mistranslations are pointed out.

**§7a. `appPortMap` keyed by port only** — In practice App Store apps use standard web ports (80, 443, 3000, 8080, etc.) that don't collide with UDP services. Can tighten to `map[portKey]string` if a real collision is reported.

**§7b. `ContainerList(All: false)`** — The panel scans currently-exposed ports. A stopped container has no listening socket and no active DNAT — the user can re-scan after restarting it.

**§7c. Scan on every mount** — Agreed that a TTL cache or explicit "Scan now" button would be better. Follow-up improvement.

**§7d. `fmt.Errorf`** — The single `fmt.Errorf` (connection-scan failure) is an internal error wrapped by `helper.InternalServer` before reaching the user. `buserr` is for user-facing domain validation errors.

**§7e. Interface split** — Agree this doesn't belong on `IFirewallService` long-term. Deferring to keep this PR's diff surface on existing files minimal.

</details>

zhangt2333:
Hi @wanghe-fit2cloud, Just checking in on this PR. I've addressed all the previous comments. 

Would appreciate your feedback when you have a chance! Let me know if you need any clarification. 

Thanks!

ssongliu:
感谢贡献，这里方向没啥问题，检查显示出未被放行但实际上可被外部访问的端口

但是直接放到上面会占用太大的篇幅，这边最好是在上面增加一个类似于端口安全扫描的按钮，点击按钮显示当前有哪些风险端口，给提示的同时最好也能支持一键修复

后面版本我们会考虑直接支持上该功能，再次感谢！

zhangt2333:
Hi @ssongliu, thank you for the feedback.

For context, I’m a **1Panel commercial license holder** and have used 1Panel for about **two years**. I ran into this issue myself: after using Docker port publishing and adding firewall rules in 1Panel, I thought the port was protected, but it was still externally reachable. I then found many related issues from other users, so I researched the problem and opened this PR out of genuine interest in improving 1Panel and contributing back to open source.

I’m still **strongly motivated to keep working on this feature**. If the team prefers a different UX, such as a **“Port Security Scan” button** with risk details and possible one-click fixes, I’m willing to rework the PR in that direction.

That said, I’m honestly disappointed by the process. I addressed the earlier feedback and made it clear that I was willing to keep iterating, but it is now unclear to me whether **community-driven work on this feature is welcome**.

Could you please clarify whether the team is open to **a community contributor continuing this feature**?

Thanks.

zhangt2333:
I have to admit that I’m becoming more pessimistic about whether this PR can move forward, and how long it would take to be implemented and released (as you can see above, I listed at least 18 related issues). So I want to leave a few thoughts here.

I don’t think this problem should be solved all at once. A complete fix would likely require larger and potentially breaking changes, such as changing how 1Panel handles Docker-related firewall behavior. My idea was to take an incremental path: **first add awareness**, so users can see which ports are exposed and whether their firewall rules are actually effective. That is what this PR tries to do, with mostly additive code to avoid interfering with existing behavior (as you can see, the code is almost entirely additive rather than removing or modifying existing logic). The next step could then be **actual remediation**, such as one-click fixes, in a follow-up PR or future iteration.

Thanks to `@HynoR`, `@wanghe-fit2cloud`, `@rowanchen-com` (using backticks to avoid bothering anyone) and others who paid attention to this PR and shared feedback. I’m not sure whether I can still push this PR forward by myself, but I hope these thoughts are useful for whoever continues working on this issue in the future.

**I still respect the work behind 1Panel and appreciate everyone’s time!**

ssongliu:
We have already identified this optimization opportunity and have started a systematic update of the firewall to address issues of this type.

zhangt2333:
Thanks, @ssongliu. Can't wait to see it in action :)

---

Nice! Found it in the v2.3.0 release

PR #12713, opened 2026-05-12
fix: parse supervisor uptime with optional days segment

nowubh:
 supervisorctl status reports running process uptime as part of the process description, for example:

```text
  pid 123, uptime 0:12:34
  pid 123, uptime 103 days, 0:12:34
```

  The previous parser used a fixed field index (fields[5]), which only worked for uptime values without the optional days segment. When the uptime contains days, only the day count was kept and the remaining time part was lost.
#### Summary of your change
- Parse RUNNING status details by locating the pid and uptime fields instead of relying on fixed indexes.
- Preserve the full uptime value by joining all fields after uptime.
- Add tests for standard Supervisor uptime formats with and without the optional days segment.

```release-note
Fix supervisor uptime parsing when `supervisorctl status` reports an uptime with days.
  ```

### DIFF
--- a/agent/app/service/host_tool.go
+++ b/agent/app/service/host_tool.go
@@ -613,8 +613,7 @@ func getProcessStatus(config *response.SupervisorProcessConfig, containerName st
 				Status: fields[1],
 			}
 			if fields[1] == "RUNNING" {
-				status.PID = strings.TrimSuffix(fields[3], ",")
-				status.Uptime = fields[5]
+				status.PID, status.Uptime = parseSupervisorRunningDetails(fields)
 			} else {
 				status.Msg = strings.Join(fields[2:], " ")
 			}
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: agent/app/service/host_tool.go [hunk 1]
```diff
@@ -623,3 +622,22 @@ func getProcessStatus(config *response.SupervisorProcessConfig, containerName st
...
+			}
+		case "uptime":
+			if i+1 < len(fields) {
+				uptime = strings.Join(fields[i+1:], " ")
+			}
+			return pid, uptime
+		}
+	}
+	return pid, uptime
+}
```
zhengkunwang223, line 640: 是不是可以限制一下从 fields[3：] 之后开始处理呢
避免 supvervisor name 取名   uptime 的情况(虽然概率很低)

nowubh: 已调整，pid 从固定的 status description 位置解析，uptime 只从 `fields[4:]` 开始 查找，避免进程名为 uptime 时误命中。同时补充了对应的测试用例。

PR #12826, opened 2026-05-23
Validate generated terminal commands

glmgbj233:
## Vulnerability
This fixes a prompt-injection data flow where attacker-controlled or otherwise untrusted input can influence LLM output, and that model output reaches a privileged sink.

- Source: `terminalAIClient.chatCompletionOpenAIResponses - agent/utils/terminal/ai/client.go:228`
- Sink: `LocalWsSession.sendWebsocketInputCommandToSshSessionStdinPipe - agent/utils/terminal/ws_local_session.go:205 - sink kind: command_exec`

## Attack scenario
An attacker can influence the LLM output at `terminalAIClient.chatCompletionOpenAIResponses - agent/utils/terminal/ai/client.go:228` so the model returns command text. If that text reaches `LocalWsSession.sendWebsocketInputCommandToSshSessionStdinPipe - agent/utils/terminal/ws_local_session.go:205 - sink kind: command_exec`, the prompt injection can become command execution rather than remaining untrusted generated content.

## Attack path
- 1. terminalAIClient.chatCompletionOpenAIResponses - agent/utils/terminal/ai/client.go:228
- 2. terminalAIClient.chatCompletionOpenAIResponses - agent/utils/terminal/ai/client.go:232
- 3. CommandGenerator.Generate - agent/utils/terminal/ai/command_generator.go:57
- 4. CommandGenerator.Generate - agent/utils/terminal/ai/command_generator.go:67
- 5. CommandGenerator.Generate - agent/utils/terminal/ai/command_generator.go:72
- 6. LocalWsSession.receiveWsMsg - agent/utils/terminal/ws_local_session.go:129
- 7. LocalWsSession.sendWebsocketInputCommandToSshSessionStdinPipe - agent/utils/terminal/ws_local_session.go:205

## Fix
The fix adds a trust-boundary check before LLM-derived data can reach the sink, while preserving the intended benign workflow. The dangerous effect is blocked after the LLM step instead of only reformatting or re-marshalling model output.

Changed files:
- `agent/utils/terminal/ai/command_generator.go`
- `agent/utils/terminal/ai/command_generator_test.go`
- `agent/utils/terminal/ws_local_session.go`
- `agent/utils/terminal/ws_local_session_test.go`
- `agent/utils/terminal/ws_session.go`

## Why this mitigates the issue
Prompt injection is mitigated by preventing model-controlled text from directly selecting, poisoning, replaying, executing, or persisting a privileged action across the identified boundary. Benign model output can still follow the constrained safe path.

## Functionality preservation
The repair is intended to keep normal safe behavior available and restrict only unsafe LLM-derived behavior at the sink boundary. Normal-case and exploit-regression coverage should be reviewed in the changed tests.

## Tests / verification
Local Go verification: failed, return code 1

```bash
GOTOOLCHAIN=auto GOPROXY=https://goproxy.cn,direct GOCACHE=<go-build-cache> GOTMPDIR=<go-tmp> go test ./utils/terminal ./utils/terminal/ai
```


ssongliu:
Thanks for the PR. I agree with adding a trust-boundary check before AI-generated text reaches PTY stdin, but the current implementation is too broad and may break normal terminal AI usage.

Please narrow the fix to the actual issue: AI-generated text must be single-line printable terminal input. It should reject \r, \n, NUL, ESC, and other control characters before writing to the PTY. This check should live close to the sink, e.g. in buildAIPastePayload, not only in the generator.

I don't think we should reject normal shell syntax such as |, ;, &, $(), ${}, or interpreter commands like python3, bash, node, etc. These are valid terminal commands and blocking them causes unnecessary behavior regression.

glmgbj233:
Thanks for the review. You’re right that the fix should be narrower. I’ll move the validation closer to the PTY sink and only reject control characters that can break terminal input, such as \r, \n, NUL, ESC, and other non-printable control bytes. I’ll keep normal shell syntax like |, ;, &, (), {} and interpreter commands allowed to avoid breaking legitimate terminal AI usage.

PR #12829, opened 2026-05-23
feat: upgrade lego to v5.1.0 and add Dynadot DNS provider

rowanchen-com:

Adding Dynadot requires lego v5.1.0 (the provider is not in v4), so this
PR does both at once: upgrade lego from v4.35.2 to v5.1.0, and wire up
the new Dynadot DNS provider.

#### Summary of your change

Commit 1 -- lego v5 upgrade

- Adapt to v5 API breaks: Obtain/ObtainForCSR/Revoke take a context.Context;
  AcmeUser.Key is crypto.Signer instead of crypto.PrivateKey; account
  registration is *acme.ExtendedAccount instead of *registration.Resource;
  certificate.Resource.Domain was renamed to Domains ([]string).
- dns01.AddRecursiveNameservers and dns01.AddDNSTimeout were removed;
  use dns01.SetDefaultClient(dns01.NewClient(&Options{...})) instead. The
  custom recursive nameserver fields in the website SSL form keep working.
- ObtainRequest disables the Common Name by default in v5. EnableCommonName
  is set to true to keep the v4 behaviour, otherwise older clients (Java
  keystores, embedded routers) that still rely on the CN field would fail
  the TLS handshake.
- lego.CertificateConfig.KeyType field was removed; the matching code path
  is dropped.
- lego log.Logger is now *slog.Logger; bridge it to the existing per-domain
  *log.Logger via slog.NewTextHandler so SSL apply logs still land in
  SSLLogDir.

Backward compatibility for users upgrading from v4

- certcrypto.KeyType string values were renamed: P256 -> EC256,
  P384 -> EC384, 2048 -> RSA2048, 3072 -> RSA3072, 4096 -> RSA4096,
  8192 -> RSA8192. A gormigrate migration rewrites the key_type column in
  website_acme_accounts, website_ssls and website_cas on first start, so
  old certificates renew automatically. A normalizeKeyType helper also
  handles the mapping at runtime as a safety net.
- v5 serializes private keys as PKCS#8 while v4 used SEC1 (EC) and PKCS#1
  (RSA). A new parsePrivateKeyPEM helper tries PKCS#8 first and falls back
  to the legacy formats, so existing account keys persisted under v4 keep
  loading without manual conversion.
- Frontend KeyTypes select options and the four form-initial keyType
  defaults are updated to the v5 form.

Commit 2 -- Dynadot / DnsPod

- Add Dynadot DNS provider, surfaced in the DNS account dialog next to
  the other API-key/secret based providers.
- Remove DnsPod from the provider list (lego v5 has no DnsPod
  implementation). The startup migration counts existing DnsPod rows in
  website_dns_accounts and logs a warning so the operator can switch them
  to TencentCloud, which now manages DNSPod-hosted zones via the same
  underlying API. Existing rows are not deleted.

Tested locally against AliYun, Spaceship, CloudFlare and Dynadot.
Certificates issued under v4 renew under the v5 binary in seconds (LE
authorization is reused), and brand-new domains go through the full
DNS-01 flow correctly.

- [ ] Considered the docs impact and opened a new docs issue or PR with docs changes if needed.

b0a450fd22f29ac60ed430d89d389765
3cbe75f061009b585206421b66d441ac
9d130586ce67a02acd7c1ef3a0a3ca58
98d7ed534e81c8aae2466e76324824a3
836234002216d5d7a5a8129cbb416c9f
bc6fdfedb4ce13c2a39e13c06d26222d
4bdbd96717254fe0f90c1081e8c8bc08


zhengkunwang223:
lego v5 升级是大改 
涉及证书签发、续签、撤销、DNS-01、私钥格式和迁移
你是否有做过所有的测试？

agent/app/dto/request/website_ssl.go
```diff
@@ -61,7 +61,7 @@ type WebsiteSSLObtain struct {
 type WebsiteAcmeAccountCreate struct {
 	Email      string `json:"email" validate:"required"`
 	Type       string `json:"type" validate:"required,oneof=letsencrypt zerossl buypass google custom"`
-	KeyType    string `json:"keyType" validate:"required,oneof=P256 P384 2048 3072 4096 8192"`
+	KeyType    string `json:"keyType" validate:"required,oneof=P256 P384 2048 3072 4096 8192 EC256 EC384 RSA2048 RSA3072 RSA4096 RSA8192"`
```
zhengkunwang223: P256 P384 2048 3072 4096 8192
这些是不是可以删掉

zhengkunwang223:
我看到 lego v5 使用了新的密钥常量

```
EC256
EC384
RSA2048
RSA3072
RSA4096
RSA8192
```

存量证书升级之后会导致续签失败
你参考 lego v5 中的 
github.com/go-acme/lego/v5/certcrypto/compat

```
func (k *KeyTypeCompat) UnmarshalText(text []byte) error {
    switch string(text) {
    case `P256`:
        *k = EC256
    case `P384`:
        *k = EC384
    case `2048`:
        *k = RSA2048
    case `3072`:
        *k = RSA3072
    case `4096`:
        *k = RSA4096
    case `8192`:
        *k = RSA8192
    default:
        kt, err := certcrypto.ToKeyType(string(text))
        if err != nil {
            return err
        }
        *k = KeyTypeCompat(kt)
    }
    return nil
}
```

这个方法 做一下兼容

agent/app/dto/request/website_ssl.go
zhengkunwang223: 不删的话 后端要有兜底

frontend/src/global/mimetype.ts
```diff
@@ -273,8 +273,8 @@ export const DNSTypes = [
         value: 'PorkBun',
     },
     {
-        label: 'DNSPod (' + i18n.global.t('ssl.deprecated') + ')',
-        value: 'DnsPod',
+        label: 'Dynadot',
+        value: 'Dynadot',
```
zhengkunwang223, line 282: DNSPOD 不要直接删除 
要考虑用户存量数据如何显示

PR #13399, opened 2026-07-28
feat(ai-agents): accept video input for MiniMax M3 model

octo-patch:
Reason: MiniMax M3 advertises video input, but the account-model sanitizer only accepted text and image.

## Changes

- `agent/app/service/agents_utils.go`: add `openclawVideoInputModelPattern` (matches `minimax-m3`) and extend the model entry input modalities to `["text", "image", "video"]` for matching vision models in `buildOpenclawModelEntry`. Other vision models keep `["text", "image"]`.
- `agent/app/service/agents_utils_input_test.go`: add unit tests covering MiniMax-M3 (video), MiniMax-M2.7 (text only), and a non-MiniMax vision model (text + image).

## Checks

- `git diff` secret scan (clean).
- Go toolchain was not available in the worker environment, so `go build`/`go test` were not executed here; the change is a small, self-contained addition to an existing regex-based branch and is covered by the new test.


zhengkunwang223:
不建议通过这种方式来特殊处理
我们看一下用什么方式好一些

